PR #259, opened 2014-06-01
Add out of the box support for reserve

sol:


snoyberg:
Just going through on an issue cleanup, and I realized that this is very relevant to the yesod devel GSoC project to use ide-backend for doing devel versions of various web applications. Are you familiar with that work?

In any event, this PR won't merge as-is, so I'm closing. If you think that it's still relevant, let's discuss and then figure out if it's worth rebasing on master.

PR #298, opened 2014-10-21
Add defaultAuthSettings and export AuthSettings data constructor.

cgag:
I didn't want to use the recommended method of creating AuthSettings by using a string literal, but it seemed that was actually the only way.  This PR exports the AuthSettings data constructor and a defaultAuthSettings value so that AuthSettings can be created manually.  

I'm not sure if this is desirable for anyone else, but I don't see any reason to prevent people from doing this.  If I'm missing something, please let me know.  I also wasn't sure what I should do to the wai-extra version number for something like this, let me know if I should have bumped number C instead of D or something else entirely.

Unrelated to this PR, but I wanted to note that I had some trouble realizing an http-basic middleware existed because wai-extra's description is vague on hackage and it's bundled with many other things.  I think it'd be useful to list capabilities in the package description instead of "Provides some basic WAI handlers and middleware" or perhaps break this package up.  Thoughts?

### DIFF
--- a/wai-extra/Network/Wai/Middleware/HttpAuth.hs
+++ b/wai-extra/Network/Wai/Middleware/HttpAuth.hs
@@ -5,10 +5,8 @@
 module Network.Wai.Middleware.HttpAuth
     ( basicAuth
     , CheckCreds
-    , AuthSettings
-    , authRealm
-    , authOnNoAuth
-    , authIsProtected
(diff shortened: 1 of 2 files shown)


creichert:
@cgag I augmented the wai-extra package description in #342. Does that fix work for you?

Issue #315, opened 2014-12-15
Changes to static files not picked up when running in a vboxsf mount.


pparkkin:
I'm building a web app using Warp (3.0.4), WAI (3.0.2), and Scotty (0.9.0). I develop my app using Vagrant on OS X with a Linux (Ubuntu Server) guest. I have my app in a directory on my OS X hard drive, which gets mounted into the Vagrant development box where I build and run and test.

Serving static files, changes to the files are not picked up even after killing and restarting the app. The response Content-Length header is updated, but the actual file content remains the same. (Somewhat similar behaviour is also reported in https://github.com/scotty-web/scotty/issues/55.)

And to sort of justify that this not simply an issue with Vagrant, or the vboxsf mount, I want to point out that the Python SimpleHTTPServer is able to pick up changes to files just fine.

Running the same app on the regular ext4 mount inside the VM changes to static files are picked up just fine as one would expect. Somehow this is still linked to the fact that it is running in a vboxsf mount.

$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 7.8.3
$ uname -a
Linux vagrant-ubuntu-trusty-64 3.13.0-40-generic *#*69-Ubuntu SMP Thu Nov 13 17:53:56 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

$ vagrant --version
Vagrant 1.6.5

If there is any more information I can provide, or any way I can help investigate this, please let me know.

kazu-yamamoto:
Closing this thanks to #317.

PR #317, opened 2014-12-22
WAI middleware that converts a ResponseFile into ResponseStream.

pparkkin:
Code and a simple test case.

I had a hard time figuring out a good test case, since the whole point is to make file responses look like file responses despite being a stream internally. I did add a couple to make sure it's not completely broken at least.

In response to issue #315 .


pparkkin:
Just commenting to make sure you see that I made the requested changes.

snoyberg:
Thanks, I don't get notifications for updating commits, so a message is very useful.

PR #322, opened 2015-01-05
Fix some warp comment typos

tekul:
I found a couple of typos/errors while reading through the source. I think the `connClose/closeConn` code has changed since the comment was written, so I'm not sure what's best, but it's definitely not "onnClose". There's also a user-supplied function called `onClose` which someone might think the comment is referring to.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -89,7 +89,7 @@ runSettings set app = withSocketsDo $
 -- 'serverPort' record.
 --
 -- When the listen socket in the second argument is closed, all live
--- connections are gracefully shut-downed.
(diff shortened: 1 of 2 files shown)


warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -259,7 +259,7 @@ fork set mkConn addr app dc fc tm counter = settingsFork set $ \ unmask ->
        bracket (onOpen addr) (onClose addr) $ \goingon ->
 
        -- Actually serve this connection.
-       -- onnClose above ensures the termination of the connection.
+       -- bracket with closeConn above ensures the connection is closed.
```
snoyberg: Yes, that looks correct.

snoyberg:
Thanks!

PR #323, opened 2015-01-07
Make sure Timer is tickled in sendfile.

derekelkins:
Without this, if it takes longer than 60 seconds to send a file (though it should be 30 seconds...), the connection will be dropped.


snoyberg:
Thanks

Issue #173, opened 2013-09-24
Application exception handling with warp


singpolyma:
`warp` has a way to provide an exception handler for every possible exception that happens while handling a request.  I would be nice if there could be _two_ exception handlers: one for before the `env` has been produced, and one for after.  The latter would then be able to receive the parsed request as an argument, which is very useful for error reporting tools, etc.  (I am trying to use HoneyBadger with WAI and warp.)

PR #326, opened 2015-01-13
Pass the Request to settingsOnException handlers when available.

romanb:
This PR tries to address what looks like a regression. At one point the `Maybe Request` argument was added to the `settingsOnException` handler and the `Request` passed along [1]. The related ticket was https://github.com/yesodweb/wai/issues/173. Only a couple of commits later the relevant line of code got lost [2]. As a result, it seems as if the `Maybe Request` argument is actually never passed a request, which is unfortunate.

[1] https://github.com/yesodweb/wai/commit/b2fa01e715b0703676e819f73db7a84ad451c9b7#diff-39726960319491135a952279b66943f2R249
[2] https://github.com/yesodweb/wai/commit/67a0bf58f0aaea3e68a13b05c9fe479352b67ccd#diff-39726960319491135a952279b66943f2L245

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -339,6 +339,13 @@ serveConnection conn ii origAddr isSecure' settings app = do
     recvSendLoop addr istatus fromClient = do
         (req', mremainingRef, idxhdr) <- recvRequest settings conn ii addr fromClient
         let req = req' { isSecure = isSecure' }
+        processRequest addr istatus fromClient req mremainingRef idxhdr
+            `catch` \e -> do
+                -- Call the user-supplied exception handlers, passing the request.
+                sendErrorResponse addr istatus e
+                onE settings (Just req) e
+
+    processRequest addr istatus fromClient req mremainingRef idxhdr = do
         -- Let the application run for as long as it wants
         T.pause th
 


gregwebs:
So the second commit could be a bad merge.

snoyberg:
Good catch, thank you!

kazu-yamamoto:
This patch introduced a huge space leak because this code is not tail recursive. I'm now working to solve this.

fumieval:
The cause is that `recvSendLoop` and `processRequest` are mutually recursive, while `recvSendLoop` piles up the user-supplied handler repeatedly.

Does just puting them into L281 (https://github.com/yesodweb/wai/blob/4b97050f96c5b2f0a300103da80698e00222e7b5/warp/Network/Wai/Handler/Warp/Run.hs#L281) make sense?

kazu-yamamoto:
@fumieval Yeah. What's exactly I'm wondering. 

@romanb What was your intention to introduce one more `catch`?

kazu-yamamoto:
After reverting #326, the performace of Mighty got back to what I expect.

core | 1 | 2 | 4 | 8 | 16
---- | ---- | ---- | ---- | ---- | ----
old | 20,710 | 39,237 | 76,326 | 127,449 | 199,093
#328 | 25,337 | 46,791 | 97,477 | 146,601 | 228,108
#326 | 32,965 | 64,128 | 112,599 | 177,649 | 243,256

kazu-yamamoto:
OK. `req` can be only used there. I understand.

kazu-yamamoto:
The patch above solves the space leak. It's not elegant. If you guys have any ideas to make code more elegant, please let me know. Anyway, I have released a new version of Warp to save the world.

snoyberg:
Thanks Kazu

romanb:
@kazu-yamamoto Apologies for overlooking this important detail and thanks for fixing it!

PR #330, opened 2015-01-28
Modified websockets server.lhs to better detect client disconnects.

dschalk:
In my experience, neither wai-websockets nor yesod-websockets detected client disconnects. Today I discovered that the underlying app at  http://jaspervdj.be/websockets/ had been revised; and low and behold, it now reliably detects client disconnects, quickly updating the state MVar and all connected browsers. I'm not sure that the breakthrough came with the most recent revision, but it wasn't there the last time I looked.

I changed "import Control.Exception (from Exception, handle)" to "import Control.Exception (finally)" and modified the code to mirror what Jasper Van der Jeugt recently published (ref. above).

I also took the second sentence off of: "The main function first creates a new state for the server, then spawns the actual server. For this purpose, we use the simple server provided by 'WS.runServer'." The second sentence is in Jasper's server.lhs file.

### DIFF
--- a/wai-websockets/server.lhs
+++ b/wai-websockets/server.lhs
@@ -11,13 +11,12 @@ nearby to check out the functions we use.
 > import Data.Char (isPunctuation, isSpace)
 > import Data.Monoid (mappend)
 > import Data.Text (Text)
-> import Control.Exception (fromException, handle)
+> import Control.Exception (finally)   
 > import Control.Monad (forM_, forever)
-> import Control.Concurrent (MVar, newMVar, modifyMVar_, readMVar, forkIO)
+> import Control.Concurrent (MVar, newMVar, modifyMVar_, readMVar, forkIO, modifyMVar)
 > import Control.Monad.IO.Class (liftIO)
 > import qualified Data.Text as T
 > import qualified Data.Text.IO as T
-
 > import qualified Network.WebSockets as WS
 > import qualified Network.Wai
 > import qualified Network.Wai.Handler.Warp as Warp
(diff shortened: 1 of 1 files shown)


creichert:
Looks good to me, I'm compiling and testing now.

Also, I did apply this patch to yesod-websockets: https://github.com/yesodweb/yesod/commit/595ee70386d63c760ecab07c876a11b4cb17fd9d

creichert:
Thanks! I'm wondering if the new disconnect code is relevant to yesod-websockets as well.

wai-websockets/server.lhs
```diff
@@ -35,42 +36,41 @@ later on.
...
 > removeClient :: Client -> ServerState -> ServerState
 > removeClient client = filter ((/= fst client) . fst)
 
-Send a message to all clients, and log it on stdout.
+Send a message to all clients, and log it on stdout:
 
 > broadcast :: Text -> ServerState -> IO ()
 > broadcast message clients = do
 >     T.putStrLn message
->     forM_ clients $ \(_, conn) -> forkIO $ WS.sendTextData conn message
```
creichert: Is removing the forkIO intentional?

dschalk:
I think it is.  Would you like for me to experiment and make a pull request if the patch works?

dschalk:
I took another look at the yesod-websockets chat example. It doesn't broadcast information about disconnects, so the browser interface wouldn't benefit from the patch. What about the server? Does it maintain sockets to nowhere when browsers disconnect? If so, some under-the-hood disconnect boilerplate might be beneficial.

dschalk:
I'm using Stackage. To further investigate the Yesod question, I ran "cabal update" and "cabal get websockets". The following code is in Network/ WebSockets/server.hs:
```haskell
runServerWith :: String -> Int -> ConnectionOptions -> ServerApp -> IO ()
runServerWith host port opts app = S.withSocketsDo $ do
    sock <- makeListenSocket host port
    _    <- forever $ do
        -- TODO: top level handle
        (conn, _) <- S.accept sock
        _         <- forkIO $ finally (runApp conn opts app) (S.sClose conn)
        return ()
    S.sClose sock
```
The use of "finally" is reassuring. I have no reason to think there is a problem, but I am certainly not a TCP/IP expert.

PR #332, opened 2015-02-01
Add additional DHE RSA and DSA ciphers to warp-tls default ciphers list.

creichert:
I was debugging TLS some tls related issues in Keter and noticed some potentially useful ciphers could be added to the `warp-tls` ciphers list.

The ciphers list in the Stunnel example from `hs-tls` has also been updated since the warp function was written: https://github.com/vincenthz/hs-tls/blob/master/debug/src/Stunnel.hs#L33

It might also help to use ciphersuite_all: https://github.com/vincenthz/hs-tls/blob/master/core/Network/TLS/Extra/Cipher.hs#L106.

Any feedback on this one would be greatly appreciated.

### DIFF
--- a/warp-tls/ChangeLog.md
+++ b/warp-tls/ChangeLog.md
@@ -1,3 +1,7 @@
+## 3.0.1.4
+
+Add additional Diffie-Hellman RSA and DSA ciphers to warp-tls.
+
 ## 3.0.1.3
 
 [Unable to allow insecure connections with warp-tls #324](https://github.com/yesodweb/wai/issues/324)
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
I would also like to add TLSExtra.cipher_ECDHE_RSA_AES128GCM_SHA256 and
TLSExtra.cipher_DHE_RSA_AES128GCM_SHA256 to the top of this list. They are used in HTTP/2.
The version of tls should be ">= 1.2.16".

snoyberg:
@kazu-yamamoto I have no objection. Can you push through those changes?

kazu-yamamoto:
Done in 49a40e7f2e190f6aa31a8c2ce429f3169dd8d7cb.

PR #333, opened 2015-02-02
Improve range handling to return 416

dylex:
Prevent parsing negative range end-points (ignored as invalid rather than generating an error), truncate ranges to file size, and correctly return 416 and range "*" when requested range is outside file bounds.


snoyberg:
Looks good to me. @kazu-yamamoto mind sanity checking this as well to make sure you're OK with the change?

kazu-yamamoto:
OK. I confirmed that this does not introduce performance regression. Let's merge this.

PR #334, opened 2015-02-06
Add runEnv: like run but uses $PORT

pbrisbin:
When making smaller Yesod sites not using the scaffold, by far the most common
way I run it is to look at $PORT then fall back. I thought maybe the helper
we've been using for this would be generally useful. It would certainly be nice
not to have to define it again in all our projects.

As usual, I'm not tied to the name.

If this is accepted, and when the version bump happens, please also update the
"Since" comment.

### DIFF
--- a/warp/Network/Wai/Handler/Warp.hs
+++ b/warp/Network/Wai/Handler/Warp.hs
@@ -20,6 +20,7 @@
 module Network.Wai.Handler.Warp (
     -- * Run a Warp server
     run
+  , runEnv
   , runSettings
   , runSettingsSocket
   , runSettingsConnection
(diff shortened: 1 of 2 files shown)


gregwebs:
Would it make more sense to call this something like `runPortEnv` since it is only checking the port?

pbrisbin:
Maybe. I don't find that any clearer or more intention-revealing, myself.

I went with `runEnv` because where `run` uses the the Port given, `runEnv` gets it from the environment -- neither mention `Port` in their name even though both involve one. In other words, I'd be more on board with `runPortEnv` if `run` were already named `runPort`.

snoyberg:
I'm OK with this as-is. @gregwebs I'll hold off on merging based on your objections.

gregwebs:
I agree with @pbrisbin's reply. However, I do object to the use of the partial `read` function. It is fine to throw an exception, but it should mention the input value that was read in from the env variable and `runEnv`.

pbrisbin:
That's a good point. I should be able to make that change on Monday.

pbrisbin:
I went with `fail` as a first swing since it seemed like the simplest thing that would work. If you'd prefer throwing some kind of actual exception type, let me know.

Another option would be to use the fallback on an invalid value (not just a missing value) but that felt too surprising for an end-user -- I'd rather a mis-configuration result in a useful failure than a silent change in behavior.

gregwebs:
looks great, thanks!

PR #338, opened 2015-02-22
warp: support blaze-builder >= 0.4

manny-fp:


manny-fp:
Not thoroughly tested, but builds and passes tests with both blaze-builder 0.3.3.4 and 0.4.

manny-fp:
I'll look into the Travis failure.

manny-fp:
@snoyberg I haven't been able to reproduce this exact error locally, but I've gotten other package version-related errors from doctest.  I have no prior experience with doctest, but my guess is that it's not defining the same preprocessor macros as cabal does.

kazu-yamamoto:
I cannot reproduce this doctest failure in my environment, either.
Which version of GHC is used in travis-ci?

@snoyberg It would be usefull if we add "ghc --version" and "cabal --version" to .`travis.yml`.

snoyberg:
@kazu-yamamoto Added :)

MaxGabriel:
We could setup Travis CI to use each of the GHC versions so that all of the supported versions are tested. Since each version would run on a separate travis instance it shouldn't add much time to the builds.

kazu-yamamoto:
@snoyberg Thanks. GHC is 7.6.3. OK. I will look into the source of this problem.

kazu-yamamoto:
@MaxGabriel It's awesome!

kazu-yamamoto:
I cannot reproduce this doctest errors with any versions of GHC and any versions of Cabal. So, I'm going to give up.

kazu-yamamoto:
I created warp repo solely and test it with travis. See this result: 

https://travis-ci.org/kazu-yamamoto/warp/builds/51978170

So, the doctest failure is specific to cabal-meta, I guess.

kazu-yamamoto:
Cc: @gregwebs

MaxGabriel:
@snoyberg Your latest commit to allow fast-logger 2.3.0 has doctests passing on GHC 7.6 and 7.8 btw https://travis-ci.org/yesodweb/wai/builds/52230990

kazu-yamamoto:
Can anyone explain what happened?

kazu-yamamoto:
The dependency of fast-logger to blaze-builder is the source of this problem, right?

snoyberg:
I'm guessing it pulled in conflicting versions of dependencies.

@MaxGabriel Are you suggesting we roll back my commit that disabled the doctests by default? I'm fine with that.

MaxGabriel:
@snoyberg yeah.

kazu-yamamoto:
I think that we should merge the repoDocTestFailures branch.

kazu-yamamoto:
Oh, that is #343! I agree to merge the pull request.

MaxGabriel:
I made a new branch b/c the old one had some commit noise in it

kazu-yamamoto:
Thank you for your excellent work!

PR #342, opened 2015-02-24
Augment wai-extra package description.

creichert:
Resolves #298.

### DIFF
--- a/wai-extra/wai-extra.cabal
+++ b/wai-extra/wai-extra.cabal
@@ -1,7 +1,69 @@
 Name:                wai-extra
 Version:             3.0.4.3
 Synopsis:            Provides some basic WAI handlers and middleware.
-description:         API docs and the README are available at <http://www.stackage.org/package/wai-extra>.
+description:
+  Provides basic WAI handler and middleware functionality:
+  .
+  * WAI Testing Framework
+  .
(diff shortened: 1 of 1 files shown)


snoyberg:
Does this actually add anything over the list of modules?

creichert:
@snoyberg good point. I added some enhancements but I'm okay if this doesn't add anything over the module list and existing haddocks.

What is with the build failures? https://travis-ci.org/yesodweb/wai/builds/51985424#L1111

Does #338 fix that?

snoyberg:
I'm tempted to disable the doctests on Travis, since they are very unreliable. @kazu-yamamoto any objection?

snoyberg:
Generally speaking, I'm not thrilled about the concept of cabal file descriptions, due to the necessity of duplicating information with the README.md file. Hopefully Hackage will fix this issue soon (https://github.com/haskell/hackage-server/issues/262 and https://github.com/haskell/hackage-server/issues/279).

creichert:
I agree. I primarily wanted to put the docs there for `wai-extra` since the complaint was specifically that they could not find the info.

Once the issue is fixed in Hackage I'll help move this directly to the README.md.

snoyberg:
Awesome, thank you!

PR #343, opened 2015-02-27
Reenable testing

MaxGabriel:
Based off of discussion in #338

This also enables travis CI testing on GHC 7.6 and 7.8, which @kazu-yamamoto and I thought was a good idea. There was a period where the doctests worked on 7.8 but not 7.6 so I wanted to get that in with the PR that reenabled doctests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: haskell
 
+ghc:
+    - 7.6
+    - 7.8
+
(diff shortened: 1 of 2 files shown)


snoyberg:
Heh, I was about to write a comment to say I'll merge once the tests pass... and then they passed while typing. Nice work!

PR #348, opened 2015-03-09
Use of Transport requires warp >= 3.0.8

### LINKED ISSUE #347: wai-tls 3.0.1.4 and 3.0.2 fail to build
wai-tls requires warp >= 3.0. But wai-tls references the `Transport` type from warp, which has only existed since warp 3.0.8.

ygale:
Fixes #347.


snoyberg:
Thanks!

PR #349, opened 2015-03-09
Support chain certs

ygale:
Allow wai-tls to support popular low-cost chain certificates, such as those issued by GoDaddy. This has already been supported by the tls library since version 1.2.15.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -18,8 +18,8 @@ module Network.Wai.Handler.WarpTLS (
     , tlsWantClientCert
     , tlsServerHooks
     , defaultTlsSettings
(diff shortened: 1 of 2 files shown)


warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -18,8 +18,8 @@ module Network.Wai.Handler.WarpTLS (
     , tlsWantClientCert
     , tlsServerHooks
     , defaultTlsSettings
-    , tlsSettings
```
snoyberg: Does this have to be a breaking change? That makes me very hesitant to move ahead with it.

ygale:
Good catch, thanks. No, it does not need to break anything; that was just a mistake. Sorry.

snoyberg:
Thanks!

PR #350, opened 2015-03-18
Update Mime.hs

tolysz:
Hi, there are fonts which already use WOFF2 standard.
http://dev.w3.org/webfonts/WOFF2/spec/#IMT


snoyberg:
Thanks!

PR #353, opened 2015-03-27
Fix Windows build with time-1.5

RyanGlScott:
`time-1.5` changed the type signature of `formatTime` and now exports `defaultTimeLocale`, which clashes with `defaultTimeLocale` from `old-locale`. This causes `warp` to fail when building on Windows using `time-1.5`. This commit uses CPP pragmas to fix the issue.


snoyberg:
Thanks!

PR #356, opened 2015-04-01
Add Cookie Handling to Network.Wai.Test

taladar:
This would be a useful addition (already hinted at in TODOs in the code anyway) to test WAI applications with some kind of login or session mechanism.

yesod-test already implements something similar and this implementation has the same limitations (no support for max-age) but not all WAI applications with login/sessions are Yesod applications.

This should also benefit hspec-wai if it is built against a version of wai-extra including this change (I did not test that but the source of hspec-wai seems to use Network.Wai.Test)

I am open for ideas for additional tests to add or changes to bring it in line with any layout or style rules for your packages I missed.

### DIFF
--- a/wai-extra/Network/Wai/Test.hs
+++ b/wai-extra/Network/Wai/Test.hs
@@ -25,8 +25,10 @@ module Network.Wai.Test
 
 import Network.Wai
 import Network.Wai.Internal (ResponseReceived (ResponseReceived))
+import Control.Applicative ((<$>))
 import Control.Monad.IO.Class (liftIO)
-import Control.Monad.Trans.State (StateT, evalStateT)
+import Control.Monad.Trans.Class (lift)
+import qualified Control.Monad.Trans.State as ST
 import Control.Monad.Trans.Reader (ReaderT, runReaderT, ask)
 import Control.Monad (unless)
 import Control.DeepSeq (deepseq)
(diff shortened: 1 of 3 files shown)


snoyberg:
Just headed to bed now so I can't look into this, but in principle I'm in
favor.

wai-extra/Network/Wai/Test.hs
```diff
@@ -47,18 +50,28 @@ import qualified Data.Text as T
...
+type Session = ReaderT Application (ST.StateT ClientState IO)
+
+type ClientCookies = Map ByteString Cookie.SetCookie
 
 data ClientState = ClientState
-    { _clientCookies :: Map ByteString ByteString
+    { clientCookies :: ClientCookies
     }
 
+getClientCookies :: Session ClientCookies
```
snoyberg: Do we need to expose any functions to users to inspect the cookies?

snoyberg:
Looks great to me. My only question is about whether we need to expose some functionality to users. Once we discuss that, I'm happy to merge and release. Thanks!

taladar:
I assume you are thinking about assertions about set cookies and the ability to set or delete cookies easily before or between requests?

snoyberg:
Yes, something along those lines.

taladar:
This should cover all the use cases I could think of.

snoyberg:
Looks great, thank you for the thoroughness!

taladar:
You are correct about Map.insert, for some reason I did not even think of that, probably because I started out with a copy of the Map.union bit merging new cookies into the client cookie store and changed that to merge only one cookie.

snoyberg:
OK, I'll modify it while doing the version bump.

taladar:
Thanks

Issue #354, opened 2015-03-31
Headers differ between HEAD and GET requests


DeathByTape:
I noticed today that the headers differ between HEAD and GET requests for the same resource. Is there a reason for this? [The RFC](http://tools.ietf.org/html/rfc2616#section-9.4) suggests that these headers be the same unless there exists a strong reason or special circumstance otherwise (i.e. the use of "SHOULD"). 

Here are the results of a local test:
```bash
$ curl -4 -i -X HEAD localhost:3000/wai/LICENSE
HTTP/1.1 200 OK
Date: Tue, 31 Mar 2015 01:07:02 GMT
Server: Warp/3.0.10.1
Content-Type: application/octet-stream
last-modified: Sun, 29 Mar 2015 16:31:40 GMT
```

```bash
$ curl -4 -i -X GET localhost:3000/wai/LICENSE
HTTP/1.1 200 OK
Content-Length: 1085
Accept-Ranges: bytes
Date: Tue, 31 Mar 2015 01:07:39 GMT
Server: Warp/3.0.10.1
Content-Type: application/octet-stream
last-modified: Sun, 29 Mar 2015 16:31:40 GMT

... Truncated content
```

This feature is particularly useful in conjunction with #345 (i.e. having Content-Length and Content-Type).

snoyberg:
What application are you running?

DeathByTape:
Good question (made me dig a little deeper into the code). `wai-app-static`, I believe; I am just running the `warp` binary directly. That said, after reading the source more closely it looks like this really should be giving a 405 in this application (I have tested that this is the case for POST, PUT, DELETE).

I am just starting to dig into the wai code (so I am largely unfamiliar with it at this point), but I suspect that this issue is coming from a bug in parsing the request data (i.e. `requestMethod` for a HEAD request is being assigned `H.methodGet`). I will investigate further when I have a chance a little bit later today.

snoyberg:
It looks to me like you're encountering the behavior of the [autohead](https://www.stackage.org/haddock/nightly-2015-03-31/wai-extra-3.0.5/Network-Wai-Middleware-Autohead.html) middleware, which automatically translates a HEAD request into a GET request. It looks like that middleware should be made smarter in the case of a file response. This may actually be trivial...

snoyberg:
[...]

PR #357, opened 2015-04-02
ensuring that GET and HEAD returns the same headers. (#354)

kazu-yamamoto:
@DeathByTape @snoyberg Please review this patch.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -193,9 +193,9 @@ sendResponse defServer conn ii req reqidxhdr src response = do
     mRange = reqidxhdr ! idxRange
     reqinfo@(isPersist,_) = infoFromRequest req reqidxhdr
     (isKeepAlive, needsChunked) = infoFromResponse rspidxhdr reqinfo
-    isGet = requestMethod req == H.methodGet
+    isHead = requestMethod req == H.methodHead
     rsp = case response of
-        ResponseFile _ _ path mPart -> RspFile path mPart mRange isGet (T.tickle th)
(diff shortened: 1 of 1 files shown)


snoyberg:
@DeathByTape I think we need to modify the autohead middleware to keep the ResponseFile response if available.

@kazu-yamamoto I'm actually somewhat confused about the `isGet` stuff; isn't the point to check if it's *not* HEAD?

kazu-yamamoto:
`isGet` means `isNotHead` in this case. Other methods are unsuitable here, I believe.

snoyberg:
I may be reading this code incorrectly, but I thought the code path is simply for a response which is returning a `ResponseFile`. While a GET response is by far the most likely case for that, it seems reasonable that an application could similarly use a `ResponseFile` for a POST or PUT request if desired. Is there any downside to changing `isGet` to `isNotHead`?

kazu-yamamoto:
I wanted to give a name of `hasBody` to this variable. But I did not since it is used already. 

I would avoid negative names for readability. Any suggestions? What about `useSendFile`?

snoyberg:
How about `isHead` and make the test `not isHead`?

kazu-yamamoto:
Is ed5afae what you want?

DeathByTape:
I submitted #358 with some changes to the middleware (effectively neutralizes it) and update `wai-app-static` to handle this. It is a PR into this branch (get-head) since it all seems like a single feature to me. I can change the PR location to master, however, if that's better.

kazu-yamamoto:
@snoyberg Please merge this pull request to master if you think OK.

snoyberg:
Merged and released, thanks guys!

PR #358, opened 2015-04-02
Update wai-app-static for proper HEAD support

DeathByTape:
Update wai-app-static to handle HEAD requests.

### DIFF
--- a/wai-app-static/Network/Wai/Application/Static.hs
+++ b/wai-app-static/Network/Wai/Application/Static.hs
@@ -214,10 +214,10 @@ staticApp set req = staticAppPieces set (W.pathInfo req) req
 
 staticAppPieces :: StaticSettings -> [Text] -> W.Application
 staticAppPieces _ _ req sendResponse
-    | W.requestMethod req /= "GET" = sendResponse $ W.responseLBS
+    | notElem (W.requestMethod req) ["GET", "HEAD"] = sendResponse $ W.responseLBS
         H.status405
         [("Content-Type", "text/plain")]
-        "Only GET is supported"
+        "Only GET or HEAD is supported"
 staticAppPieces _ [".hidden", "folder.png"] _ sendResponse = sendResponse $ W.responseLBS H.status200 [("Content-Type", "image/png")] $ L.fromChunks [$(embedFile "images/folder.png")]
 staticAppPieces _ [".hidden", "haskell.png"] _ sendResponse = sendResponse $ W.responseLBS H.status200 [("Content-Type", "image/png")] $ L.fromChunks [$(embedFile "images/haskell.png")]
 staticAppPieces ss rawPieces req sendResponse = liftIO $ do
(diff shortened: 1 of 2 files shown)


wai-extra/Network/Wai/Middleware/Autohead.hs
```diff
@@ -7,9 +7,7 @@ import Network.Wai
 import Data.Monoid (mempty)
 
 autohead :: Middleware
-autohead app req sendResponse
-    | requestMethod req == "HEAD" = app req { requestMethod = "GET" } $ \res -> do
-        let (s, hs, _) = responseToStream res
-        sendResponse $ responseBuilder s hs mempty
-    | otherwise = app req sendResponse
+autohead app req sendResponse = app req sendResponse
```
snoyberg: This change is wrong; if wai-app-static doesn't require the autohead middleware, then it should be removed from https://github.com/yesodweb/wai/blob/master/wai-app-static/WaiAppStatic/CmdLine.hs#L103, not disabling the middleware itself, which is used in other places.

DeathByTape: Thanks for pointing this out @snoyberg. Moving too fast this morning. I have updated the PR to reflect this. Thanks!

snoyberg:
Looks good. We should probably wait until the warp changes are merged and released, and then add a lower bound on the new version of warp. Otherwise, this may result in worse behavior than we have currently. (If that's *not* the case, please tell me.)

wai-app-static/Network/Wai/Application/Static.hs
```diff
@@ -214,10 +214,10 @@ staticApp set req = staticAppPieces set (W.pathInfo req) req
 
 staticAppPieces :: StaticSettings -> [Text] -> W.Application
 staticAppPieces _ _ req sendResponse
-    | W.requestMethod req /= "GET" = sendResponse $ W.responseLBS
+    | notElem (W.requestMethod req) ["GET", "HEAD"] = sendResponse $ W.responseLBS
```
snoyberg: Minor note, mostly just to pass it on: using `notElem` will be slightly slower than using an `||` and testing each possibility explicitly, since the latter can avoid any list creation/traversal overhead. It's theoretically possible that this would be optimized away, but I generally don't rely on it. In this case, I'm not convinced the performance difference is worth changing the code.

DeathByTape: Thanks for the note :) My haskell work has been limited to this point (my day job is mostly in [pick your favorite imperative language]), so any optimization tips are pretty useful to me.

In regards to your first comment, this PR is merging directly into the get-head branch and not master so that it could all go in as one feature. If you would prefer, I can reopen this as a PR to master (in which case I believe yes, we should wait for warp merge).

snoyberg: I didn't notice the branch, thanks for pointing that out. OK, this is good to merge, thank you!

Issue #186, opened 2013-10-28
handleExpect


kazu-yamamoto:
I resumed to hack the request parser of Warp. I have a question.

`Request.handleExpect` removes the Expect header from a list. Why should it be removed?

reiddraper:
> The expect header for 100-continue values is an artifact of the HTTP layer itself, and would have no bearing on the web application. There's no theoretical downside to passing it on to the application, but the application can't do anything meaningful with that information.

I think the web application _can_ do something meaningful with this. From [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt), section 8.2.3:

> In some cases, it might either be inappropriate or highly inefficient for the client to send the body if the server will reject the message without looking at the body.

As a concrete example, imagine writing an API or web application that processes large POST or PUT requests. Each request includes an authentication header (cookie, or otherwise). Based on the URL and header alone, the server may determine that the request is not authorized, and return HTTP 403, before the client has sent a multi gigabyte or megabyte request body.

Another example would be the web application responding with HTTP 413, having just looked at the Content-Length header.

How this fits in with Wai/Warp, I'm not entirely sure. One idea might be to introduce a Warp configuration option which would _not_ automatically respond with `100-continue`. _IFF_ the application called `Wai. requestBody` would Warp send `100-continue`, and begin reading. Since this would be an opt-in configuration option, existing users would not be affected.

Thoughts?

kazu-yamamoto:
I also feel the current behavior is hard-coding, too. I would like to let applications to decide their behavior.

reiddraper:
I'd be happy to take a try at changing this, if its something you'd be interested in merging.

kazu-yamamoto:
Yes. Please go ahead!

PR #360, opened 2015-04-07
Only conditionally produce HTTP 100 Continue

reiddraper:
Only produce HTTP 100 Continue if the application asks for the request
body. This allows the application to respond with a 4xx or 5xx status,
having only looked at the request headers, and never receiving the body.
An example use-case for this would be responding with HTTP 403 or 413,
without having to waste resources sending a large request body.

Addresses discussion in #186.

---

This is mostly a PR to start a discussion for now, as the behavior I've implemented is not (yet) behind a setting. This _could_ become the default behavior, but I wanted to start a discussion about this particular implementation, before getting into those details. Feedback appreciated.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Request.hs
+++ b/warp/Network/Wai/Handler/Warp/Request.hs
@@ -50,10 +50,12 @@ recvRequest :: Settings
             -> Source -- ^ Where HTTP request comes from.
             -> IO (Request
                   ,Maybe (I.IORef Int)
-                  ,IndexedHeader) -- ^
+                  ,IndexedHeader
+                  ,IO ByteString) -- ^
             -- 'Request' passed to 'Application',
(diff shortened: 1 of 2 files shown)


reiddraper:
cc @kazu-yamamoto

snoyberg:
I haven't followed this discussion closely enough, but at an initial review, it seems like a good change in behavior that shouldn't break existing applications. I'm not sure we need to hide the behavior behind any flags. I'll defer to @kazu-yamamoto though, as he probably understands this better than I do.

kazu-yamamoto:
LGTM. I have merged this patch with rebasing.

kazu-yamamoto:
@reiddraper Please close this if everything is OK for you.

reiddraper:
@kazu-yamamoto :+1:

PR #362, opened 2015-04-07
Extend isSecure and remove caveats about reverse proxy

pbrisbin:
This all started because I want to update `sslOnlyMiddleware` to not only add
the HSTS header, but also redirect non-http to https itself. Otherwise, users
still need to get to some https page organically for the header to be respected
and browser redirects to begin at all. This is what the analogous `force_ssl`
setting of Rails (and maybe other frameworks?) does, so I'm assuming it's
reasonable behavior to add to Yesod.

To do this, I first need a robust way to check if the current request is SSL.
Having read the caveats for the current `isSecure`, I decided to <del>update it</del> 
add a new function with slightly more elaborate checks that I believe still work in 
the presence of reverse proxies.

The checks were lifted from [here](https://github.com/rack/rack/blob/d68b93a9922b8bbfd76d5c7bcf5b63f2aec8f36f/lib/rack/request.rb#L72-88).

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,3 +1,8 @@
+## 3.0.7
+
+* Add appearsSecure: check if a request appears to be using SSL even in the
+  presence of reverse proxies [#362](https://github.com/yesodweb/wai/pull/362)
+
 ## 3.0.6.1
 
 * Test code: only include a Cookie header if there are cookies. Without this
--- a/wai-extra/Network/Wai/Request.hs
+++ b/wai-extra/Network/Wai/Request.hs
@@ -21,6 +21,8 @@ import qualified Data.ByteString.Char8 as C
 -- @'isSecure'@. This is not always the case though: for example, deciding to
 -- force a non-SSL request to SSL by redirect. One can safely choose not to
 -- redirect when the request /appears/ secure, even if it's actually not.
+--
+-- Since 3.0.7
 appearsSecure :: Request -> Bool
 appearsSecure request = isSecure request || any (uncurry matchHeader)
     [ ("HTTPS"                  , (== "on"))
--- a/wai-extra/Network/Wai/Request.hs
+++ b/wai-extra/Network/Wai/Request.hs
@@ -1,0 +1,34 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- | Some helpers for interrogating a WAI 'Request'.
+
+module Network.Wai.Request
+    ( appearsSecure
+    ) where
+
+import Data.ByteString (ByteString)
+import Network.HTTP.Types (HeaderName)
+import Network.Wai (Request, isSecure, requestHeaders)
+
+import qualified Data.ByteString.Char8 as C
+
+-- | Does this request appear to have been made over an SSL connection?
+--
+-- This function first checks @'isSecure'@, but also checks for headers that may
+-- indicate a secure connection even in the presence of reverse proxies.
+--
+-- Note: these headers can be easily spoofed, so decisions which require a true
+-- SSL connection (i.e. sending sensitive information) should only use
+-- @'isSecure'@. This is not always the case though: for example, deciding to
+-- force a non-SSL request to SSL by redirect. One can safely choose not to
+-- redirect when the request /appears/ secure, even if it's actually not.
+appearsSecure :: Request -> Bool
+appearsSecure request = isSecure request || any (uncurry matchHeader)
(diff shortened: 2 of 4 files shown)


reiddraper:
I think adding a helper that also checks the headers would be preferable. These headers are easily spoofed (unless I'm missing something), and this change currently removes the ability (unless `_isSecure` is exported) from easily checking _for sure_ if the request was made over TLS.

pbrisbin:
Seems reasonable.

At what level (module) do you think this would make the most sense? Any ideas on name?

pbrisbin:
I just put it in the same `Internal` module for now.

reiddraper:
How about putting it in [wai-extras](https://hackage.haskell.org/package/wai-extra)? Seems like a valuable helper function, and that would be where I might look for it, personally.

snoyberg:
I agree with @reiddraper, putting it in wai-extra is a good idea.

pbrisbin:
Moved to wai-extra, added a spec, did the version bump thing.

snoyberg:
LGTM, just waiting for Travis to complete. This reminded me of some code I wrote recently:

https://github.com/fpco/wai-middleware-crowd/blob/master/src/Network/Wai/Approot.hs

Perhaps I should throw my `smartApproot` in that module too and have it use your `appearsSecure`, what do you think?

pbrisbin:
My only concern for adding `smartApproot` here is that "APPROOT" is strictly a Yesod concept (right?), so having wai know about it seems like coupling in the wrong direction.

What about breaking it up into "HOST || localhost" and "appearsSecure ? https : http" functions and add those here? Then your `smartApproot` is reduced to a thin wrapper that checks `APPROOT` (a Yesod concern) or calls into these functions (a generic convenience).

pbrisbin:
And I say this as someone who could actually use `smartApproot` in what I'm planning next... http://sprunge.us/EDHV?hs

snoyberg:
There's no reason for APPROOT to be Yesod specific; any web application that needs to generate links needs some concept of an application root. For example, in wai-middleware-crowd, we need to generate complete URLs, and that solution doesn't touch Yesod at all.

I'm fine with breaking it up into more fine-grained units, but I do believe `smartApproot` is generally applicable.

pbrisbin:
> I do believe smartApproot is generally applicable.

Clearly, I defer to you -- was just offering my opinion. Go for it :)

snoyberg:
No, please *don't* defer to me, I'd rather have a discussion on the subject than just make a decision :). It could be that the terminology is Yesod-specific, I'm not sure. If there are ways to make it seem more generally useful, I'd like to hear it.

pbrisbin:
If wai needs some concept of an application root (which does seem reasonable), I think it'd make sense to add that somewhere (A new field on `Request`, incorporated somehow in `Application`, etc). In a Yesod app (for example) you could read ENV/yaml/whatever, and pass it in when you build your `waiApp`. Alternatively, how about a Middleware that takes an env var name, reads it, and puts the value somewhere where wai code can access it when needed?

For some reason, having wai code reach out and read some arbitrary environment variable feels wrong, kind of global-variable-ish (because it is, really). Do we know similar libraries in other languages handle this?

snoyberg:
I don't think it warrants a new field, but we could certainly use the `Vault` field to add this data, and provide it via a middleware. The idea of looking up an environment variable is just one of many options for how to get this information: it could be hard-coded, read from a config file, or determined on the fly from the request each time. In wai-middleware-crowd, I was setting this up to be a configuration option (in fact, any function with the same type as `smartApproot` would be a valid approot-getting function). Perhaps I should add this logic to a Middleware.Approot module with a few useful examples, and then leave the .Request module you just created for the functionality you mentioned above (scheme and host).

pbrisbin:
> it could be hard-coded, read from a config file, or determined on the fly from the request each time

I don't 100% follow this. Wouldn't hard-coding/reading-from-config require the kind of thing I'm describing where you pass the app-root in from the outside? Isn't determining it on the fly from the request basically everything you have in -crowd *but* the `APPROOT` reading?

Anyway, I think it's OK not to get bogged down in these details because...

> Perhaps I should add this logic to a Middleware.Approot module with a few useful examples, and then leave the .Request module you just created for the functionality you mentioned above (scheme and host).

That sounds great to me. Seeing those examples for context should clear up the confusion I'm having -- I suspect we're advocating the same things at the end of the day.

snoyberg:
I'll write this up as a PR and ping you, most likely tomorrow (I'm tied up
the rest of the day).

PR #364, opened 2015-04-09
Add approot middleware

snoyberg:
Pinging @pbrisbin following our discussion. Also, @gregwebs, you may be interested in this.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/Approot.hs
+++ b/wai-extra/Network/Wai/Middleware/Approot.hs
@@ -24,6 +24,7 @@ module Network.Wai.Middleware.Approot
     , fromRequest
       -- * Functions for applications
     , getApproot
+    , getApprootMay
     ) where
 
 import           Control.Exception     (Exception, throw)
@@ -34,7 +35,7 @@ import           Data.Maybe            (fromMaybe)
 import           Data.Typeable         (Typeable)
 import qualified Data.Vault.Lazy       as V
 import           Network.Wai (Request, vault, Middleware, requestHeaderHost)
-import           Network.Wai.Request   (appearsSecure)
+import           Network.Wai.Request   (appearsSecure, guessApproot)
 import           System.Environment    (getEnvironment)
 import           System.IO.Unsafe      (unsafePerformIO)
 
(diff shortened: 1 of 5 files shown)


wai-extra/Network/Wai/Middleware/Approot.hs
```diff
@@ -0,0 +1,120 @@
...
+
+data ApprootMiddlewareNotSetup = ApprootMiddlewareNotSetup
+    deriving (Show, Typeable)
+instance Exception ApprootMiddlewareNotSetup
+
+-- | Get the approot set by the middleware. If the middleware is not in use,
+-- then this function will return an exception.
+--
+-- Since 3.0.7
+getApproot :: Request -> ByteString
```
pbrisbin: Thoughts on preferring `Maybe` over exceptions here?

wai-extra/Network/Wai/Middleware/Approot.hs
```diff
@@ -0,0 +1,120 @@
...
+-- URL. If you are relying on the approot for some security-sensitive purpose,
+-- it is highly recommended to use @hardcoded@, which cannot be spoofed.
+--
+-- Since 3.0.7
+fromRequest :: Middleware
+fromRequest = approotMiddleware (return . extract)
+
+extract :: Request -> ByteString
+extract req =
+    (if appearsSecure req then "https://" else "http://") `S.append`
```
pbrisbin: For my own curiosity, is `S.append` better than `Data.Monoid.(<>)`?

wai-extra/Network/Wai/Middleware/Approot.hs
```diff
@@ -0,0 +1,120 @@
...
+    (fromMaybe "localhost" (requestHeaderHost req))
```
pbrisbin: Random style nit, but `(fromMaybe "localhost" $ requestHeaderHost req)`

wai-extra/Network/Wai/Middleware/Approot.hs
```diff
@@ -0,0 +1,120 @@
...
+-- approot, the worst that can happen is that the client will get an incorrect
+-- URL. If you are relying on the approot for some security-sensitive purpose,
+-- it is highly recommended to use @hardcoded@, which cannot be spoofed.
+--
+-- Since 3.0.7
+fromRequest :: Middleware
+fromRequest = approotMiddleware (return . extract)
+
+extract :: Request -> ByteString
+extract req =
```
pbrisbin: Is this function generally useful? Worth moving to `Network.Wai.Request` with a more descriptive name?

pbrisbin:
Few minor/random questions -- LGTM either way.

wai-extra/Network/Wai/Middleware/Approot.hs
snoyberg: Yeah, I debated that too. Yes, I think we should make that move.

wai-extra/Network/Wai/Middleware/Approot.hs
snoyberg: No, not that I'm aware of. I just did it this way to avoid potential future warnings when `<>` is (almost inevitably) exported from `Prelude`.

wai-extra/Network/Wai/Middleware/Approot.hs
snoyberg: I preferred exceptions because in the use cases I was considering, a `Nothing` should never happen. A `Nothing` only occurs in the case of user error (not setting up the middleware). However, now that you raise it, I suppose an application could speculatively check for the presence of the middleware and, if not present, do something else. So now my thought is: let's provide two versions, `getApproot` as defined, and `getApprootMay`, with a short explanation in the comments along what I've just written.

snoyberg:
Good feedback, thanks. I'm not going to be able to make these changes until Sunday, but I'll update then (unless someone else feels like patching it up and merging before then, which I'm fine with).

wai-extra/Network/Wai/Middleware/Approot.hs
pbrisbin: Gotcha, avoiding the CPP seems reason enough for now.

wai-extra/Network/Wai/Middleware/Approot.hs
pbrisbin: :+1:

snoyberg:
Never mind, I had a few extra minutes before having to shut down for the weekend, changes pushed.

snoyberg:
Any objections to merging as-is?

pbrisbin:
Sorry for the delay. This LGTM.

PR #377, opened 2015-05-24
Update sample.hs -> handler-webkit

BethAr:
I believe `return` was incorrect, it produced a type error:
```
  Couldn't match type `WAI.Response' with `IO WAI.ResponseReceived'
    Expected type: (WAI.Response -> IO WAI.ResponseReceived)
                   -> IO WAI.ResponseReceived
      Actual type: (WAI.Response -> IO WAI.ResponseReceived)
                   -> WAI.Response
```
Passing a response instead of return solved the issue and the webkit browser works.

### DIFF
--- a/wai-handler-webkit/sample.hs
+++ b/wai-handler-webkit/sample.hs
@@ -7,4 +7,4 @@ main :: IO ()
 main = run "Sample App" app
 
 app :: Application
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks!

PR #380, opened 2015-06-10
Shutdown faster.

tvh:
This replaces the `IORef` in `Counter` with a `TVar`. That way `gracefulShutdown` can just use `retry` to listen on changes to the counter instead of waiting arbitrarily long.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Counter.hs
+++ b/warp/Network/Wai/Handler/Warp/Counter.hs
@@ -1,24 +1,27 @@
 module Network.Wai.Handler.Warp.Counter (
     Counter
   , newCounter
-  , isZero
+  , waitForZero
   , increase
   , decrease
   ) where
 
-import Network.Wai.Handler.Warp.IORef
+import Control.Concurrent.STM
+import Control.Monad (unless)
 import Control.Applicative ((<$>))
 
-newtype Counter = Counter (IORef Int)
+newtype Counter = Counter (TVar Int)
 
 newCounter :: IO Counter
-newCounter = Counter <$> newIORef 0
+newCounter = Counter <$> newTVarIO 0
 
-isZero :: Counter -> IO Bool
-isZero (Counter ref) = (== 0) <$> readIORef ref
+waitForZero :: Counter -> IO ()
+waitForZero (Counter ref) = atomically $ do
+    x <- readTVar ref
+    unless (x == 0) retry
 
 increase :: Counter -> IO ()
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
@tvh Do you have a problem on the original code in the real world? And do you have any data that this patch makes graceful shutdown faster? I'm just asking.

A while ago, I heard that Michael is afraid of introducing STM to Warp. So, you need to persuade him. :-)

I'm planning to introduce STM to support HTTP/2 in the future. So, I stand by you.

meteficha:
Isn't a STM transaction blocked on `retry` guaranteed to be retried as soon as any variables it touched are changed?  This patch seems to change an expected delay of 5s to effectively zero, but I didn't try it.

tvh:
Yes, if a transaction does a `retry`, it will listen to any changes on any variable read up to that point. When it is then changed, it will be scheduled to run again.

Another side effect of this may even be a (negligible) performance increase. I benchmarked both `atomicModifyIORef'` and `modifyTVar'` against each other and it seems `modifyTVar'`wins out.

```.haskell
λ> getNumCapabilities 
4
λ> benchmark . whnfIO $ newIORef False >>= \ref -> mapConcurrently (const $ replicateM_ 100 $ atomicModifyIORef' ref $ \x -> (not x, ())) [1..100]
benchmarking...
time                 45.08 ms   (42.78 ms .. 46.69 ms)
                     0.994 R²   (0.987 R² .. 0.998 R²)
mean                 41.43 ms   (40.42 ms .. 42.64 ms)
std dev              2.444 ms   (2.042 ms .. 2.905 ms)
variance introduced by outliers: 20% (moderately inflated)

λ> benchmark . whnfIO $ newTVarIO False >>= \ref -> mapConcurrently (const $ replicateM_ 100 $ atomically $ modifyTVar' ref not) [1..100]
benchmarking...
time                 5.060 ms   (5.015 ms .. 5.107 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 5.072 ms   (5.028 ms .. 5.138 ms)
std dev              156.6 μs   (108.1 μs .. 218.4 μs)
variance introduced by outliers: 14% (moderately inflated)
```

tvh:
Also: Warp uses Async, so STM is already in the code.

kazu-yamamoto:
OK. Let's merge.

@tvh Thank you for your contribution!

@snoyberg If you have concerns, let's continue discussion.

snoyberg:
No concerns, just didn't have a chance to review. I trust your judgement
here.

PR #386, opened 2015-06-17
Exception handling

### LINKED ISSUE #382: warp: no exception upon an empty request
I registered an exception handler through `setOnException.`

When I send some junk instead of a request, I see an exception, as expected:

```
Warp: Invalid first line of request: "asdf"
```

However, if I send a single newline, the client gets back `500 Something went wrong`, but no exception is printed.

I think the property that any 500 error is logged on the server side is a very desirable one.

This is warp-3.0.2.3; apologies if this has been fixed in a later version.

kazu-yamamoto:
Trying to fix #382.

- EOF is now `ConnectionClosedByPeer` instead of `NotEnoughLines []`
- `onE` does not ignore `NotEnoughLines []`
- `defaultOnException` and `defaultOnExceptionResponse` are exported
- `defaultOnException and defaultOnExceptionResponse` sends 400 or 500
- Documentation is updated

Behavior:

- A single line:

> Incomplete request headers, received: []

and 400 response

- Junk:

> Warp: Invalid first line of request: "junk"

and 400 response

- EOF:

> Warp: Client closed connection prematurely

and no response

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -179,9 +179,7 @@ runSettingsConnectionMakerSecure set getConnMaker app = do
                    f
 
 onE :: Settings -> Maybe Request -> SomeException -> IO ()
-onE set mreq e = case fromException e of
-    Just (NotEnoughLines []) -> return ()
-    _                        -> settingsOnException set mreq e
+onE set mreq e = settingsOnException set mreq e
 
 -- Note that there is a thorough discussion of the exception safety of the
 -- following code at: https://github.com/yesodweb/wai/issues/146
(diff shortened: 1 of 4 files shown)


warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -179,9 +179,7 @@ runSettingsConnectionMakerSecure set getConnMaker app = do
                    f
 
 onE :: Settings -> Maybe Request -> SomeException -> IO ()
-onE set mreq e = case fromException e of
-    Just (NotEnoughLines []) -> return ()
-    _                        -> settingsOnException set mreq e
+onE set mreq e = settingsOnException set mreq e
```
snoyberg: Maybe just replace usage of `onE` with `settingsOnException`?

kazu-yamamoto: Thanks. It's done on master.

PR #389, opened 2015-06-20
Cryptonite and non-deprecated Warp APIs

nwf:
These patches allow my system to build with cryptonite (and not the packages it replaces) installed.  It was also seemingly necessary to modify packages in this repository to work together after c7072101959982e95745c3898a1e81fede8b4bee.  While here, some warnings were silenced.

### DIFF
--- a/wai-app-static/Network/Wai/Application/Static.hs
+++ b/wai-app-static/Network/Wai/Application/Static.hs
@@ -41,7 +41,7 @@ import qualified Data.Text as T
 
 import Data.Either (rights)
 import Network.HTTP.Date (parseHTTPDate, epochTimeToHTTPDate, formatHTTPDate)
-import Data.Monoid (First (First, getFirst), mconcat)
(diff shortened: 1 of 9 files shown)


kazu-yamamoto:
@nwf Thank you for your contribution. May I ask you to fix Travis CI errors before merging so that we can be confident about the merging. Yes, I admit that this is my fault.

snoyberg:
Thanks for the PR, but I'm going to need to address the tls issue at least with conditional compilation to keep support with both tls 1.2 and 1.3. I'll push a commit for that soon. The rest of the commit looks good, once the build failure is addressed as @kazu-yamamoto mentioned.

kazu-yamamoto:
I think we can close this issue thanks to 725b03a2409cc863426e22c69a1cb84316cb902c and 1fc3e1123248da2290ed9e4262f21f1ff2973942.

PR #392, opened 2015-06-29
Input buffer

kazu-yamamoto:
This is the second step to improve the performance of WarpTLS. Since TLS record tells the length of the body, we can make use of the length to read input data. This code is a little bit redundant because some APIs are also used in my HTTP/2 implementation. Note that HTTP/2 frames also tell payload length. I designed APIs very carefully and hope that they are clean enough.

The current bottleneck is cryptonite. This patch does not improve the file upload performance in HTTP/1.1 over TLS because cyptonite consumes 40+ percent of CPU powers. But I believe that this is worth merging.

@snoyberg Please review this.


snoyberg:
LGTM, I'm OK with merging.

kazu-yamamoto:
Thanks. Merged.

PR #393, opened 2015-07-03
wai-extra: Add stripHeader/s Middleware component

erikd:
This Middleware component allows specific HTTP response headers
(chosen by header name) to be dropped if the Wai Request matches
a predicate.

Tests included.

I can merge this to master myself, but just want to see if everyone thinks its ok.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/AddHeaders.hs
+++ b/wai-extra/Network/Wai/Middleware/AddHeaders.hs
@@ -3,6 +3,7 @@
 -- Since 3.0.3
 module Network.Wai.Middleware.AddHeaders
     ( addHeaders
+    , mapResponseHeader
     ) where
 
 import Network.HTTP.Types   (ResponseHeaders, Header)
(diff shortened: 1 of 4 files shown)


creichert:
Looks great, nice tests and docs to go with it. +1.

snoyberg:
Thanks! @kazu-yamamoto is in the middle of some refactoring of these libraries right now, so I'm not sure if it's safe to release a new version of wai-extra. @kazu-yamamoto any objection?

gregwebs:
I think this would be improved if `stripHeader` was not conditional on the request and instead it could just be combined with an `ifRequest` function.

erikd:
@gregwebs Can you give me a type signature for `ifRequest`?

kazu-yamamoto:
@snoyberg I'm just developing HTTP/2. I don't think this is not related to wai-extra. Am I missing something?

After HTTP/2 work, I will try the multiple range problem (by introducing file information cache).

gregwebs:
@erikd @snoyberg let me know what you think of the `ifRequest` branch
https://github.com/yesodweb/wai/compare/ifRequest?expand=1

erikd:
@gregwebs  Looks fine by me. It might also be nice to be able to condtitionally add headers in the `AddHeaders` module.

gregwebs:
ok, I made PR #395 with the changes. I moved some helpers into the WAI package so they can be reused with AddHeaders

PR #394, opened 2015-07-03
wai-extra: Add NOINLINE pragmas as needed

erikd:
The docs for `unsafePerformIO` say:

    Use {-# NOINLINE foo #-} as a pragma on any function foo that
    calls unsafePerformIO. If the call is inlined, the I/O may be
    performed more than once.


gregwebs:
Thanks!

PR #397, opened 2015-07-09
RequestLogger doc fixes

levinotik:
Removed incorrect doc notes about log functions being implemented on top of logCallback


snoyberg:
Thanks!

PR #399, opened 2015-07-14
supporting HTTP/2.

kazu-yamamoto:
This patch implements HTTP/2 ([RFC 7540](https://tools.ietf.org/html/rfc7540)). 

This code passes the following three tests:

-- All test cases provided by [h2spec](https://github.com/summerwind/h2spec)
-- Test cases provided by [nghttp](https://nghttp2.org/documentation/nghttp.1.html)
-- Field test in the real web world

Recent Firefox and Chrome supports HTTP/2 over TLS. With this patch, WarpTLS automatically communicates with Firefox and Chrome in HTTP/2. There is no settings to disable HTTP/2 at this moment. We should implement them if necessary.

The following picture illustrates rough architecture:

```

         Input queue                          Output queue
Receiver ----------> Thread pool for WAI apps -----------> Sender
```

In HTTP/2, multiple streams are asynchronously sent over a TCP connection.

- The receiver receives HTTP/2 frames, decodes them and enqueues WAI requests to the input queue. The logic of the input buffer is highly tuned.
- A worker thread in the thread pool dequeues a WAI request, gives it to a WAI application, takes its response and enqueues the response to the output priority queue.
- The sender packs a response in the output buffer and send it.

One concern: the number of worker threads in the worker pool is now hard coded -- 10. Unlike ResponseFile and ResponseBuilder, ResponseStream occupies the worker thread. So, if 10 streamings are served, no worker is available. We should fix this if necessary.

@snoyberg Please review and merge this if you think OK.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -289,6 +289,7 @@ httpOverTls TLSSettings{..} s bs0 params = do
     TLS.contextHookSetLogging ctx tlsLogging
     TLS.handshake ctx
     writeBuf <- allocateBuffer bufferSize
+    -- Creating a cache for leftover input data.
     ref <- I.newIORef ""
     tls <- getTLSinfo ctx
     return (conn ctx writeBuf ref, tls)
@@ -318,6 +319,8 @@ httpOverTls TLSSettings{..} s bs0 params = do
                 void (tryIO $ TLS.bye ctx) `finally`
                 TLS.contextClose ctx
 
+        -- TLS version of recv with a cache for leftover input data.
+        -- The cache is shared with recvBuf.
         recv cref = do
             cached <- I.readIORef cref
             if cached /= "" then do
@@ -326,6 +329,7 @@ httpOverTls TLSSettings{..} s bs0 params = do
               else
                 recv'
 
+        -- TLS version of recv (decrypting) without a cache.
         recv' = handle onEOF go
           where
             onEOF e
@@ -338,6 +342,7 @@ httpOverTls TLSSettings{..} s bs0 params = do
                   else
                     return x
 
+        -- TLS version of recvBuf with a cache for leftover input data.
         recvBuf cref buf siz = do
             cached <- I.readIORef cref
             (ret, leftover) <- fill cached buf siz recv'
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -239,12 +239,26 @@ runTLSSocket' tlsset@TLSSettings{..} set credential sock app =
       , TLS.serverSupported = def {
           TLS.supportedVersions = tlsAllowedVersions
         , TLS.supportedCiphers  = tlsCiphers
+        , TLS.supportedHashSignatures = [
+            (TLS.HashSHA256, TLS.SignatureRSA)
+          , (TLS.HashSHA1, TLS.SignatureRSA)
+          ]
         }
       , TLS.serverShared = def {
           TLS.sharedCredentials = TLS.Credentials [credential]
         }
-      , TLS.serverHooks = tlsServerHooks
+      , TLS.serverHooks = tlsServerHooks {
+          TLS.onALPNClientSuggest = Just alpn
+        }
       }
+
+alpn :: [S.ByteString] -> IO S.ByteString
+alpn xs
+  | "h2"    `elem` xs = return "h2"
+  | "h2-16" `elem` xs = return "h2-16"
+  | "h2-15" `elem` xs = return "h2-15"
+  | "h2-14" `elem` xs = return "h2-14"
+  | otherwise         = return "http/1.1"
 
 ----------------------------------------------------------------
 
@@ -275,8 +289,9 @@ httpOverTls TLSSettings{..} s bs0 params = do
     TLS.contextHookSetLogging ctx tlsLogging
     TLS.handshake ctx
     writeBuf <- allocateBuffer bufferSize
+    ref <- I.newIORef ""
     tls <- getTLSinfo ctx
-    return (conn ctx writeBuf, tls)
+    return (conn ctx writeBuf ref, tls)
   where
     backend recvN = TLS.Backend {
         TLS.backendFlush = return ()
@@ -284,12 +299,13 @@ httpOverTls TLSSettings{..} s bs0 params = do
       , TLS.backendSend  = sendAll s
       , TLS.backendRecv  = recvN
       }
-    conn ctx writeBuf = Connection {
+    conn ctx writeBuf ref = Connection {
         connSendMany         = TLS.sendData ctx . L.fromChunks
       , connSendAll          = sendall
       , connSendFile         = sendfile
       , connClose            = close
-      , connRecv             = recv
+      , connRecv             = recv ref
+      , connRecvBuf          = recvBuf ref
       , connWriteBuffer      = writeBuf
       , connBufferSize       = bufferSize
       }
@@ -302,7 +318,15 @@ httpOverTls TLSSettings{..} s bs0 params = do
                 void (tryIO $ TLS.bye ctx) `finally`
                 TLS.contextClose ctx
 
-        recv = handle onEOF go
+        recv cref = do
+            cached <- I.readIORef cref
+            if cached /= "" then do
+                I.writeIORef cref ""
+                return cached
+              else
+                recv'
+
+        recv' = handle onEOF go
           where
             onEOF e
               | Just TLS.Error_EOF <- fromException e       = return S.empty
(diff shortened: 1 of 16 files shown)


snoyberg:
Whoa, awesome! Will review now.

snoyberg:
I got the following when trying to test this:

```
Preprocessing test suite 'spec' for warp-3.1.0...

Network\Wai\Handler\Warp\HTTP2\Types.hs:25:8:
    Could not find module `Network.HTTP2.Priority'
    Perhaps you meant Network.HTTP.Proxy (from HTTP-4000.2.19)
    Use -v to see a list of the files searched for.
Completed all 8 actions.

--  While building package warp-3.1.0 using:
      C:\\Users\\Michael\\AppData\\Local\\Programs\\stack\\i386-windows\\ghc-7.8.4\\bin\\runhaskell.exe -package=Cabal-1.18.1.5 -clear-package-db -global-package-db -package-db=C:\Users\Michael\AppData\Roaming\stack\snapshots\i386-windows\lts-2.9\7.8.4\pkgdb\ C:\Users\Michael\AppData\Local\Temp\stack6928\Setup.hs --builddir=.stack-work\dist\i386-windows\Cabal-1.18.1.5\ build test:doctest test:spec
    Process exited with code: ExitFailure 1
```

snoyberg:
I can't say I understand the details going on here, but I'm OK with merging once I can compile and run the test suite on Windows.

creichert:
Very cool. I don't have anything to add but I was able to compile and
run the tests. Looking forward to testing this out more.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -284,12 +299,13 @@ httpOverTls TLSSettings{..} s bs0 params = do
...
       }
-    conn ctx writeBuf = Connection {
+    conn ctx writeBuf ref = Connection {
         connSendMany         = TLS.sendData ctx . L.fromChunks
       , connSendAll          = sendall
       , connSendFile         = sendfile
       , connClose            = close
-      , connRecv             = recv
+      , connRecv             = recv ref
+      , connRecvBuf          = recvBuf ref
```
alexanderkjeldaas: document the ref, recvBuf or recv functions?

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -314,6 +338,36 @@ httpOverTls TLSSettings{..} s bs0 params = do
                   else
                     return x
 
+        recvBuf cref buf siz = do
+            cached <- I.readIORef cref
+            (ret, leftover) <- fill cached buf siz recv'
+            I.writeIORef cref leftover
+            return ret
+
+fill :: S.ByteString -> Buffer -> BufSize -> Recv -> IO (Bool,S.ByteString)
```
alexanderkjeldaas: document this function?

warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
```diff
@@ -0,0 +1,307 @@
...
+
+    loop = do
+        hd <- recvN frameHeaderLength
+        if BS.null hd then
+            enqueue outputQ OFinish highestPriority
+          else do
+            cont <- processStreamGuardingError $ decodeFrameHeader hd
+            when cont loop
+
+    processStreamGuardingError (_, FrameHeader{..})
```
alexanderkjeldaas: when there are multiple `Context{...}` and `FrameHeader{..}` it is a bit hard to know where `streamId`, `continued` etc are coming from.   Suggest `FrameHeader{streamId}`.

warp/Network/Wai/Handler/Warp/HTTP2/Types.hs
```diff
@@ -0,0 +1,235 @@
...
+              | LOne B.BufferWriter
+              | LTwo ByteString B.BufferWriter
+
+----------------------------------------------------------------
+
+data Context = Context {
+    http2settings      :: IORef Settings
+  , streamTable        :: StreamTable
+  , concurrency        :: IORef Int
+  , continued          :: IORef (Maybe StreamId)
```
alexanderkjeldaas: This seems to indicate the stream id that the next frame should be a continuation of?  Documenting this context record would make reading the code a lot easier I think.

kazu-yamamoto:
@snoyberg `Network.HTTP2.Priority` is provided in the "http2" lib. Would you add the "http2" lib in the necessary place in `warp.cabal`? I cannot reproduce this, so I cannot try.

warp-tls/Network/Wai/Handler/WarpTLS.hs
kazu-yamamoto: `Network.Wai.Handler.Warp.Types` says:

```
    -- | The connection receiving function. This returns "" for EOF.
    , connRecv        :: Recv
    -- | The connection receiving function. This tries to fill the buffer.
    --   This returns when the buffer is filled or reaches EOF.
    , connRecvBuf     :: RecvBuf
```

`ref` implements a cache for leftover input data.

kazu-yamamoto:
@snoyberg I remembered that older "http2" does not provide `Network.HTTP2.Priority`. So, I added the lower boundary.

warp/Network/Wai/Handler/Warp/HTTP2/Types.hs
kazu-yamamoto: When a header is large, it is divided into a HEADERS frame and CONTINUATION frames. In HTTP/2, frames for several streams can be mixed. But for CONTINUATION, this is not the case. RFC 7540 says:

> Other frames (from any stream) MUST NOT occur between the HEADERS
>   frame and any CONTINUATION frames that might follow.

This field is to implement this requirement.

warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
kazu-yamamoto: Thanks and done.

meteficha:
Awesome work, @kazu-yamamoto!  

This bit worried me:

> Unlike ResponseFile and ResponseBuilder, ResponseStream occupies the worker thread. So, if 10 streamings are served, no worker is available. We should fix this if necessary.

Then I read the source code: the 10 worker threads are _per connection_, not global :).  So the only concern should be resource usage.

Perhaps [`parallel-io`](http://hackage.haskell.org/package/parallel-io-0.3.3/docs/Control-Concurrent-ParallelIO-Local.html) could be used to manage the thread pools?  It even has a `extraWorkerWhileBlocked` which could be used for `ResponseStream`s (assuming there's some other limit to the maximum number of concurrent streams).

meteficha:
On another thought... it doesn't look like `parallel-io` would fit the pattern here.  It's synchronous.

kazu-yamamoto:
@meteficha Yes, 10 workers are spawn per HTTP/2 connection.

kazu-yamamoto:
@snoyberg Does the current version still fail on Windows?

snoyberg:
I'll test tomorrow, sorry for the delay.

kazu-yamamoto:
No problem. You waited for me. I can wait for you.

alexbiehl:
Windows build went without problems using stack (needed to add http2 and hex packet to extra-deps) with lts-2.9 (ghc-7.8.4).

Are there any tests which specifically target the new protocol?

kazu-yamamoto:
> Are there any tests which specifically target the new protocol? 

No. I have not implemented HTTP/2 client side. So, I cannot automate testing at this moment.

Rather, I manually test the HTTP/2 feature with Firefox, h2spec and nghttp as mentioned above.

snoyberg:
Looks good on Windows (a few test failures, but nothing that doesn't exist on Windows). I think this is good to be merged, go for it when you're ready!

W00t! Congratulations Kazu, awesome work :+1:

kazu-yamamoto:
Thanks. Merged.

Now I have an idea to spawn worker threads dynamically. I will implement it today.

kazu-yamamoto:
I changed the version of warp-tls to 3.1.0. @snoyberg Please revert this if you don't like this.

I also updated the documents of warp and warp-tls so that they explicitly mention HTTP/2.

snoyberg:
That's perfect. Should we release this to Hackage, perhaps with a caveat
announcement that this is some new behavior?

kazu-yamamoto:
Let's release Warp 3.1.0 and WarpTLS 3.1.0 in the next week. I will write a blog article in advance.

snoyberg:
Cool, feel free to do so yourself, or just ping me and I'll do the release.

kazu-yamamoto:
>Now I have an idea to spawn worker threads dynamically. I will implement it today.

Done in fedab88cf1d84456a2cf19ce453b7571fc24c68b.

creichert:
@kazu-yamamoto Do you know exactly which sha1 was uploaded to Hackage as Warp-3.1? Was it 70fa8a695f33875e5ef4528db2cd871eb9de411f ? I want to tag it for helping some testing in the `servant` repository.

kazu-yamamoto:
@creichert Sorry. I forgot to push tags. I did now.

creichert:
Thanks!

PR #401, opened 2015-07-23
Adding Request Body to RequestLogger

jarjee:
The body is correctly retrieved, but is never included in the final log message. This commit adds a request body field to the log output.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RequestLogger.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger.hs
@@ -222,6 +222,7 @@ detailedMiddleware' cb ansiColor ansiMethod ansiStatusCode app req sendResponse 
                  return (req', body)
             _ -> return (req, [])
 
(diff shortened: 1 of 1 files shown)


andrewthad:
Quick question. If the request body is binary data (like an image), would this potentially send control characters to the terminal and mess up the logs? You might want to do something like call `decodeUtf8'` on it first to make sure that it's textual data and not binary data.

snoyberg:
Agreed with @andrewthad. Also, the current behavior is intended for just capturing form fields, which are in fact being logged. Logging the full request body is a new piece of functionality. We should also take into account the fact that the request body may be very large, and therefore only a subset should likely be displayed.

jarjee:
@andrewthad Just tested it, and can send control characters to the terminal. Whoops. I've made a commit to use decodeUtf8'.
@snoyberg The current default is to only display a body as long as it is <2048, which may still be too large. I'd imagine that this is something that would probably be best if it is tweak-able, since part of the logger is specifically for debugging output, after all.

snoyberg:
Thanks!

PR #405, opened 2015-07-27
Treat exceeding the concurrency limit as a stream error.

### LINKED ISSUE #404: Warp treats exceeding the concurrency limit as a connection error.
Per http://http2.github.io/http2-spec/index.html#rfc.section.5.1.2, servers are required to treat this as a stream error.  Warp should reset the offending stream instead of terminating the connection.  The relevant code is Receiver.hs:158.

I have a potential fix that I'll push to a fork shortly.

awpr:
Previously it caused the entire connection to be torn down.  The HTTP2 spec requires servers to treat this as a stream error, not a connection error.

Fixes #404.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
@@ -324,5 +324,7 @@ stream FramePriority header bs Context{outputQ} s Stream{streamNumber} = do
 
 -- this ordering is important
 stream _ _ _ _ (Open Continued{}) _ = E.throwIO $ ConnectionError ProtocolError "an illegal frame follows header/continuation frames"
+-- Ignore frames to streams we have just reset, per section 5.1.
+stream _ _ _ _ st@(Closed (ResetByMe _)) _ = return st
 stream FrameData FrameHeader{streamId} _ _ _ _ = E.throwIO $ StreamError StreamClosed streamId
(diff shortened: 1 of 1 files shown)


creichert:
Awesome, the commit looks good to me. Do you think it's worth adding a
test for this behavior? My thought is it would be very nice to start
getting features specific to the spec under test coverage.

kazu-yamamoto:
Please add a test case to h2spec if possible.

kazu-yamamoto:
There is not http2 client packages, so it is not possible to test all in Haskell at this moment.

awpr:
Actually, based on the following sentence from section 5.4.1, it's acceptable to kill the connection in this case: "An endpoint can end a connection at any time.  In particular, an endpoint MAY choose to treat a stream error as a connection error."  That explains why the relevant h2spec test passed.

Still, I think it's preferable to let the connection survive here -- nothing is corrupted, and the client is required not to continue using a stream after we send a RST_STREAM for it.

Maybe it would make sense for h2spec to have a "don't wantonly kill the connection" mode that warns if you upgrade stream errors to connection errors.  If I change it to disallow treating stream errors as connection errors, a lot of things start failing, which suggests Warp is terminating the connection in a lot of cases where it could keep serving.

kazu-yamamoto:
Thank you for your contribution. I have merged your patches.

kazu-yamamoto:
@summerwind I would like to invite you to #404 and this issue. Is it possible to display a warning on 5.1.2?

PR #408, opened 2015-07-28
Add X-Forwarded-Proto to appearsSecure.

mschristiansen:
The X-Forwarded-Proto header is the default for Nginx[1] and is also used by Amazon's Elastic Load-Balancers (which run Nginx)[2].

I've left the other "X_FORWARDED_" headers although I'm not aware in which settings they would be used and suspect they never would.

[1] http://wiki.nginx.org/SSL-Offloader
[2] http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html


creichert:
Great, thanks!

PR #409, opened 2015-07-28
Remove causes of compiler warnings. Use headers from http-types.

mschristiansen:
A lot of small changes to remove causes to compiler warnings.
- Use CPP to only import Monoid and Applicative when less than GHC 7.10
- Rename shadowing variables
- Remove unnecessary imports

Also use header functions from the http-types package where possible and moved OverloadedStrings extension to cabal file, since used in almost every module.

Have compiled and tests passing with both GHC 7.8 and GHC 7.10.


creichert:
Thanks! This looks great, I'm actually pushing up changes  to the travis config to add ghc 7.10 support to our CI.

PR #413, opened 2015-07-29
Change from deprecated `breakByte` to `break`.

mschristiansen:
Small change to avoid deprecated `breakByte` function.


snoyberg:
Thanks!

Issue #403, opened 2015-07-25
Warp sends an empty data frame if no data is immediately available after the headers.


awpr:
This isn't illegal, but it's wasteful, and some clients don't handle it correctly.

This also appears to happen when a streaming application pushes an empty Builder, since it wakes up the waiter but does not write any actual payload to the frame.

kazu-yamamoto:
Are you talking about `ResponseStream`?

kazu-yamamoto:
`Builder` is not an instance of `Eq`. I have no idea to tell a Builder is empty or not.

awpr:
Yes, for ResponseStream.  I have a change stashed currently that makes this work, by checking the amount of data that was actually written by the Builders, and skipping the frame if it was zero bytes.

I initially saw this happen when there was nothing in the output queue after sending headers, so there were no Builders at all.  Then I checked to see what would happen when sending an empty Builder.  So there are two separate causes, but both can be avoided with the same technique: if no data was written, do nothing instead of sending the frame.

kazu-yamamoto:
Thanks. I'm waiting for your pull request.

kazu-yamamoto:
Let's close this.

PR #416, opened 2015-07-30
Avoid sending an empty data frame after headers

awpr:
Previously the data frame header was always written and sent even if
zero bytes were available from the stream; now if zero bytes were
buffered, the data frame will not be sent.  This means we now need to
know whether the headers were already flushed, and possibly flush just
the headers after getting zero bytes of data.

Ameliorates #403.  (Empty data frames can still be sent by yielding a Builder for 0 bytes).

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2/Sender.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/Sender.hs
@@ -10,7 +10,7 @@ import Control.Applicative
 import Control.Concurrent (forkIO)
 import Control.Concurrent.STM
 import qualified Control.Exception as E
-import Control.Monad (void, unless)
+import Control.Monad (unless, void, when)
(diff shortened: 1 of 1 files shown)


awpr:
Good catch, that was wrong.  I fixed it, did some minor refactoring to make it clearer, and also added the same empty-data check to the case of continued streams.

kazu-yamamoto:
Thanks. Merged with refactoring. Please check just in case.

awpr:
Looks good.  I'm not a fan of that if-then-else layout, but it looks like that's the accepted style in this repo.

kazu-yamamoto:
If you want to release a new version of Warp, I will.

awpr:
I don't think this is urgent enough to warrant its own release.  The one client known to misbehave with empty data frames has also already been fixed.

PR #420, opened 2015-08-09
Use bufferIO in place of toBS in Network.Wai.Handler.Warp.IO.

### LINKED ISSUE #419: Warp 3.1.1 fails with "Not in scope: toBS"
Getting this build error when installing latest warp via cabal:

```
Network/Wai/Handler/Warp/IO.hs:50:17: Not in scope: ‘toBS’
```

creichert:
When using blaze-builder > 0.4.

Attempt to fix #419.

@snoyberg @kazu-yamamoto  Looking through the history, this should be sufficient to fix the issue. However, the build failes with more errors. Most of my attempts to fix these result in more widespread conflict issues. Any advice?

```
[29 of 33] Compiling Network.Wai.Handler.Warp.HTTP2.Sender ( Network/Wai/Handler/Warp/HTTP2/Sender.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.2.0/build/Network/Wai/Handler/Warp/HTTP2/Sender.o )

Network/Wai/Handler/Warp/HTTP2/Sender.hs:247:35:
    Couldn't match expected type ‘Data.ByteString.Builder.Internal.Builder’
                with actual type ‘Blaze.ByteString.Builder.Internal.Types.Builder’
    NB: ‘Data.ByteString.Builder.Internal.Builder’
          is defined in ‘Data.ByteString.Builder.Internal’
              in package ‘bytestring-0.10.6.0’
        ‘Blaze.ByteString.Builder.Internal.Types.Builder’
          is defined in ‘Blaze.ByteString.Builder.Internal.Types’
              in package ‘blaze-builder-0.3.3.4’
    In the first argument of ‘B.runBuilder’, namely ‘bb’
    In a stmt of a 'do' block:
      (len, signal) <- B.runBuilder bb datBuf room

Network/Wai/Handler/Warp/HTTP2/Sender.hs:342:47:
    Couldn't match expected type ‘Data.ByteString.Builder.Internal.Builder’
                with actual type ‘Blaze.ByteString.Builder.Internal.Types.Builder’
    NB: ‘Data.ByteString.Builder.Internal.Builder’
          is defined in ‘Data.ByteString.Builder.Internal’
              in package ‘bytestring-0.10.6.0’
        ‘Blaze.ByteString.Builder.Internal.Types.Builder’
          is defined in ‘Blaze.ByteString.Builder.Internal.Types’
              in package ‘blaze-builder-0.3.3.4’
    In the first argument of ‘B.runBuilder’, namely ‘builder’
    In a stmt of a 'do' block:
      (len, signal) <- B.runBuilder builder buf room
```

Also, is the mention of `bytestring` necessary here since it's in both the if else for a cabal flag?
https://github.com/yesodweb/wai/blob/master/warp/warp.cabal#L42

### DIFF
--- a/warp/Network/Wai/Handler/Warp/IO.hs
+++ b/warp/Network/Wai/Handler/Warp/IO.hs
@@ -47,7 +47,7 @@ toBufIOWith buf !size io (Builder build) = loop firstStep
     firstStep = build (buildStep finalStep)
     finalStep (BufRange p _) = return $ Done p ()
     bufRange = BufRange buf (buf `plusPtr` size)
-    runIO ptr = toBS buf (ptr `minusPtr` buf) >>= io
+    runIO ptr = bufferIO buf (ptr `minusPtr` buf) io
     loop step = do
         signal <- runBuildStep step bufRange
         case signal of


creichert:
I'm hoping the CI will also fail with the same issue, I had to make some modification to my local stack.yaml when testing this but I did not check  them in.

creichert:
Looks like it's probably not picking up the code path for the new changes

kazu-yamamoto:
@creichert You should merge your patch. It is a correct fix.

The errors of `Sender.hs` are my fault. I did not support old blaze-builders when I was implementing HTTP/2. But we need to decide whether or not we should support old blaze-builders.

Is supporing bytestring-builder only good enough?

snoyberg:
Yes, I think that's fine. I started looking into this earlier, but got confused about `blaze-builder > 0.4`. @creichert did you mean `blaze-builder < 0.4`?

creichert:
@snoyberg Yes, that was my fault. Merging this now.

snoyberg:
Awesome, thanks for fixing this!

PR #423, opened 2015-08-12
Refactor http2

kazu-yamamoto:
1. With this patches, `Priority` is included in `Stream`. `Prirority` of a stream now can be updated by the priority frame.

2. The old code checks stream window size after dequeuing. If the size is 0, it is enqueued again. This is redundant and inefficient. So, the new code check it before enqueuing.

@awpr @AaronFriel Would you take time to review this patches?

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/Receiver.hs
@@ -316,11 +316,12 @@ stream FrameRSTStream header bs ctx _ strm = do
     closed ctx strm cc
     return $ Closed cc -- will be written to streamState again
 
-stream FramePriority header bs Context{outputQ} s Stream{streamNumber} = do
+stream FramePriority header bs Context{outputQ} s Stream{streamNumber,streamPriority} = do
     PriorityFrame p <- guardIt $ decodePriorityFrame header bs
     checkPriority p streamNumber
-    -- fixme: the case where priority is changed.
-    prepare outputQ streamNumber p
+    -- checkme: this should be tested
+    writeIORef streamPriority p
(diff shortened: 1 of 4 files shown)


awpr:
Looks good.  It might make sense to add another function alongside 'enqueueWhenWindowIsOpen' to check and enqueue immediately or forkIO and enqueue later, since that logic seems to be duplicated between 'maybeEnqueueNext' and 'response'.

Could you add a comment on the new purpose of 'checkWindowSize'?  It took me a while to figure out that it actually did intend to block the sender thread until the connection window was nonzero.  It might actually be clearer now to un-CPS it since it unconditionally calls 'body' now.

kazu-yamamoto:
@awpr Now looks better?

awpr:
Looks good.

A separate issue that wasn't introduced by this pull request but I may as well bring up here: if the sender gets an Output for a closed stream it will terminate rather than just skipping it, because the tail call to 'loop' is inside the 'unlessClosed' in 2 or 3 cases.

kazu-yamamoto:
@awpr Thank you for you review. I will merge this pull request by myself.

For the closed stream issue, would you send a pull request?

AaronFriel:
@kazu-yamamoto I'm sorry I didn't get a chance to look at this, I will defer to @awpr on all things HTTP/2.0. He's got the right ideas and the time, and I'm not yet familiar enough with `wai` and `warp` to comment on big pieces like this.

I am excited to see more HTTP/2.0 support merged in, however!

awpr:
@kazu-yamamoto Working on it -- it was an easy change to make, but my test server depends on my HTTP2Application changes, and it's a time-consuming merge with this pull.

PR #424, opened 2015-08-15
Continue with the Sender loop despite closed streams

awpr:


awpr:
Finally finished rebasing my changes onto the recent priority changes.  I added a hack to close all streams immediately and confirmed that without this change, this made the whole connection end, and with the change, the connection survived.

awpr:
Context: my comment on #423

awpr:
(the diff looks big, but it's basically just indenting a bunch of lines by 4 spaces and moving 'loop' outside of 'unlessClosed').

creichert:
BTW, you can diff with no whitespace using the `w=1` query parameter. https://github.com/yesodweb/wai/pull/424/files?w=1

creichert:
@awpr This is great stuff, thanks a lot

kazu-yamamoto:
@awpr Please separate "layout change only" commits and other commits. If mixed, it's quite hard to review.

kazu-yamamoto:
The point is that `loop` is the outside of `unlessClosed`, right?

kazu-yamamoto:
@awpr Sorry for my previous comment. Your patch does not incude "layout change only".

kazu-yamamoto:
Now I understand the patch completely. Again, sorry for my confusion.

kazu-yamamoto:

PR #425, opened 2015-08-29
Add hashable dependency to all non-Windows platforms

iquiw:
Network.Wai.Handler.Warp.FdCache depends on it.


iquiw:
It seems MultiMap module also needs to be moved.
I will check and update this PR.

iquiw:
I will close this and check it again. Sorry.

Issue #415, opened 2015-07-30
Server push in HTTP/2


kazu-yamamoto:
- WAI API does not allow to support server pushing which can be controlled by WAI applications.
- Jetty has a smart push mechanism with a learning dictionary based on Referer:. E.g. `default.css` is referred by `index.html`. So, if `index.html` is required next time, we can push `default.css`. It's not a good idea to support this by Warp itself. Rather, we should provide a library which WAI applications can use.
- Another idea is cookie. If a client tells which resources are necessary via a cookie, a server can surely push them.

awpr:
One observation: if the separate Http2Application type is accepted, the normal WAI Application will not grow an explicit interface for pushing streams, as Warp's HTTP/2 stream-push handling will be explicit in the Http2Application type.  In this case any of the several proposed "push inference" mechanisms would necessarily take the form of HTTP/2-only middleware.  This would mean it's applicable to natively-HTTP/2 applications as well as normal Applications served over HTTP/2 (by composing it with promoteApplication), but not to pure-HTTP/1 applications.  This seems to be the desirable outcome anyway (since it's nonsense for an application being served over HTTP/1 to attempt to push a stream).

Unfortunately it would require callers to use 'runHttp2 port (smartPush $ promoteApplication myApp) myApp' instead of simply 'run port myApp'.  Maybe Settings could grow an 'upgradeware' field for middleware that's automatically applied to promoted apps, to allow 'runSettings (setPort port $ setUpgradeware smartPush defaultSettings) myApp' instead.

davean:
I generally prefer there be a different interface for HTTP2. I for one can
make far better use of it by custom coding to it then any middleware could
hope to.

It would be easy to provide HTTP and HTTP2 versions for me, and a HTTP to
HTTP2 wrapper appears easy for the cases that don't write custom code. This
approach also trivially allows HTTP2 middleware.

awpr:
@davean We're in agreement -- see https://github.com/awpr/wai/commits/wai_http2 and https://github.com/yesodweb/wai/issues/402.

My comment here was only concerning the interface of an automatic server push middleware that would be implemented against the explicit interface (which would be available only to HTTP/2-aware applications).

awpr:
[...]

PR #426, opened 2015-08-29
Add an HTTP/2-aware application type supporting trailers and pushed streams.

### LINKED ISSUE #402: Support Trailers in HTTP/1.1 chunked encoding and HTTP/2
The HTTP/1.1 chunked encoding and HTTP/2 both allow responses to be followed by a second set of headers, referred to as "trailers".  The warp server and the Network.Wai Response type do not currently allow applications to provide these trailers.

Unfortunately I don't see a clean and backwards-compatible way to add support for this.  Adding an extra constructor to Response is awkward (since it's essentially the same as ResponseStream) and breaks servers.  Changing the ResponseStream constructor and leaving responseStream as an adapter breaks servers and any middleware that assumes a Response can be faithfully encoded as (StreamingBody, Headers, Status).  Depositing trailers in the Vault and hoping the server notices them is awkward.

awpr:
There's a lot of prose in the commit messages if you find yourself wondering about the reasons behind some particular change.

I still need to try the Windows-specific stuff -- I haven't installed GHC on my Windows partition yet and I've heard it's rather difficult.

Fixes #402, relevant to #415.

### DIFF
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,13 +1,20 @@
 {-# LANGUAGE OverloadedStrings, RankNTypes #-}
 {-# LANGUAGE CPP #-}
 module Network.Wai.HTTP2
-    ( Chunk(..)
-    , HTTP2Application
-    , PushPromise(..)
-    , RespondFunc
+    (
+    -- * Applications
+      HTTP2Application
     , Responder
+    , Body
+    , Chunk(..)
+    , BodyOf
+    , PushFunc
     , Trailers
+    -- * Server push
+    , PushPromise(..)
     , promiseHeaders
+    -- * Conveniences
+    , SimpleBody
     , promoteApplication
     , responder
     , streamFilePart
@@ -58,15 +65,16 @@ type Body a = BodyOf Chunk a
 -- 'Network.Wai.StreamingBody' with @c ~ Builder@ and @a ~ ()@.
 type BodyOf c a = (c -> IO ()) -> IO () -> IO a
 
--- | A function given to an 'HTTP2Application' used to send the response.  This
--- is similar to the respond function in 'Network.Wai.Application', but the
--- status and headers are curried, and it passes on any result value from the
--- stream body.
-type RespondFunc = forall a. H.Status -> H.ResponseHeaders -> Body a -> IO a
-
--- | A continuation-passing style function that provides a response by calling
--- the 'RespondFunc', then returns the request's 'Trailers'.
-type Responder = RespondFunc -> IO Trailers
+-- | The result of an 'HTTP2Application'; or, alternately, an application
+-- that's independent of the request.  This is a continuation-passing style
+-- function that first provides a response by calling the given respond
+-- function, then returns the request's 'Trailers'.
+--
+-- The respond function is similar to the one in 'Network.Wai.Application', but
+-- it only takes a streaming body, the status and headers are curried, and it
+-- passes on any result value from the stream body.
+type Responder = (forall a. H.Status -> H.ResponseHeaders -> Body a -> IO a)
+              -> IO Trailers
 
 -- | A function given to an 'HTTP2Application' to initiate a server-pushed
 -- stream.  Its argument is the same as the result of an 'HTTP2Application', so
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -35,23 +35,42 @@ data PushPromise = PushPromise
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
 type HTTP2Application = Request -> PushFunc -> Responder
 
+-- | Part of a streaming response -- either a 'Builder' or a range of a file.
 data Chunk = FileChunk FilePath (Maybe FilePart) | BuilderChunk Builder
 
+-- TODO send -> write
+-- | Given the write function of a stream body, write the given 'Builder'.
 sendBuilder :: (Chunk -> IO ()) -> Builder -> IO ()
 sendBuilder = (. BuilderChunk)
 
+-- | Given the write function of a stream body, write part or all of the given
+-- file.
 sendFilePart :: (Chunk -> IO ()) -> FilePath -> Maybe FilePart -> IO ()
 sendFilePart = (.: FileChunk)
   where f .: g = curry $ f . uncurry g
 
+-- | The streaming body of a response.  Equivalent to
+-- 'Network.Wai.StreamingBody' except that it can also write file ranges and
+-- return a result of type @a@.
 type Body a = (Chunk -> IO ()) -> IO () -> IO a
 
+-- | A function given to an 'HTTP2Application' used to send the response.  This
+-- is similar to the respond function in 'Network.Wai.Application', but the
+-- status and headers are curried, and it passes on any result value from the
+-- stream body.
 type RespondFunc = forall a. H.Status -> H.ResponseHeaders -> Body a -> IO a
 
+-- | A continuation-passing style function that provides a response by calling
+-- the 'RespondFunc', then returns the request's 'Trailers'.
 type Responder = RespondFunc -> IO Trailers
 
+-- | A function given to an 'HTTP2Application' to initiate a server-pushed
+-- stream.  Its argument is the same as the result of an 'HTTP2Application', so
+-- you can either implement the response inline, or call your own application
+-- to create the response.
 type PushFunc = PushPromise -> Responder -> IO Bool
 
+-- | Create the 'H.RequestHeaders' corresponding to the given 'PushPromise'.
 promiseHeaders :: PushPromise -> H.RequestHeaders
 promiseHeaders p =
   [ (":method", promisedMethod p)
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -14,13 +14,17 @@ module Network.Wai.HTTP2
     , streamSimple
     ) where
 
+import           Blaze.ByteString.Builder (Builder)
 import           Data.ByteString (ByteString)
-
-import           Data.ByteString.Builder (Builder)
 import qualified Network.HTTP.Types as H
 
-import Network.Wai (Application)
-import Network.Wai.Internal (FilePart, Request, Response(..), ResponseReceived(..))
+import           Network.Wai (Application)
+import           Network.Wai.Internal
+    ( FilePart
+    , Request
+    , Response(..)
+    , ResponseReceived(..)
+    )
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,11 +1,16 @@
 module Network.Wai.HTTP2
     ( Http2Application
+    , Response(..)
     , Trailers
+    , responseStatus
+    , responseHeaders
     ) where
 
+import           Blaze.ByteString.Builder     (Builder)
+import qualified Data.ByteString              as B
 import qualified Network.HTTP.Types as H
 
-import Network.Wai.Internal (Request, Response)
+import Network.Wai.Internal (Request, FilePart)
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
@@ -13,3 +18,21 @@ type Trailers = H.ResponseHeaders
 
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
 type Http2Application = Request -> (Response -> IO ()) -> IO Trailers
+
+type StreamingBody = (Builder -> IO ()) -> IO () -> IO ()
+
+data Response
+    = ResponseFile H.Status H.ResponseHeaders FilePath (Maybe FilePart)
+    | ResponseBuilder H.Status H.ResponseHeaders Builder
+    | ResponseStream H.Status H.ResponseHeaders StreamingBody
+    | ResponseRaw (IO B.ByteString -> (B.ByteString -> IO ()) -> IO ()) Response
+
+responseStatus :: Response -> H.Status
+responseStatus (ResponseFile s _ _ _) = s
+responseStatus (ResponseBuilder s _ _) = s
+responseStatus (ResponseStream s _ _) = s
+responseStatus (ResponseRaw _ res) = responseStatus res
+
+responseHeaders :: Response -> H.ResponseHeaders
+responseHeaders _ = []
+
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -14,9 +14,14 @@ module Network.Wai.HTTP2
     , PushPromise(..)
     , promiseHeaders
     -- * Conveniences
-    , SimpleBody
     , promoteApplication
+    -- ** 'Responder's
     , responder
+    , respondFile
+    , respondFilePart
+    , respondNotFound
+    -- ** 'Body's
+    , SimpleBody
     , streamFilePart
     , streamBuilder
     , streamSimple
@@ -31,10 +36,13 @@ import qualified Network.HTTP.Types as H
 
 import           Network.Wai (Application)
 import           Network.Wai.Internal
-    ( FilePart
-    , Request
+    ( FilePart(..)
+    , Request(requestHeaders)
     , Response(..)
     , ResponseReceived(..)
+    , adjustForFilePart
+    , chooseFilePart
+    , tryGetFileSize
     )
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
@@ -54,7 +62,7 @@ data PushPromise = PushPromise
 type HTTP2Application = Request -> PushFunc -> Responder
 
 -- | Part of a streaming response -- either a 'Builder' or a range of a file.
-data Chunk = FileChunk FilePath (Maybe FilePart) | BuilderChunk Builder
+data Chunk = FileChunk FilePath FilePart | BuilderChunk Builder
 
 -- | The streaming body of a response.  Equivalent to
 -- 'Network.Wai.StreamingBody' except that it can also write file ranges and
@@ -91,9 +99,49 @@ promiseHeaders p =
   , (":scheme", promisedScheme p)
   ] ++ promisedHeader p
 
--- | Create a response body consisting of a single range of a file.
-streamFilePart :: FilePath -> Maybe FilePart -> Body ()
+-- | Create a response body consisting of a single range of a file.  Does not
+-- set Content-Length or Content-Range headers.  For that, use
+-- 'respondFilePart' or 'respondFile'.
+streamFilePart :: FilePath -> FilePart -> Body ()
 streamFilePart path part write _ = write $ FileChunk path part
+
+-- | Respond with a single range of a file, adding the Content-Length and
+-- Content-Range headers and changing the status to 206 as appropriate.
+--
+-- If you want the range to be inferred automatically from the Range header,
+-- use 'respondFile' instead.
+respondFilePart :: H.Status -> H.ResponseHeaders -> FilePath -> FilePart -> Responder
+respondFilePart s h path part respond = do
+    let (s', h') = adjustForFilePart s h part
+    respond s' h' $ streamFilePart path part
+    return []
+
+-- | Serve the requested range of the specified file (based on the Range
+-- header), using the given 'H.Status' and 'H.ResponseHeaders' as a base.  If
+-- the file is not accessible, the status will be replaced with 404 and a
+-- default not-found message will be served.  If a partial file is requested,
+-- the status will be replaced with 206 and the Content-Range header will be
+-- added.  The Content-Length header will always be added.
+respondFile :: H.Status -> H.ResponseHeaders -> FilePath -> H.RequestHeaders -> Responder
+respondFile s h path req respond = do
+    fileSize <- tryGetFileSize path
+    case fileSize of
+        Left _ -> respondNotFound h respond
+        Right size -> respondFileExists s h path size req respond
+
+-- As 'respondFile', but with prior knowledge of the file's existence and size.
+respondFileExists :: H.Status -> H.ResponseHeaders -> FilePath -> Integer -> H.RequestHeaders -> Responder
+respondFileExists s h path size reqHdrs =
+    respondFilePart s h path $ chooseFilePart size $ lookup H.hRange reqHdrs
+
+-- | Respond with a minimal 404 page with the given headers.
+respondNotFound :: H.ResponseHeaders -> Responder
+respondNotFound h respond = do
+    respond H.notFound404 h' $ streamBuilder "File not found."
+    return []
+  where
+    contentType = (H.hContentType, "text/plain; charset=utf-8")
+    h' = contentType:filter ((/=H.hContentType) . fst) h
 
 -- | Create a response body consisting of a single builder.
 streamBuilder :: Builder -> Body ()
@@ -110,15 +158,25 @@ streamSimple body write flush = body (write . BuilderChunk) flush
 
 -- | Use a normal WAI 'Response' to send the response.  Useful if you're
 -- sharing code between HTTP\/2 applications and HTTP\/1 applications.
-responder :: Response -> Responder
-responder response respond = case response of
-    (ResponseFile s h path part) -> [] <$ respond s h (streamFilePart path part)
+--
+-- The 'Request' is used to determine the right file range to serve for
+-- 'ResponseFile'.
+responder :: Request -> Response -> Responder
+responder req response respond = case response of
     (ResponseBuilder s h b)      -> [] <$ respond s h (streamBuilder b)
     (ResponseStream s h body)    -> [] <$ respond s h (streamSimple body)
-    (ResponseRaw _ fallback)     -> responder fallback respond
+    (ResponseRaw _ fallback)     -> responder req fallback respond
+    (ResponseFile s h path mpart) -> go respond
+      where
+        -- Hooray, forcing things to be universally quantified.
+        go :: Responder
+        go = maybe
+            (respondFile s h path $ requestHeaders req)
+            (respondFilePart s h path)
+            mpart
 
 -- | Promote a normal WAI 'Application' to an 'HTTP2Application' by ignoring
 -- the HTTP/2-specific features.
 promoteApplication :: Application -> HTTP2Application
 promoteApplication app req _ respond = [] <$ app req respond'
-  where respond' r = ResponseReceived <$ responder r respond
+  where respond' r = ResponseReceived <$ responder req r respond
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -7,7 +7,6 @@ module Network.Wai.HTTP2
     ) where
 
 import           Blaze.ByteString.Builder     (Builder)
-import qualified Data.ByteString              as B
 import qualified Network.HTTP.Types as H
 
 import Network.Wai.Internal (Request, FilePart)
@@ -25,14 +24,14 @@ data Response
     = ResponseFile H.Status H.ResponseHeaders FilePath (Maybe FilePart)
     | ResponseBuilder H.Status H.ResponseHeaders Builder
     | ResponseStream H.Status H.ResponseHeaders StreamingBody
-    | ResponseRaw (IO B.ByteString -> (B.ByteString -> IO ()) -> IO ()) Response
 
 responseStatus :: Response -> H.Status
 responseStatus (ResponseFile s _ _ _) = s
 responseStatus (ResponseBuilder s _ _) = s
 responseStatus (ResponseStream s _ _) = s
-responseStatus (ResponseRaw _ res) = responseStatus res
 
 responseHeaders :: Response -> H.ResponseHeaders
-responseHeaders _ = []
+responseHeaders (ResponseFile _ h _ _) = h
+responseHeaders (ResponseBuilder _ h _) = h
+responseHeaders (ResponseStream _ h _) = h
 
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -14,7 +14,7 @@ import qualified Data.ByteString.Lazy as L
 import           Data.ByteString.Builder (Builder)
 import qualified Network.HTTP.Types as H
 
-import qualified Network.Wai.Internal as H1 (Request)
+import Network.Wai.Internal (Request)
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
@@ -29,18 +29,16 @@ data PushPromise = PushPromise
     , promisedHeader :: H.RequestHeaders
     }
 
--- | The type of an HTTP\/2 request: a normal HTTP request and an action to
--- push streams associated with this request.
-type Request = (H1.Request, PushPromise -> Responder -> IO Bool)
-
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
-type Http2Application = Request -> Responder
+type Http2Application = Request -> PushFunc -> Responder
 
 type Body a = (Builder -> IO ()) -> IO () -> IO a
 
 type RespondFunc = forall a. H.Status -> H.ResponseHeaders -> Body a -> IO a
 
 type Responder = RespondFunc -> IO Trailers
+
+type PushFunc = PushPromise -> Responder -> IO Bool
 
 promiseHeaders :: PushPromise -> H.RequestHeaders
 promiseHeaders p =
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -7,17 +7,20 @@ module Network.Wai.HTTP2
     , Responder
     , Trailers
     , promiseHeaders
-    , writeBuilder
-    , writeFilePart
+    , promoteApplication
+    , responder
+    , streamFilePart
+    , streamBuilder
+    , streamSimple
     ) where
 
 import           Data.ByteString (ByteString)
-import qualified Data.ByteString.Lazy as L
 
 import           Data.ByteString.Builder (Builder)
 import qualified Network.HTTP.Types as H
 
-import Network.Wai.Internal (FilePart, Request)
+import Network.Wai (Application)
+import Network.Wai.Internal (FilePart, Request, Response(..), ResponseReceived(..))
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
@@ -38,20 +41,14 @@ type HTTP2Application = Request -> PushFunc -> Responder
 -- | Part of a streaming response -- either a 'Builder' or a range of a file.
 data Chunk = FileChunk FilePath (Maybe FilePart) | BuilderChunk Builder
 
--- | Given the write function of a stream body, write the given 'Builder'.
-writeBuilder :: (Chunk -> IO ()) -> Builder -> IO ()
-writeBuilder = (. BuilderChunk)
-
--- | Given the write function of a stream body, write part or all of the given
--- file.
-writeFilePart :: (Chunk -> IO ()) -> FilePath -> Maybe FilePart -> IO ()
-writeFilePart = (.: FileChunk)
-  where f .: g = curry $ f . uncurry g
-
 -- | The streaming body of a response.  Equivalent to
 -- 'Network.Wai.StreamingBody' except that it can also write file ranges and
 -- return a result of type @a@.
-type Body a = (Chunk -> IO ()) -> IO () -> IO a
+type Body a = BodyOf Chunk a
+
+-- | Generalization of 'Body' to arbitrary chunk types; this unifies with
+-- 'Network.Wai.StreamingBody' with @c ~ Builder@ and @a ~ ()@.
+type BodyOf c a = (c -> IO ()) -> IO () -> IO a
 
 -- | A function given to an 'HTTP2Application' used to send the response.  This
 -- is similar to the respond function in 'Network.Wai.Application', but the
@@ -77,3 +74,35 @@ promiseHeaders p =
   , (":authority", promisedAuthority p)
   , (":scheme", promisedScheme p)
   ] ++ promisedHeader p
+
+-- | Create a response body consisting of a single range of a file.
+streamFilePart :: FilePath -> Maybe FilePart -> Body ()
+streamFilePart path part write _ = write $ FileChunk path part
+
+-- | Create a response body consisting of a single builder.
+streamBuilder :: Builder -> Body ()
+streamBuilder builder write _ = write $ BuilderChunk builder
+
+-- | Equivalent to 'Body' but only streaming 'Builder's.
+--
+-- @'Network.Wai.StreamingBody' ~ SimpleBody ()@.
+type SimpleBody a = BodyOf Builder a
+
+-- | Create a response body of a stream of 'Builder's.
+streamSimple :: SimpleBody a -> Body a
+streamSimple body write flush = body (write . BuilderChunk) flush
+
+-- | Use a normal WAI 'Response' to send the response.  Useful if you're
+-- sharing code between HTTP\/2 applications and HTTP\/1 applications.
+responder :: Response -> Responder
+responder response respond = case response of
+    (ResponseFile s h path part) -> [] <$ respond s h (streamFilePart path part)
+    (ResponseBuilder s h b)      -> [] <$ respond s h (streamBuilder b)
+    (ResponseStream s h body)    -> [] <$ respond s h (streamSimple body)
+    (ResponseRaw _ fallback)     -> responder fallback respond
+
+-- | Promote a normal WAI 'Application' to an 'HTTP2Application' by ignoring
+-- the HTTP/2-specific features.
+promoteApplication :: Application -> HTTP2Application
+promoteApplication app req _ respond = [] <$ app req respond'
+  where respond' r = ResponseReceived <$ responder r respond
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -2,12 +2,9 @@
 module Network.Wai.HTTP2
     ( Http2Application
     , PushPromise(..)
+    , RespondFunc
     , Responder
-    , Response
     , Trailers
-    , responseStatus
-    , responseStream
-    , responseHeaders
     , promiseHeaders
     ) where
 
@@ -41,18 +38,9 @@ type Http2Application = Request -> Responder
 
 type Body a = (Builder -> IO ()) -> IO () -> IO a
 
-type Response a = (H.Status, H.ResponseHeaders, Body a)
+type RespondFunc = forall a. H.Status -> H.ResponseHeaders -> Body a -> IO a
 
-type Responder = (forall a. Response a -> IO a) -> IO Trailers
-
-responseStatus :: Response a -> H.Status
-responseStatus (s, _, _) = s
-
-responseHeaders :: Response a -> H.ResponseHeaders
-responseHeaders (_, h, _) = h
-
-responseStream :: H.Status -> H.ResponseHeaders -> Body a -> Response a
-responseStream = (,,)
+type Responder = RespondFunc -> IO Trailers
 
 promiseHeaders :: PushPromise -> H.RequestHeaders
 promiseHeaders p =
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -11,18 +11,24 @@ module Network.Wai.HTTP2
 import           Blaze.ByteString.Builder     (Builder)
 import qualified Network.HTTP.Types as H
 
-import Network.Wai.Internal (Request)
+import qualified Network.Wai.Internal as H1 (Request)
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
 type Trailers = H.ResponseHeaders
 
+-- | The type of an HTTP\/2 request.
+-- TODO(awpr): specialize this for HTTP\/2.
+type Request = H1.Request
+
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
-type Http2Application = Request -> (forall a. Response a -> IO a) -> IO Trailers
+type Http2Application = Request -> Responder
 
 type Body a = (Builder -> IO ()) -> IO () -> IO a
 
 type Response a = (H.Status, H.ResponseHeaders, Body a)
+
+type Responder = (forall a. Response a -> IO a) -> IO Trailers
 
 responseStatus :: Response a -> H.Status
 responseStatus (s, _, _) = s
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE EmptyCase, EmptyDataDecls, RankNTypes #-}
 module Network.Wai.HTTP2
     ( Http2Application
+    , PushPromise
     , Response
     , Trailers
     , absurd
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -34,7 +34,7 @@ data PushPromise = PushPromise
 
 -- | The type of an HTTP\/2 request: a normal HTTP request and an action to
 -- push streams associated with this request.
-type Request = (H1.Request, PushPromise -> Responder -> IO ())
+type Request = (H1.Request, PushPromise -> Responder -> IO Bool)
 
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
 type Http2Application = Request -> Responder
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -7,8 +7,8 @@ module Network.Wai.HTTP2
     , Responder
     , Trailers
     , promiseHeaders
-    , sendBuilder
-    , sendFilePart
+    , writeBuilder
+    , writeFilePart
     ) where
 
 import           Data.ByteString (ByteString)
@@ -38,15 +38,14 @@ type HTTP2Application = Request -> PushFunc -> Responder
 -- | Part of a streaming response -- either a 'Builder' or a range of a file.
 data Chunk = FileChunk FilePath (Maybe FilePart) | BuilderChunk Builder
 
--- TODO send -> write
 -- | Given the write function of a stream body, write the given 'Builder'.
-sendBuilder :: (Chunk -> IO ()) -> Builder -> IO ()
-sendBuilder = (. BuilderChunk)
+writeBuilder :: (Chunk -> IO ()) -> Builder -> IO ()
+writeBuilder = (. BuilderChunk)
 
 -- | Given the write function of a stream body, write part or all of the given
 -- file.
-sendFilePart :: (Chunk -> IO ()) -> FilePath -> Maybe FilePart -> IO ()
-sendFilePart = (.: FileChunk)
+writeFilePart :: (Chunk -> IO ()) -> FilePath -> Maybe FilePart -> IO ()
+writeFilePart = (.: FileChunk)
   where f .: g = curry $ f . uncurry g
 
 -- | The streaming body of a response.  Equivalent to
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,5 +1,27 @@
 {-# LANGUAGE OverloadedStrings, RankNTypes #-}
 {-# LANGUAGE CPP #-}
+
+-- | An HTTP\/2-aware variant of the 'Network.Wai.Application' type.  Compared
+-- to the original, this exposes the new functionality of server push and
+-- trailers, allows stream fragments to be sent in the form of file ranges, and
+-- allows the stream body to produce a value to be used in constructing the
+-- trailers.  Existing @Applications@ can be faithfully upgraded to HTTP\/2
+-- with 'promoteApplication' or served transparently over both protocols with
+-- the normal Warp 'Network.Wai.Handler.Warp.run' family of functions.
+--
+-- An 'HTTP2Application' takes a 'Request' and a 'PushFunc' and produces a
+-- 'Responder' that will push any associated resources and send the response
+-- body.  The response is always a stream of 'Builder's and file chunks.
+-- Equivalents of the 'Network.Wai.responseBuilder' family of functions are
+-- provided for creating 'Responder's conveniently.
+--
+-- Pushed streams are handled by an IO action that triggers a server push.  It
+-- returns @True@ if the @PUSH_PROMISE@ frame was sent, @False@ if not.  Note
+-- this means it will still return @True@ if the client reset or ignored the
+-- stream.  This gives handlers the freedom to implement their own heuristics
+-- for whether to actually push a resource, while also allowing middleware and
+-- frameworks to trigger server pushes automatically.
+
 module Network.Wai.HTTP2
     (
     -- * Applications
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -92,8 +92,9 @@ data Chunk = FileChunk FilePath FilePart | BuilderChunk Builder
 -- return a result of type @a@.
 type Body a = BodyOf Chunk a
 
--- | Generalization of 'Body' to arbitrary chunk types; this unifies with
--- 'Network.Wai.StreamingBody' with @c ~ Builder@ and @a ~ ()@.
+-- | Generalization of 'Body' to arbitrary chunk types.
+--
+-- 'Network.Wai.StreamingBody' is identical to @BodyOf Builder ()@.
 type BodyOf c a = (c -> IO ()) -> IO () -> IO a
 
 -- | The result of an 'HTTP2Application'; or, alternately, an application
@@ -175,7 +176,7 @@ streamBuilder builder write _ = write $ BuilderChunk builder
 
 -- | Equivalent to 'Body' but only streaming 'Builder's.
 --
--- @'Network.Wai.StreamingBody' ~ SimpleBody ()@.
+-- 'Network.Wai.StreamingBody' is identical to @SimpleBody ()@.
 type SimpleBody a = BodyOf Builder a
 
 -- | Create a response body of a stream of 'Builder's.
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -5,11 +5,11 @@ module Network.Wai.HTTP2
 
 import qualified Network.HTTP.Types as H
 
-import Network.Wai.Internal (Request, Response, ResponseReceived)
+import Network.Wai.Internal (Request, Response)
 
 -- | Headers sent after the end of a data stream, as defined by section 4.1.2 of
 -- the HTTP\/1.1 spec (RFC 7230), and section 8.1 of the HTTP\/2 spec.
 type Trailers = H.ResponseHeaders
 
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
-type Http2Application = Request -> (Response -> IO ResponseReceived) -> IO ResponseReceived
+type Http2Application = Request -> (Response -> IO ()) -> IO Trailers
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -20,9 +20,9 @@ type Trailers = H.ResponseHeaders
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
 type Http2Application = Request -> (forall a. Response a -> IO a) -> IO Trailers
 
-type StreamingBody a = (Builder -> IO ()) -> IO () -> IO a
+type Body a = (Builder -> IO ()) -> IO () -> IO a
 
-type Response a = (H.Status, H.ResponseHeaders, StreamingBody a)
+type Response a = (H.Status, H.ResponseHeaders, Body a)
 
 responseStatus :: Response a -> H.Status
 responseStatus (s, _, _) = s
@@ -30,5 +30,5 @@ responseStatus (s, _, _) = s
 responseHeaders :: Response a -> H.ResponseHeaders
 responseHeaders (_, h, _) = h
 
-responseStream :: H.Status -> H.ResponseHeaders -> StreamingBody a -> Response a
+responseStream :: H.Status -> H.ResponseHeaders -> Body a -> Response a
 responseStream = (,,)
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE RankNTypes #-}
 module Network.Wai.HTTP2
     ( Http2Application
     , Response
@@ -17,17 +18,17 @@ import Network.Wai.Internal (Request)
 type Trailers = H.ResponseHeaders
 
 -- | The HTTP\/2-aware equivalent of 'Network.Wai.Application'.
-type Http2Application = Request -> (Response -> IO ()) -> IO Trailers
+type Http2Application = Request -> (forall a. Response a -> IO a) -> IO Trailers
 
-type StreamingBody = (Builder -> IO ()) -> IO () -> IO ()
+type StreamingBody a = (Builder -> IO ()) -> IO () -> IO a
 
-type Response = (H.Status, H.ResponseHeaders, StreamingBody)
+type Response a = (H.Status, H.ResponseHeaders, StreamingBody a)
 
-responseStatus :: Response -> H.Status
+responseStatus :: Response a -> H.Status
 responseStatus (s, _, _) = s
 
-responseHeaders :: Response -> H.ResponseHeaders
+responseHeaders :: Response a -> H.ResponseHeaders
 responseHeaders (_, h, _) = h
 
-responseStream :: H.Status -> H.ResponseHeaders -> StreamingBody -> Response
+responseStream :: H.Status -> H.ResponseHeaders -> StreamingBody a -> Response a
 responseStream = (,,)
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -1,8 +1,9 @@
-{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE EmptyCase, EmptyDataDecls, RankNTypes #-}
 module Network.Wai.HTTP2
     ( Http2Application
     , Response
     , Trailers
+    , absurd
     , responseStatus
     , responseStream
     , responseHeaders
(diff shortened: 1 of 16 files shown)


awpr:
I realized I didn't actually need to use Windows to try the Handle-based file streaming, since it's about as cross-platform as it gets, so I swapped the ifdefs with ifndefs and did some testing, which uncovered some warnings.

Also fixed compatibility with old versions of 'bytestring' from before 'bytestring-builder' was moved into 'bytestring'.

kazu-yamamoto:
I'm reviewing now. Please give time.

Current my impression is that we need to convince @snoyberg to accept this because this adds a new module `Network.Wai.HTTP2`.

When I fully understand the code, I will summarize pros and cons.

kazu-yamamoto:
I guess that `PushFunc` should be exported from `Network.Wai.HTTP2`.

kazu-yamamoto:
Documentation of `runHTTP2TLS` in Hadock is missing.

kazu-yamamoto:
@awpr What do you think of the following procedure?

- We create a new topic branch for this in this wai repo
- Continue reviewing on the topic branch
- If I find small issues, I will fix them and push the patches to the topic branch
- When we fully understand the code and convince this change, we will merge the topic branch to the master

awpr:
Sure, that sounds reasonable.  I've noticed some potential issues with this FileChunk interface, though:

* It doesn't allow the user to manage the file handle themselves without sacrificing performance (by reading chunks and sending them as Builders to be copied), even though the sender only cares that it can dump a bunch of bytes into its buffer.  In theory we could have all chunks be a 'BufferWriter' (which doesn't promise referential transparency), but then we'd be pushing the responsibility for opening the file (and using an FD cache for performance) onto the user.
* It doesn't provide a good way to have the headers depend on the file contents.  (For a contrived example, imagine an application that opens some kind of indexed archive file, reads the index to determine appropriate offsets for a contained file, and sends a single chunk straight from the file.)  To do this under the current interface, you'd have to open the file once as a Handle, then send a FileChunk that will cause the sender to re-open it as an Fd.

I think these aren't deal-breakers and we can improve them later -- the normal Application can't do this either, so it's not a loss of generality.

I also need to change or replace streamFile so that it can stat the file and add a content-length header.  (Although apparently responseToStream doesn't do that currently either).

awpr:
Added Haddock for runHTTP2TLS and cleaned up export list of Network.Wai.HTTP2.

kazu-yamamoto:
@snoyberg May I add @awpr to the contributors of this repository? I guarantee his skill.

kazu-yamamoto:
The `http2-push` branch is created based.

kazu-yamamoto:
All tests of h2spec are passed. Good. :-)

kazu-yamamoto:
Would do give me examples of `HTTP2Application`? I would like to consider how to support https://github.com/h2o/h2o/issues/421.

awpr:
Sure, I've got a few examples ranging from tame to hangs-Chrome's-HTTP2-session (https://code.google.com/p/chromium/issues/detail?id=526585).  I just need to separate them out into their own file and I'll put them in a Gist.

kazu-yamamoto:
> * It doesn't allow the user to manage the file handle themselves without sacrificing performance (by reading chunks and sending them as Builders to be copied), even though the sender only cares that it can dump a bunch of bytes into its buffer.  In theory we could have all chunks be a 'BufferWriter' (which doesn't promise referential transparency), but then we'd be pushing the responsibility for opening the file (and using an FD cache for performance) onto the user.

I don't understand this. The current code uses `positionRead` and does not create `Builder`s for `SFile`. Would you point the exact line which you talk about?

awpr:
I meant that if the user wants to do something like:

    withFile "/dev/urandom" $ \h -> do
        -- Do something with h before responding, like read part of it and log it, or read a header to determine the offsets to send.
        hGetSome h >>= print
        respond ok200 [] $ \write flush -> do
            write {- Some chunk of 'h' -}
            write {- Some other chunk of 'h' -}

, they have two choices:
- Use 'h' to read into a ByteString, wrap it in a Builder, and send it with 'write $ BuilderChunk b', where it will be copied.
- Use 'write $ FileChunk path someFilePart' and let Warp re-open the file, which would be wasteful.

An alternative would be to allow writing buffer-filling actions ('BufferWriter').

awpr:
https://gist.github.com/awpr/e9aecd949d6546e88688 has some HTTP/2 applications -- it's configured to use SSL certs at /tmp/certificate.pem and /tmp/test_key.pem, and it will try to serve an image out of my home folder, which you can just change to any image.

snoyberg:
Yes that's fine with me.

kazu-yamamoto:
@snoyberg I appeared that I have no right to add new members. Would you add @awpr by yourself?

kazu-yamamoto:
> I meant that if the user wants to do something like:

If a user wants to do this complicated thing, the user should create a builder stream from files, I think.

snoyberg:
No problem, done

kazu-yamamoto:
@snoyberg Thank you.

awpr:
Yeah, it seems like an acceptable limitation.

I just pushed a commit to this branch that brings HTTP2Applications serving plain files up to parity with Applications using ResponseFile, largely by moving a few chunks of the relevant code into wai.  The new dependencies are bytestring-builder (which I assume is fine) and unix-compat.  Is it okay to add this dependency and the utility code to wai, or does it need to be ruthlessly lightweight?

kazu-yamamoto:
Since Warp is dependent on both bytestring-builder and unix-compat, I think it is OK for wai, too.

kazu-yamamoto:
Please push your commits to the `http2-push` branch of the wai repo, not your repo.

awpr:
Done.  I had to merge 2 incarnations of the same 2 changes, amusingly.  (Import Body, categorize Network.Wai.HTTP2 imports)

kazu-yamamoto:
@awpr Would you send an e-mail message to me (kazu iij ad jp)?  I would like to know your e-mail address and explain my personal status in a reply message.

AaronFriel:
Is it too late to propose an API change? I am working on a performance-sensitive web application, and the use of `builders` is a bit of an anathema to performance. 

Why? Because `Builder` ***always*** creates copies when converting to `ByteString`.

Adding an alternative constructor:

```
data Chunk = FileChunk FilePath FilePart | BuilderChunk Builder
```

Is only a partial solution. There is no way to easily write a zero-copy, iteratee/conduit/etc. based webserver. I've been meaning to write and submit a blog post for the perusal of Haskell users on /r/Haskell about this, actually. What I've found in trying to write a high performance web proxy (not a HTTP proxy, but a web proxy) is that the following two things kill Haskell web application performance:

1. `Builder` based code is incredibly widespread, and by default `Builder` creates copies of input strings. Web application frameworks typically involve `Builder` at multiple stages. That means even when the conduit is just doing the identity transform, if it does `lazyByteString . toLazyByteString`, you've now added an extra `memcpy` to your workload.

2. `ByteString` is hostile to appending substrings. Writing a parser that takes small portions of input and then reconstitutes them produces lots, and lots of copies. Each append creates a copy, actually, which causes performance to scale miserably when the appends are contiguous substrings.

Here's an example:

```
-- Two definitions
take1Drop n = take 1 . drop n
x = "Foobar" :: ByteString

x' = take1Drop 0 x
     <> take1Drop 1 x
     <> take1Drop 2 x
     <> take1Drop 3 x
     <> take1Drop 4 x
     <> take1Drop 5 x
```

How many copies does this create? How many bytes are allocated doing this?

The answer is a bit disgusting. On its own, `take1Drop n` allocates a `ByteString`, which is a `PS (ForeignPtr Word8) Int Int`. For the sake of argument we'll conservatively say the `ForeignPtr` is only pointer sized. It actually has its own dynamics in terms of overhead which make the issue more complex, see [Optimizing Garbage Collection Overhead in Sigma](https://simonmar.github.io/posts/2015-07-28-optimising-garbage-collection-overhead-in-sigma.html).

On x64 then, `take1Drop` 8+4+4 bytes, plus a non-garbage collected byte sequence 1 byte long. So 17 bytes plus, thanks to the runtime, a couple cheap allocations.

How many bytes does `take1Drop 0 x <> take1Drop 1 x` allocate?

17 bytes for each term in `<>`, and then because the `Monoid` instance for `ByteString` doesn't optimize for contiguous strings, we allocate another `PS (ForeignPtr Word8) Int Int` and two more bytes on the heap. 52 bytes allocated. Yikes. (Hopefully GHC is optimizing away some of the `PS` constructors and reusing them, but due to `ForeignPtr` I'm not sure if it can!)

How many bytes does `take1Drop 0 x <> take1Drop 1 x <> take1Drop 2 x` allocate?

Same as before, 52 bytes, plus 17, plus another 16 + 3: 88 bytes.

How many bytes does `x'` allocate? 232, assuming conservatively that `ForeignPtr` is just the size of `Addr#`, and liberally that GHC doesn't re-use the constructors (as in a strict fold).

How many `ForeignPtr`s does `x'` allocate? 13.

All to just produce a string denotationally and bitwise-equal to `x`.

I realize this is a late comment after much of the API work has already been done by @awpr, but I would urge hesitation on merging these changes. I am working on seeing if it is possible to make `Builder` not produce copies unless necessary, and I am working on drafting pull requests for my own private branches of `ByteString` and `Text` as well.

I am *not sure* if the changes that suit me will not cause other performance regressions. The changes I've made on my private branches add additional tests to the append logic, those tests will be compiled down to branches, and that means slightly larger binaries and branches can also kill performance quite often.

Given my uncertainty, I ask that we hold off on merging this request, at least for a little while, so that I can explore options for zero-copy sending of streams and produce some benchmarks.

awpr:
I think you're misunderstanding Builders -- they are *exactly* the answer to your point 2, in that they're a value that knows how to write itself directly to a buffer, with constant-time concatenation.  In fact, I believe they were literally designed to address the problem of the quadratic cost of appending ByteStrings.  This problem is well-known in every language, even everyone's favorite, Java (where it manifests as "use StringBuilder, not string + string + string + ...").

Moreover it's simply not true that Builders always copy by default.  The internal representation is quite capable of returning pre-existing chunks of data, and the default is to copy chunks smaller than a threshold, which is a guess at what size of ByteString is worth copying vs. inserting (because it's assumed there's a fixed overhead per chunk, e.g. for the cost of a syscall); there are functions to explicitly make a Builder insert a chunk wholesale regardless of its size.  Whether the driver actually takes advantage of that or not is a totally different issue; currently Warp will copy ByteString chunks into the send buffer for HTTP/2, but it's in theory possible to keep holding onto chunks in a queue until the buffer is full, *then* fill the DATA frame header, *then* send each fragment (whether it's part of the buffer or a ByteString chunk).  But, it doesn't sound like that's the part you're concerned about.

An application that repeatedly executes a Builder to a ByteString and then converts it back to a Builder is using Builders extremely wrong.  For that, the answer is indeed to avoid Builders -- but WAI only requires Builders at the very bottom level, when they're about to be sent.  That imposes nothing on client applications except one (constant-time) call to 'byteString' at the lowest layer of iteratees/conduits/Proxies/Quivers/Machines/whatever-else, at which point (assuming a perfect implementation of Warp) it would not be copied if it's large enough to outweigh the cost of a syscall.  Above that lowest layer, if the application is thrashing between Builders and ByteStrings, it might be more appropriate to just use fixed size strict ByteString chunks, or even use a raw Ptr Word8 buffer directly if it needs to edit the data.

Now I'm curious what kind of proxy this is that's (apparently?) filtering byte-sequence data through multiple layers of conduits before sending it off to Warp.  I'm more than willing to keep working on the API if a more-performant option is discovered -- I'm just not yet convinced that shunning Builders is the answer.

TL;DR: changing from Builder to ByteString at the interface level would be strictly worse, because Builders can already insert existing ByteStrings, but ByteStrings can't coalesce tiny chunks to save on syscalls.

Finally, an aside (because I just noticed the motivating use case for point 2 was parsers): Attoparsec has parser primitives that take a byte count and return in-place slices of the input buffer.  If you're already using them and necessarily need to consume and reassemble small chunks of input, try using Builders for the concatenation instead, since they have constant-time appends.

awpr:
Another quick thought: I suspect the tests you're talking about in the ByteString append functions are something like "append x y | x `isAdjacentTo` y = PS (offset x) (length x + length y) (ptr x)".  If that's the case, the Attoparsec primitive 'take' is probably what you want, but anyway that could just as well be a separate append function that people can use when they know beforehand their inputs are likely to be adjacent.

davean:
Perhaps what he wants is something more like writev? My understanding is
that buffers don't perform vectored IO, but perhaps they could in special
cases.

Its something I meant to look at a while back but I hit performance issues
far earlier then that with other areas of Haskell.

awpr:
Cool, I wasn't aware of writev.  I don't think it would help directly with the parser issue or the thrashing between Builders and ByteStrings, but it's exactly what I'd want for the optimization I mentioned in passing for the HTTP/2 sender.  I even thought it'd be nice if a syscall existed to send multiple buffers at once... (I was trying to go to sleep at the time, so Google wasn't readily at hand)

It seems to me that Builders are extremely well suited to vectored I/O -- doing or not doing vectored I/O isn't a property of Builders themselves but of the code that consumes them.  Obviously 'toLazyByteString' can't do vectored I/O since it's not doing I/O in the first place, but anything that runs Builders directly against a single reused buffer for writing to an FD could probably do it perfectly: run a Builder until it yields a pre-allocated chunk, mark the buffer position and yielded ByteString in a FIFO queue, and repeat until the buffer is full; then flush it all at once with a single writev call (whose buffers would alternate between fragments of the reused buffer and the yielded pre-existing chunks as appropriate).

kazu-yamamoto:
@davean The current Warp is carefully designed to not use `writev()`. We cannot use it because the tls library does not support it.

kazu-yamamoto:
Just FYI: `ByteString` has the following disadvantages:

- Fragmented because of pinned objects
- Using global lock when allocated

For more information, please read http://www.yesodweb.com/blog/2014/02/new-warp

We would be able to use `ShortByteString` in the near future.

kazu-yamamoto:
@AaronFriel 

> ````
> x' = take1Drop 0 x
>      <> take1Drop 1 x
>      <> take1Drop 2 x
>      <> take1Drop 3 x
>      <> take1Drop 4 x
>      <> take1Drop 5 x
> ````

For this kind of examples, we should use `Ptr` to manipulate buffers. This is what Warp is actually doing. See `ResponseHeader.hs` for instance.

`Builder` is a function which copies its inside data to a buffer. You can prepare your own fixed-sized buffer to manipulate I/O. We can use the buffer as `ByteString` with a necessary metadata (i.e. `PS`).

The purpose of `Builder` is postpone the concatenation. You can append or prepend `Builder` with O(1). Then we can concatenate the entire with O(N) finally. (As I said above, you can concatenate the first M bytes and repeat this job if necessary.)

So, we should use `Ptr` in proper place. Also, we should use `Builder` in proper place. I hope that Warp achieves this.

kazu-yamamoto:
Recently, copying data are not costly. We should not be too nervous on this.

Global lock is really costly on multicore. We should not use functions such as `mallocByteString`. My current concern is `cryptonite` which uses global locks a lot.

Network is also slow. So, we should not send separate chunks. Rather, we should pack chunks to a single buffer even if copying is used, then send it at once. `writev()` is nice but it cannot be used in Warp as I explained.

kazu-yamamoto:
@awpr I think I can resume the review tomorrow. On next Sunday, the Japanese HTTP/2 community will have meet up. One of main topic is server push. So, I should understand our code before that. :-)

awpr:
One more change I've made is to apply the "ST" trick to Responder so that it must call the given function or raise (like Application but with a real universally quantified type); then wrap it in a newtype so Responders can be manipulated nicely (without running afoul of impredicative types).  This meant I had to provide trailers as a separate argument to the respond function, and led to a few other convenience functions in Network.Wai.HTTP2; I think the result is actually nicer to program against.  I'm pushing this to https://github.com/awpr/wai/commits/wai_http2_staging since I'm not sure yet which interface is ultimately better.

AaronFriel:
@awpr 

Thank you for the detailed response, I will try to be as complete in my reply. 

First, I would say that my point #2 was about the general topic of writing a blog post on performance-sensitive Haskell. Simply put, `ByteString` as it is is not great, and there's probably room for improvement, but benchmarks are in order. 

> I think you're misunderstanding Builders -- they are exactly the answer to your point 2, in that they're a value that knows how to write itself directly to a buffer, with constant-time concatenation. In fact, I believe they were literally designed to address the problem of the quadratic cost of appending ByteStrings. This problem is well-known in every language, even everyone's favorite, Java (where it manifests as "use StringBuilder, not string + string + string + ...").

I am aware of this, and I realize `Builder` solves point 2. I actually intended for point 2 to be sort of an "aside", that is, a flaw in using `attoparsec` to parse small, iterative chunks of strings. A custom, online, state-based tokenizing and rewriting parser I am running on HTML streams is between 10 and 50 times faster on real-world data, and in large part it is due to inefficiencies in `attoparsec` when trying to adapt it to be an online parser. Specifically, `attoparsec` performs poorly when recursively parsing small tokens (even when those tokens are emitted unaltered) in order to provide eager output of bytes as parsed, or performing state-based or tokenization parsing where you parse according to a spec, e.g.: [HTML5 Tokenization](http://www.w3.org/TR/2011/WD-html5-20110113/tokenization.html).

So for example, if you write a set of `attoparsec` parsers using the HTML5 spec as a guide, at least on my laptop I've noticed that a simple state-based parser that elides or replaces `<!-- -->` comment sections and ignores all other tokens will take many times longer than a custom stream-based, online parser. 

I ran these benchmarks against the front pages of the Alexa top 100, and a random sample of 100 websites from the top 101-1000, and 100 of the top 1001-10000, and so on, until I had 500 websites. Using these sites, I was able to pinpoint the perf issue: and it's `Builder` and `ByteString`. And it's a two order of magnitude improvement if I use a custom `append` and eschew `Builder`.

**But I digress.**

`ByteString` has performance issues. So does `Builder`. 

> Moreover it's simply not true that Builders always copy by default. The internal representation is quite capable of returning pre-existing chunks of data, and the default is to copy chunks smaller than a threshold, which is a guess at what size of ByteString is worth copying vs. inserting (because it's assumed there's a fixed overhead per chunk, e.g. for the cost of a syscall); there are functions to explicitly make a Builder insert a chunk wholesale regardless of its size. Whether the driver actually takes advantage of that or not is a totally different issue; currently Warp will copy ByteString chunks into the send buffer for HTTP/2, but it's in theory possible to keep holding onto chunks in a queue until the buffer is full, then fill the DATA frame header, then send each fragment (whether it's part of the buffer or a ByteString chunk). But, it doesn't sound like that's the part you're concerned about.

An API that accepts `ByteString` chunks (even if they're the lazy variant) will have an easier time of managing allocations than one that accepts `Builder`. Simply because no matter what, to actually transmit a byte, the `Builder` needs to be converted. And `Builder` is not well tuned for HTTP sized pieces of data. How often in an HTTP/2.0 call do you send more than a few thousand bytes within a single frame? Is it really necessary to concatenate two 5KiB `ByteString`s? When I look at my HTTP/2.0 net statistics with Google and Facebook, I rarely see frames where more than 1500 bytes are sent. It looks like their servers are optimizing for the size of MTU, but that may be coincidental.

That said, `Builders` do always copy by default. 

```
toLazyByteString = toLazyByteStringWith
    (safeStrategy L.smallChunkSize L.defaultChunkSize) L.Empty

safeStrategy firstSize bufSize =
    AllocationStrategy nextBuffer (sanitize bufSize) trim
  where
    trim used size                 = 2 * used < size
    {-# INLINE nextBuffer #-}
    nextBuffer Nothing             = newBuffer $ sanitize firstSize
    nextBuffer (Just (_, minSize)) = newBuffer minSize

-- | Allocate a new buffer of the given size.
{-# INLINE newBuffer #-}
newBuffer :: Int -> IO Buffer
newBuffer size = do
    fpbuf <- S.mallocByteString size
    let pbuf = unsafeForeignPtrToPtr fpbuf
    return $! Buffer fpbuf (BufferRange pbuf (pbuf `plusPtr` size))
```

There are lots of places in `Builder` where allocations can happen, but for many use cases, they amount to a high constant factor. For many use cases, `memcpy` is fast enough. What I found is that there is significant overhead in even converting from `ByteString` to `Builder` to `ByteString`. Really, though, that should be essentially a no-op. But it's not, and `Builder` poorly optimizes for the small string case.

> An application that repeatedly executes a Builder to a ByteString and then converts it back to a Builder is using Builders extremely wrong. For that, the answer is indeed to avoid Builders -- but WAI only requires Builders at the very bottom level, when they're about to be sent. That imposes nothing on client applications except one (constant-time) call to 'byteString' at the lowest layer of iteratees/conduits/Proxies/Quivers/Machines/whatever-else, at which point (assuming a perfect implementation of Warp) it would not be copied if it's large enough to outweigh the cost of a syscall. Above that lowest layer, if the application is thrashing between Builders and ByteStrings, it might be more appropriate to just use fixed size strict ByteString chunks, or even use a raw Ptr Word8 buffer directly if it needs to edit the data.

I agree. But `Builder` is prolific in Haskell. People have been conditioned (through blog posts, articles on performance, etc.) to use Builder in lots of APIs. But Builder should almost never be used unless you're *actually* serializing something where the overhead of sending your average chunk size is significant. We agree on this! This is related to point #2: the conventional wisdom is that appending ByteStrings is always bad (by virtue of the API making it so), and people are told to use `Builder` always. 

**But I digress again*. This is just ranting about the proliferation of APIs that harm performance, not about WAI/Warp.

I think that if we're going to use `Builder`, some serious consideration should be given to providing guidelines for how to provide input to Warp/WAI that minimizes copying. 

> Now I'm curious what kind of proxy this is that's (apparently?) filtering byte-sequence data through multiple layers of conduits before sending it off to Warp. I'm more than willing to keep working on the API if a more-performant option is discovered -- I'm just not yet convinced that shunning Builders is the answer.

An HTML/CSS/JS rewriting web proxy. I'm working on an API to reduce the amount of copies, which may make some of this advice obsolete, but the current situation is not great for small `ByteString` data. And particularly for things like JSON APIs, which frequently return payloads <1024 bytes. Do a lot of those, and `Builder` guarantees you pay for an additional copy every time.

> TL;DR: changing from Builder to ByteString at the interface level would be strictly worse, because Builders can already insert existing ByteStrings, but ByteStrings can't coalesce tiny chunks to save on syscalls.

How-so? `ByteString` provides lower level control of the data. You can *always* write a wrapper around a `ByteString` API that takes `Builder`s and converts it to `ByteString`. A buffering wrapper is not difficult.

But it takes much more care to create a `Builder` based API that ensures small data isn't needlessly copied. The worst case scenario for a `ByteString` API is that more syscalls than necessary are made, which do have a perf cost, but with HTTP/2.0 that's almost a cost you have to pay. Again, based on some quick preliminary data from Facebook and other HTTP/2.0 appliations, the data size in frames is often small, and paying for an extra copy on every small string is unnecessary.

> Finally, an aside (because I just noticed the motivating use case for point 2 was parsers): Attoparsec has parser primitives that take a byte count and return in-place slices of the input buffer. If you're already using them and necessarily need to consume and reassemble small chunks of input, try using Builders for the concatenation instead, since they have constant-time appends.

`Builder` isn't contiguous-string aware, and parsing small tokens of input, again, kills performance on real-world benchmarks. I've tried `Builder`, it was not only less deterministic in performance, but many times slower. I'm not sure if `Builder` can be made to be aware of contiguous strings without a significant rewrite, but I haven't spent near as much time looking at that code as it was much easier to get better perf by not using it.

> Another quick thought: I suspect the tests you're talking about in the ByteString append functions are something like "append x y | x isAdjacentTo y = PS (offset x) (length x + length y) (ptr x)". If that's the case, the Attoparsec primitive 'take' is probably what you want, but anyway that could just as well be a separate append function that people can use when they know beforehand their inputs are likely to be adjacent.

`take` is alright, I ended up implementing a type class for `Mergable` types, but I'm now looking at using `IfCxt` to see if I can just write my own wrapper for fast zero-copy appends. It's not a performance panacea, either, as I'm writing an online parser. `attoparsec` just doesn't behave well when you're trying to partially parse data. It supports incremental input, but not incremental output.

> Recently, copying data are not costly. We should not be too nervous on this.
>
> Global lock is really costly on multicore. We should not use functions such as mallocByteString. My current concern is cryptonite which uses global locks a lot.

@kazu-yamamoto I am not sure if this is the case. I find that copies are the #1 cause of performance ailments in my code, whether it's because `PS` constructors aren't being intelligently reused, or a variety of other things.

I'm wondering now: `ByteString` and `Text` use different methods, respectively, to allocate and manage their bytes. Does `Text` suffer the same global lock problem as `ByteString`? I know `Text` uses a `ByteArray#`, whereas `ByteString` uses a `ForeignPtr Word8`, and I believe on a surface level they at least rely on different functions to do allocations. Do they both incur the global lock?

Moreover, why is it that a `ByteString` API surface for Warp/WAI necessarily incurs additional locking? Ultimately `ByteStrings` must be involved somewhere, and whether they're created by the response handler or produced by the request body, they end up somewhere. `Builder` just obfuscates where the allocations are occurring. It seems like it wouldn't be difficult to write a `Builder` wrapper around a `ByteString` response API.

But it seems like it would be strictly more difficult to reason about allocations when using a `Builder` API. If the user passes in 1-byte `ByteString`s, that is on them. But are small JSON payloads and MTU-sized chunks of HTML really worth duplicating in memory and thrashing your cache?

kazu-yamamoto:
@awpr The doc of `Network.Wai.HTTP2` has `~`.  I don't understand what it means.

kazu-yamamoto:
> I'm wondering now: `ByteString` and `Text` use different methods, respectively, to allocate and manage their bytes. Does `Text` suffer the same global lock problem as `ByteString`? I know `Text` uses a `ByteArray#`, whereas `ByteString` uses a `ForeignPtr Word8`, and I believe on a surface level they at least rely on different functions to do allocations. Do they both incur the global lock?

`ByteString` uses pinned objects because it is designed to interact with foreign C code. GC must not move `ByteString` since C code cannot tell the move. Allocation of `ByteString` takes a global lock if the size is large enough. If small, a global lock is not took.

`Text` uses arrays. It is movable. And it does not take a global lock.

For Warp, `ShortByteString` is desirable. It uses arrays and size of metadata is minimum. We can use it with GHC 7.8 or later.

> Moreover, why is it that a `ByteString` API surface for Warp/WAI necessarily incurs additional locking? Ultimately `ByteStrings` must be involved somewhere, and whether they're created by the response handler or produced by the request body, they end up somewhere. `Builder` just obfuscates where the allocations are occurring. It seems like it wouldn't be difficult to write a `Builder` wrapper around a `ByteString` response API.

Only Simon Marlow knows why large `ByteString` takes a global lock. I don't understand the reason yet.

> But it seems like it would be strictly more difficult to reason about allocations when using a `Builder` API. If the user passes in 1-byte `ByteString`s, that is on them. But are small JSON payloads and MTU-sized chunks of HTML really worth duplicating in memory and thrashing your cache?

I don't understand this question. My point is that copying is cheap recently. We should tackle other bottlenecks first. If copying is a bottleneck, we should consider how to avoid copying.

If you see `Recv.hs`, you can find some techniques to avoid copying for input data.

AaronFriel:
Copying is usually cheap, and allocations usually are too (in Haskell) but not for `ByteString`. If you use a `Builder` API, and small strings must be converted to `Builder` before being passed to Warp/WAI, then you end up doing an extra allocation of a `ByteString` in the common case of short strings. Users are not going to use the appropriate API to avoid that.

With HTTP/2.0, I suspect it is going to be the *uncommon* case that we can send large, greater than 4KiB, strings down the wire. That means *every* call to Warp/WAI is going to incur additional allocations and copies.

Those allocations and copies *are not* cheap, in that every `ForeignPtr` produces significant GC overhead (see: [Optimizing Garbage Collection Overhead in Sigma](https://simonmar.github.io/posts/2015-07-28-optimising-garbage-collection-overhead-in-sigma.html)) and even the constructor is not cheap.

The runtime guarantees that a `ByteArray#` passed to a foreign function call will not be moved for the duration of the call. I'm trying to figure out if that means it pins it using a global lock, or if it can use a thread-local lock to prevent the GC from being invoked or moving the array during the call.

If it turns out that `ByteArray#` solves the problem, I would suggest that instead, in an `Internal` API, and a `ByteString` or `Builder` wrapper around that. In HTTP/2.0, I don't think we can get around the fact that there will often be short chunks of data sent via a syscall. Syscalls aren't cheap, but they aren't that bad. At least, not as bad as incurring many more allocations and causing much greater GC pressure. And it would prepare Warp/WAI for use in systems where syscalls are cheap, be they rump kernels, HalVM (akin to Mirage), or even when using usermode networking on Linux, e.g.: with Intel's Data Plane Development Kit or SolarFlare's OpenOnload.

There are a lot of ways to reduce the cost of a syscall, and the number of ways to do that is growing. 

There are not a lot of ways to reduce the cost of additional allocations and copies.

We can't really expect to do great on the syscall story when HTTP/2.0 is a protocol designed to produce highly fragmented and interleaved streams of data. But we *can* ensure that Haskell can use strong typing to minimize allocations and copies.

And sure, if we're sending <250 bytes on the wire, it might be worth it for the internals to use a `Builder` to produce a single string containing the entire frame. I suspect `Data.ByteString.Builder` is not the write tool for this though. Proliferating `ByteString`s is costly, and maybe a `ByteArray#` based builder is a better way to go about this.

awpr:
There are a lot of separate topics going on here at once, so I'm going to respond to them separately; let's get the Attoparsec aside out of the way first:

Okay, to summarize: you have an attoparsec Parser that tokenizes HTML and passes a large majority of it directly through to the output byte-for-byte, but occasionally changes parts (like dropping comments or whitespace, or rewriting links, etc).  It's currently implemented as a Parser that eats tokens with, e.g. 'char' or 'string' and concatenates them back into the result with X, where X is either Builders, the normal ByteString append, or the magic adjacency-checking ByteString append function; the fastest of the three is the magic one, but they're all slow.  Is this all correct?

Assuming it is: yeah, that's a bad place to use Builders to reassemble the input.  Can you just avoid concatenating at all?  You already have (most of) the output in the form of large chunks of ByteString.  It sounds like the parser is only needed to keep track of the state of the tokenizer and find parts that need to be rewritten.  If it can parse the tiny fragments and discard them while adding up their length, it can just yield big chunks of the input without any copying.

Attoparsec has 'scan' that looks like it was made for this exact use; or you could try "parsing" into an integer that represents the number of bytes that should be sent directly from the input, and go back and slice the input ByteString.

Regarding incremental output, I don't think any special support is needed there; just complete one Parser, use the result, and start the next.  I've even done just this recently, and wrapped it up in a function that adapts it to any streaming monad that you can pass in yield/await actions for.  That might be worth packaging in its own library, because it turns out it works for any Parser, not just the one I wrote it for.

awpr:
Next up: the claim that ByteStrings and Builders have performance issues.

This is too general a statement.  ByteStrings and Builders are both good at some things and bad at others, like any other construct.  It's like saying for-loops have performance issues because a program that nests 5 for-loops over millions of elements takes hours to run.  This same issue is true in any language and any API: ripping apart HTML into tokens and putting it back together by either repeatedly appending strings or copying it into a new buffer will be slow.  Reusing the input buffer wholesale will be fast; do that.  (Whether that's by explicitly slicing the buffer given a length, or by magically combining adjacent ByteStrings doesn't really matter, although I suspect the former will be faster).

> converting from ByteString to Builder to ByteString [...] should be essentially a no-op.

Which ByteString are you referring to?  For a strict ByteString, of large enough size via 'byteString', or of any size via 'insertByteString', I agree it should be a no-op, and I would be astonished if it weren't.  For a lazy ByteString with small chunks, it is not a no-op, because Builders are all about spraying data into large contiguous buffers.  If that's not what you want for a particular use, don't use them for that use.  But that *is* what we want for the bottom level of a web server API. 

> Builder poorly optimizes for the small string case.

I'm not sure what you mean.  It does copy small strings, but that's an optimization for anything that actually moves that data through the kernel.  If you want to move lazy ByteStrings of small chunks around unchanged, then don't copy them by calling 'toLazyByteString . lazyByteString'.

> That said, Builders do always copy by default.

This shows that 'toLazyByteString' copies all chunks (and, in fact, it doesn't; it still includes yielded chunks directly in the lazy ByteString without copying them -- see ciosUnitToLazyByteString).  "Builders always copy by default" is a meaningless statement because the Builder does not control whether the chunks it yields are copied.  Always copying or not is a property of the code consuming the Builder.

> Builder should almost never be used unless you're actually serializing something where the overhead of sending your average chunk size is significant. We agree on this!

Yeah, we're in agreement here.  But that characterizes everything that gives the data to the kernel via a syscall, which is almost everything.  If you're building string data internally to be re-parsed within the same process, the tradeoffs are different.  There's still overhead per chunk (the lazy ByteString constructor, the cost of chasing a pointer), but it's much, much smaller.  That would be a cool blog post!

awpr:
Finally, the actual matter we meant to discuss: whether WAI should use Builder in the interface.

> small strings must be converted to Builder before being passed to Warp/WAI, then you end up doing an extra allocation of a ByteString in the common case of short strings.

This is just not true -- allocations are not an issue here; Warp has one big buffer that it runs Builders on, sends the buffer, and starts over on the same buffer.  You're arguing quite effectively against something Warp already does not do: "accept Builder; call toLazyByteString; send that".  Sending ByteStrings instead would have *worse* allocation behavior, because we wouldn't be able to reuse memory between writes.

The problem with accepting ByteStrings instead is that every data frame has to have a 9-byte header.  If we accept only pre-existing ByteStrings, we're faced with a choice: send the 9-byte header in its own write call followed by the data, or copy the data into a buffer together with the header and send it all at once.  It's true that syscalls aren't the end of the world, but doubling the number of write calls just to avoid Builders seems like a bad idea.

> It takes much more care to create a Builder based API that ensures small data isn't needlessly copied.

This is also misleading; it's not needless.  If you give us 1,000 2-byte buffer fragments, it's just plain wrong to send each one with 2 write calls (header then data).  Even if we knew ahead of time about all 1,000, it's still wrong to make 1,001 write calls (one header; 1,000 data).  It's better to copy it into our (already existing!) buffer and send that.

If you have special knowledge that your many fragments are contiguous, then the answer is to send them as one big fragment instead, so the Builder will yield the whole chunk at once.  This happens *by default* with 'byteString' for fragments > 256 bytes, and you can opt in explicitly for smaller fragments by using 'insertByteString' instead.  Currently Warp will copy the yielded chunk anyway, but that's an implementation detail; arguably it'd be better to eat the double write call (header; data) when the fragment is bigger than some threshold.  That's totally orthogonal to whether we use Builders as the interface, though.

The short version is: assuming the sender implementation trusts the Builder when it yields existing buffers, Builders do everything you're asking for *and more*.  If the application just wants to do what's typically best for using existing string fragments, it can use the default size-thresholding behavior of 'byteString'; if it has special knowledge, it can exert low-level control over the sender's write calls by using 'insertByteString'.

As for "and more", Builders are also capable of fabricating bytes from thin air instead of from existing buffers -- such as for writing integers as decimal, hexadecimal, or network byte order; for these cases they will not copy the data *at all* but write it directly to the buffer it's going to be sent from, while a ByteString based API would require the caller to make a tiny ByteString, and then internally copy that tiny ByteString into the larger send buffer.

AaronFriel:
This is just a reply to our discussion on `attoparsec`, so for other passive readers, it's not relevant to the API discussion.

> Assuming it is: yeah, that's a bad place to use Builders to reassemble the input.  Can you just avoid concatenating at all?  You already have (most of) the output in the form of large chunks of ByteString.  It sounds like the parser is only needed to keep track of the state of the tokenizer and find parts that need to be rewritten.  If it can parse the tiny fragments and discard them while adding up their length, it can just yield big chunks of the input without any copying.

Yes, and that mostly works. `attoparsec` leaves a lot to be desired here for producing incremental output. `scan` is interesting, but again, I find it lacking for my purposes. There is yet one more reason I am not using it: it is not easy to compose parsers.

Suppose I have a `JavaScript` parser, designed for handling the byte stream of a `.js` file. How do I embed that in a `HTML` parser? What I really want is for the parser to understand that it can apply the rules of the embedded parser until it reaches a certain state - be it `</script>`, or the curious behavior of `<!-- -->` inside a script tag.

There are a number of reasons I've decided not to use `attoparsec`, but to summarize them all, it isn't designed for my use case. It's very good at handling binary streams where you want to produce some data structure, where the size of each state (in terms of # of bytes grabbed) is high, and the transformation you want to perform is fixed. It is not designed to produce incremental output where the chunk size is very small, and embedding another parser within another is also not one of its design criteria.

This is all an aside, but I appreciate the advice and questions. 

> Regarding incremental output, I don't think any special support is needed there; just complete one Parser, use the result, and start the next.  I've even done just this recently, and wrapped it up in a function that adapts it to any streaming monad that you can pass in yield/await actions for. That might be worth packaging in its own library, because it turns out it works for any Parser, not just the one I wrote it for.

I did something similar with my `iota` library. A rather old version is on GitHub. I'm on my 5th rewrite of it, and I've been able to get drastically higher performance with each rewrite. I'm unsure if I'll release the final version, to be honest, as it's a lot of work and it will form an integral part of my startup. I think I have a design now that allows truly astounding flexibility, concurrency (tokenizing a 500KiB HTML5 document by running several parsers concurrently and patching them together, as in `trifecta`), and optimization, understanding a parser as being a patchwork of automata of different degrees of capability, e.g.: for a simple HTML5 tokenizer, it's just a DFA. A parser that interprets HTML tokenization states (TagName, AttributeValue, etc.) and rewrites them is just a DFA. Those things can be composed together, and with Template Haskell, it can be rewritten as a minimal state machine that erases the tokens, so that the final parser produces no intermediate allocations.

awpr:
Okay, I think we understand each other on Attoparsec.  Obviously you know a lot more about this problem than I do, but the conclusion seems to be that Attoparsec isn't well-suited to yielding tons of tiny fragments, and ByteStrings aren't well-suited to reconstituting tons of tiny adjacent fragments into a big buffer.  It sounds like instead you're using ByteStrings to represent mergeable fragments and getting huge performance gains -- this is a cool technique that I've not seen implemented in Haskell before, so maybe it'd make sense to make that concept official like "data ByteFragment = BF Int Int" (with the actual pointer not included), and provide a function to turn a big ByteString and a sequence of fragments into the minimal number of ByteStrings.  That seems like a moderately common case that's not handled well by existing libraries.

AaronFriel:
> This is just not true -- allocations are not an issue here; Warp has one big buffer that it runs Builders on, sends the buffer, and starts over on the same buffer. You're arguing quite effectively against something Warp already does not do: "accept Builder; call toLazyByteString; send that". Sending ByteStrings instead would have worse allocation behavior, because we wouldn't be able to reuse memory between writes.

This is something I wasn't aware of. It very much changes the dynamics, and if you're able to reliably reuse that buffer, it changes the performance dynamics significantly. I apologize for speaking from ignorance on this issue.

My understanding of `Builder` was that you essentially always have to produce a `ByteString` as a result. That produces additional copies, and additional `ForeignPtr`s which have their own costs.

If `Builder`, as you say, minimizes this, then perhaps that's what would be best. I'm still concerned that an API which only accepts `Data.ByteString.Builder` means all input to 
WAI has to pay the ForeignPtr tax of `ByteString`, and as I've found, `Builder` intermediates and inefficient appends are a performance anathema - at least in my - I think we can agree - unusual use case.

> If you have special knowledge that your many fragments are contiguous, then the answer is to send them as one big fragment instead, so the Builder will yield the whole chunk at once. This happens by default with 'byteString' for fragments > 256 bytes, and you can opt in explicitly for smaller fragments by using 'insertByteString' instead. Currently Warp will copy the yielded chunk anyway, but that's an implementation detail; arguably it'd be better to eat the double write call (header; data) when the fragment is bigger than some threshold. That's totally orthogonal to whether we use Builders as the interface, though.

I agree, I am merely arguing that in the case where moderately sized bytestrings are produced, the syscall overhead should be thoroughly examined. It may be a premature optimization to write to a buffer (incurring an additional copy) to avoid an extra syscall, and it may even be wise to make this a tunable parameter. If I am using a SolarFlare NIC, for example, OpenOnload will make `write` calls to the NIC very cheap - it implements in hardware what `Builder` is trying to do in software, and it doesn't make sense for us to duplicate that effort. An API that allows us to say "Trust me, this ByteString is large enough, or the call is cheap enough, that you shouldn't copy this to avoid extra `write`s" is going to be more flexible than one that requires us to use a `Builder`.

A Builder based API that wraps around an internal `ByteString` (or `ByteArray#`) API can implement the buffer you speak of and more. 

A `ByteString` based API that wraps around `Builder` has no way of communicating "Do not waste time copying me" down below. Although if we make it a tunable parameter when setting up WAI, or perhaps a parameter of the `Response`, we can fix that.

AaronFriel:
As an aside from this conversation, I think this weekend I am going to see if it's worthwhile to change `ByteString` from `ForeignPtr` to `ByteArray#`. To avoid breakage with any other libraries that depend on pinned memory, it seems like I could use `newPinnedByteArray#` to get the same behavior, but I could add a `Data.ByteString.Unpinned` module that permits creating unpinned chunks. 

If the global lock on pinned memory is such a significant issue, and we know for our use case that the `write` call doesn't require accessing the array once it's executed, it would be better to pass it a `ByteArray#` and we could avoid the global lock, *and* if we have a parameter that lets us set the `Response` to avoid copies for large-enough chunks, we should have a clear performance win.

awpr:
> ForeignPtr tax of ByteString

I'm not sure about this -- my intuition is that the big cost of ForeignPtrs is mostly associated with their finalizers, and ForeignPtrs that don't have a finalizer will have much lower cost.  The send buffer that we're writing to spends most of its life as a 'Ptr Word8', and is only briefly wrapped in a finalizer-less ForeignPtr when we call the actual write function.  Intuition is often wrong, though -- I should defer to Kazu for deeper knowledge of this.

> say "Trust me, this ByteString is large enough"

Builders can do this; it's called 'insertByteString'.

> A Builder based API that wraps around an internal ByteString (or ByteArray#) API can implement the buffer you speak of and more.

It can't, because the ByteString-based API would not be able to get free concatenation with the frame header when the data is being fabricated from thin air; a buffering wrapper would have to write the Builder to its buffer, then rely on Warp to concatenate it with the header some other way.  It sounds like in your case that would cost nothing, but in general that's not the case, and using Builders addresses both uses.

> A ByteString based API that wraps around Builder has no way of communicating "Do not waste time copying me" down below.

'insertByteString'.  Actually honoring this in the implementation is a separate matter, but Builders do have this information.

awpr:
> better to pass it a ByteArray# and we could avoid the global lock

I think there's some more confusion here: the global lock is for *allocating* pinned memory, right?  But we don't do that; there's no pinned memory allocation just to wrap up an existing Ptr Word8 into a ByteString, which is what happens when Warp HTTP/2 writes its buffer.

I don't know enough about GHC internals to comment authoritatively on changing ByteString itself, but I suspect people will object on the grounds that this would change the contract of the ByteString type (namely that its buffer location won't change out from under it).

AaronFriel:
> > say "Trust me, this ByteString is large enough"
> Builders can do this; it's called 'insertByteString'.

But as you said, there's no way to communicate this to WAI. Yes, I know that means two `write`s, but I really think it's important that the API support expressing this. If I'm sending a 128KiB chunk down the wire, there will be a lot more than two system calls to send it (with HTTP/2.0).

What I would like is for the API to support expressing, "Yes, I'm really sure that I want WAI to not buffer this separately." That could be done with `Builder`, I suppose, but it needs to be a parameter that can be tuned. As you wrote, 

> Currently Warp will copy the yielded chunk anyway, but that's an implementation detail; arguably it'd be better to eat the double write call (header; data) when the fragment is bigger than some threshold. That's totally orthogonal to whether we use Builders as the interface, though.

My response is then: fine, that should be something that the user can opt into. If that means I have to be careful about using `insertByteString` *and* using the appropriate WAI parameters, I won't think it's ideal, but I do think it's worthwhile finding the right sizes for both. And perhaps that means getting some real world telemetry and finding the right sizes for bytestrings that shouldn't copy by default, and how large of a chunk WAI should decide to send as two separate `write`s.

> I'm not sure about this -- my intuition is that the big cost of ForeignPtrs is mostly associated with their finalizers, and ForeignPtrs that don't have a finalizer will have much lower cost. The send buffer that we're writing to spends most of its life as a 'Ptr Word8', and is only briefly wrapped in a finalizer-less ForeignPtr when we call the actual write function. Intuition is often wrong, though -- I should defer to Kazu for deeper knowledge of this.

I'd have to dig into this too, the way I read it is that the GC still has to do extra work on the nursery that's linear in the number of `ForeignPtr`s, though now I'm not sure.

> > better to pass it a ByteArray# and we could avoid the global lock
> I think there's some more confusion here: the global lock is for allocating pinned memory, right? But we don't do that; there's no pinned memory allocation just to wrap up an existing Ptr Word8 into a ByteString, which is what happens when Warp HTTP/2 writes its buffer.

The motivation for changing WAI in the past has been to avoid the global lock. If applications that are using the API are causing ByteStrings to be created which are *unnecessarily* pinned by default, then that still negatively impacts WAI performance, but it's something Haskell applications would benefit from if it existed.

> I don't know enough about GHC internals to comment authoritatively on changing ByteString itself, but I suspect people will object on the grounds that this would change the contract of the ByteString type (namely that its buffer location won't change out from under it).

I'm also unconvinced now that the memory needs to be pinned in many cases. I'm trying to think of what sort of libraries that rely on that contract, and it occurs to me that the number is probably small enough. I agree, though, that for backwards compatibility, the default functions for `ByteString` should remain using pinned memory. 

But if the cost of pinning memory is a global lock that isn't necessary in the majority of use cases, then that is an unnecessary additional cost to using ByteString.

My stance is not that I think `ByteString` and `Builder` are always bad, I think they have done a lot of work on exploring the domain of efficient string representations. But I think there is an over-reliance on them for performance sensitive APIs where the assumptions implicit in `ByteString` and `Builder` do not hold up. For example, that strings will never be contiguous (and so appends should always produce copies) and that it is almost always cheaper to copy memory than to return a small chunk (where small has been arbitrarily defined as less than 4 KiB.) 

My hesitance on urging the HTTP/2.0 patches to be merged into WAI are based on this: I think there are some assumptions being made which should be carefully considered. Are we making users pay for unnecessary copying? It sounds like with the buffer based approach, it is mandatory that Builders will create copies. That's fine, but are we creating *two* copies in the case of strings less than 4 KiB? This is something I'll have to trace through the Builder library to check, but it looks like < 4KiB chunks written to a builder are copied even if the desired end result is to write them to a third buffer.

Another assumption is that it's always cheaper to allocate and copy memory than to perform two syscalls. There are many networking solutions, which are becoming more popular, that significantly change that dynamic. So even if we can't agree on a size, I think it should be a tunable parameter.

Regarding changing `ByteString`, if I were to develop a patch and ask them to merge it, it would be to switch the constructor to use `ByteArray#` directly and add a new module which allows permits creating unpinned ByteStrings. In other words: maintain the current contract, but add an option for unpinned memory, which I think will work well for the majority of FFI cases.

awpr:
> chunks written to a builder are copied even if the desired end result is to write them to a third buffer

Actually... it suddenly occurs to me that maybe you're mixing up Builder with something like Java's StringBuilder.  A Builder isn't a mutable/growable buffer, and it doesn't keep any direct representation in memory as a sequence of bytes -- it's just a function to put bytes into a buffer and/or yield existing buffers.  That can be a closure that (when called against a buffer) copies the bytes out of a strict ByteString (like what you'd get by calling 'byteString "a"'), a function that serializes an int (like 'intDec 42'), a function that always writes one fixed byte (something equivalent to '\ptr -> poke ptr 0x55'), or a closure that always returns an existing block of memory (such as 'insertByteString "a"').  None of these involve copying data "on the way in", but some of them are implemented via memcpy on the way out.

This mixup would explain... well, everything, really.  If a Builder were a StringBuilder, all of your arguments would be true and it would indeed be a bad choice for use in a streaming interface, and impose single-or-double copies on everything, and be expected to have zero-cost conversion to and from ByteStrings, and be unable to communicate to WAI that a chunk should be inserted vs. copied, and rely on the assumption that copies are cheaper than small chunks.

I had written a bunch of individual replies to parts of the latest message, but I've deleted them because I think this is actually the heart of the matter.

awpr:
Oh, right, I forgot about this comment in all the chaos -- @kazu-yamamoto I was using the "~" syntax as type equality, as enabled by the DataKinds extension and some other extension I can't think of right now.  I can change the Haddocks to avoid it if you think it's liable to confuse readers.

AaronFriel:
Let me get this straight then:

1. Builder as utilized by WAI is told to write to a buffer no matter what. Right now there's no signalling path for that to avoid writing to a buffer?
 
2. I think you're right that I'm confused, but I don't think it is from confusion with Java (I'm actually not very familiar with Java's StringBuilder, but what you describe sounds like the C# equivalent. Rather, it seems to me that the `Builder` API guarantees small strings are copied. 

```
byteStringThreshold maxCopySize =
    \bs -> builder $ step bs
  where
    step !bs@(S.PS _ _ len) !k br@(BufferRange !op _)
      | len <= maxCopySize = byteStringCopyStep bs k br
      | otherwise          = return $ insertChunk op bs k
```

I'm trying to wrap my head around the `Builder` internals, the continuation passing style is wonderful for performance and for the optimizer, but it does take some time to understand. If I am understanding you correctly, you're saying that `byteStringThreshold`, which uses `byteStringCopyStep`, will not create an intermediate copy of a string for small strings? 

That is: if I pass to a `Response` constructor the result of `byteString ("AUniqueString :: ByteString)`, the `Response` will write that builder to its own buffer, and so there will be *at most* two copies of "AUniqueString" in memory? The first being the source ByteString, and the second being the buffer WAI fills? If that is the case, then `Builder` is fine as an API surface and avoids that extra allocation.

It was my understanding that `byteStringCopyStep` *always* allocates an intermediate buffer, so that there would be *three* locations in memory containing the unique string. I think we would both agree that an extra copy there is completely unnecessary if it's just going to be copied again to the final buffer that WAI will send. That is the source of my confusion.

I still would like to urge that there be a `Response` API that does as you said earlier:

> Currently Warp will copy the yielded chunk anyway, but that's an implementation detail; arguably it'd be better to eat the double write call (header; data) when the fragment is bigger than some threshold. That's totally orthogonal to whether we use Builders as the interface, though.

If that threshold was a tunable parameter in the Response, I would be satisfied.

The rest of this comment is beside the point, and more for my own education than anything else.

My issues with `Builder` and `ByteString` remain, but they are orthogonal to fixing the WAI API. I still am interested in determining if there is room for improving performance by tweaking the threshold size of the `Builder`, and there is definitely room to improve `ByteString`. Simon Marlow has commented on Reddit about the weak pointer cost, and it looks like `ByteString` does avoid creating weak pointers. Still, I am going to look into submitting a patch to `ByteString` to optimize adjacent strings appends.

It sounds like you're more familiar with `Builder` than I am. Given the continuation passing style of `Builder`, is it possible to optimize writing adjacent strings without a significant amount of work? It looks like the source (ptr, offset, length) of the previous `BuildStep` isn't visible to subsequent `BuildStep`s, and this makes sense. If `Builder` can manufacture bytes out of nowhere with its numeric and other APIs, not every step comes from a definite location in memory. It would be nonsense to test for adjacency when there is no source location.

awpr:
> Let me get this straight then:
>
> 1. Builder as utilized by WAI is told to write to a buffer no matter what. Right now there's no signalling path for that to avoid writing to a buffer?

Mostly -- there's a signal produced to insert chunks directly, but Warp HTTP/2 currently handles it by copying the chunk into the buffer anyway (see https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/HTTP2/Sender.hs#L303).  This is arguably bad and we should probably open an issue to track it.

> Rather, it seems to me that the Builder API guarantees small strings are copied.

Well, *some* Builder-creating APIs guarantee that small strings are copied -- namely 'byteStringCopy' which always copies the input, 'byteStringThreshold n' which copies inputs smaller than n bytes, and 'byteString', which is 'byteStringThreshold 256'.  (And the lazy* variants of those).

> there will be at most two copies of "AUniqueString" in memory

Yes -- the Builder value 'byteString "AUniqueString"' is effectively a closure with a reference to "AUniqueString" (which presumably lives in the data segment since it's a string literal, but it could just as well have come from I/O), which takes a buffer (void* / Ptr Word8), calls memcpy, and says "I wrote 13 bytes into your buffer, and I'm done now".  This ends up happening at https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/HTTP2/Sender.hs#L335, where the provided buffer is the connection-wide output buffer.

The Builder value 'byteStringInsert "AUniqueString"' (... sorry, I've had the name backwards this whole time) is also a closure with the same reference, except when you give it a buffer it says "I wrote 0 bytes into your buffer.  Use this ByteString instead".  What Warp does *currently* is then take that ByteString and copy it into the connection send buffer anyway.  So (currently) either way exactly 2 copies exist -- the source and the send buffer.  It should be possible to take note of the length of that ByteString, add it to any bytes we previously buffered, write the frame header, and send both the buffer and the ByteString chunk, so that there would only ever be one copy of that data.  (This is the aforementioned new issue).

> that byteStringCopyStep always allocates an intermediate buffer

I really hope that's not the case, since I've been asserting it's not this whole time -- I should probably go check now.  <...>  Yeah, it won't make an intermediate buffer.  The call to 'copyBytes' there happens *inside* the BuildStep, i.e. at output time, while we're writing it to the send buffer.  Until then it's just a closure with a reference to the original ByteString (albeit one that GHC knows might call 'touchForeignPtr' on the ByteString -- I don't totally know how this works internally, but it keeps the ByteString's memory alive until it's used).

> If that threshold was a tunable parameter

I totally agree that it's important to be able to tune the threshold of copy vs. insert for these Builders.  However I don't think such a threshold should actually be represented in Warp at all; it would be a property of the code creating the Builder -- i.e. by defining at module level "myByteString = byteStringThreshold 3" in your application, and using that instead of 'byteString' to construct responses.  Of course, Warp could have some canned advice about this, with experimentally determined values, like: "On Linux 4.2, we suggest using 'byteStringThreshold 256' to optimize for HTTP/1.*, and 'byteStringThreshold 512' for HTTP/2.  This is because inserting a chunk costs 2 syscalls instead of just 1 in HTTP/2.  If you have a MagicNIC and use MagicOS, you should try setting the threshold to 3 instead because of the Magic Hardware Buffer and cheap syscalls.  If for some reason your main memory is still a rotating drum, use 0 because copying a byte is obscenely expensive."

> is it possible to optimize writing adjacent strings without a significant amount of work?

Maybe?  But probably not in the way you're imagining.  You're right that a BuildStep can't very well look at where the last BuildStep's data came from.  But, it should be possible to have something like 'byteStringCoalesce :: [ByteString] -> Builder' that picks apart the ByteStrings and turns spans of adjacent ranges into a single 'byteStringThreshold'.  Of course, it'd decompose nicely into (byteStringThreshold n . coalesce bss), where coalesce :: [ByteString] -> [ByteString].

One more thing: I don't fully understand the properties of ByteArray#, but I *think* changing ByteString to use that would outlaw a **lot** of existing ByteString uses.  I believe that change would require all ByteStrings to be allocated in a particular way by GHC itself.  But ByteStrings are often allocated and freed in arbitrary ways -- I've personally written a few modules that construct ByteStrings around memory allocated by C++ code, and I believe things like X11 and GTK bindings (libraries which have their own allocators) require this ability too.  So you might see significant resistance to the change you're suggesting.

In fact (I just Googled this...) ShortByteString sounds like exactly what you're proposing to change ByteString to: https://hackage.haskell.org/package/bytestring-0.10.6.0/docs/src/Data.ByteString.Short.Internal.html#ShortByteString

kazu-yamamoto:
@AaronFriel Do you understand that small values can be directly converted to `Builder`, not via `ByteString`. E.g. http://hackage.haskell.org/package/bytestring-0.10.6.0/docs/Data-ByteString-Builder.html#v:word8

kazu-yamamoto:
> Oh, right, I forgot about this comment in all the chaos -- @kazu-yamamoto I was using the "~" syntax as type equality, as enabled by the DataKinds extension and some other extension I can't think of right now.  I can change the Haddocks to avoid it if you think it's liable to confuse readers.

In my opinion, `=` is better than `~`.

awpr:
Updated to use words instead of symbols -- here, on yesodweb/wai:http2-push, and on awpr/wai:wai_http2_staging.

kazu-yamamoto:
I would say this again. `ByteString`s are pinned objects. But global locks are taken only when large `ByteString` (>= 3,272 bytes on 64 bit machines) are allocated. They are called large objects in the GC terminology. No global locks are taken to allocate small `ByteString`s.

kazu-yamamoto:
> That is: if I pass to a `Response` constructor the result of `byteString ("AUniqueString :: ByteString)`, the `Response` will write that builder to its own buffer, and so there will be *at most* two copies of "AUniqueString" in memory? The first being the source ByteString, and the second being the buffer WAI fills? If that is the case, then `Builder` is fine as an API surface and avoids that extra allocation.

Yes. Two copies now. 

Reading disccussion, I hit upon to reduce one copying:

- For plain HTTP, `writev()` can avoid copying. Warp has `writev()` API which is not used currently.
- For HTTP over TLS, I found that zero copy (in Warp) is still possible. `sendData` of TLS takes lazy `ByteString`. So, we can convert `Builder` to lazy `ByteString` (which is a list of strict `ByteString`) without copying and pass it to `sendData`. In TLS lib, the lazy `ByteString` will be copied. But we can reduce one copying in this way.

It's worth trying.

kazu-yamamoto:
For this, we should learn the internal of [toLazyByteString](http://hackage.haskell.org/package/bytestring-0.10.6.0/docs/Data-ByteString-Builder.html#v:toLazyByteString).

kazu-yamamoto:
> Updated to use words instead of symbols -- here, on yesodweb/wai:http2-push, and on awpr/wai:wai_http2_staging.

Nice!

kazu-yamamoto:
@awpr In `type HTTP2Application = Request -> PushFunc -> Responder`, why `PsuhFunc` is the second? Do you use partial apply somewhere?

It seems to me that `type HTTP2Application = PushFunc -> Request -> Responder` is more understandable.

awpr:
It was somewhat arbitrary, but initially `Responder` was inlined there, so the `PushFunc` was adjacent to the respond function.

I'm wondering whether it would make sense to give `Responder` a push function too, since it's possible for the associated stream of a PUSH_PROMISE frame to be itself a pushed stream.  In that case it would be just `type HTTP2Application = Request -> Responder`.  I don't mind swapping the argument order, but I'd like to consider whether this is better.

awpr:
Regarding `writev()` and sending lazy `ByteString`s to `tls`, I think Warp can actually do better than `toLazyByteString` because we also have the guarantee that the buffer is safe to reuse as soon as the write call returns.  I'm sure there's still insight to be gained from its implementation, though.

I think I might be able to reuse part of a failed experiment to make this complexity manageable -- using a bespoke Monad stack to automatically track the write offset and flush function, so that individual steps of Builders become composable.  It seems like it'd be easy enough to add a FIFO queue of strict `ByteString`s to the state, to be turned into a `writev()` call and/or lazy `ByteString` when the buffer space is exhausted.  (I implemented that while trying to move data frame writing into per-stream buffers, but it turned out that was a dead end because headers must be sent in the same order they're compressed, and the priority queue can't guarantee that if they're sent concurrently.)

kazu-yamamoto:
> I'm wondering whether it would make sense to give `Responder` a push function too, since it's possible for the associated stream of a PUSH_PROMISE frame to be itself a pushed stream.  In that case it would be just `type HTTP2Application = Request -> Responder`.  I don't mind swapping the argument order, but I'd like to consider whether this is better.

Do you mean the followings?

```
type HTTP2Application = Request -> Responder
type Responder = PushFunc -> (forall a. Status -> ResponseHeaders -> Body a -> IO a) -> IO Trailers
type PushFunc = PushPromise -> Responder -> IO Bool
```

In this definition, `PushFunc` and `Responder` are mutually recursive. Is this possible?

awpr:
Ah, yes, I think GHC would complain about infinite types there.  However if Response becomes a newtype like in https://github.com/awpr/wai/commit/211c75cfafc0e7f8edd201ac2ee9f4829fe68f4f, it would be fine:

    type HTTP2Application = Request -> Responder
    newtype Responder = Responder { runResponder :: forall s. PushFunc -> (forall a. RespondFunc s a) -> IO s }
    type PushFunc = PushPromise -> Responder -> IO Bool

kazu-yamamoto:
@awpr I feel that the original one is much easier to understand.

kazu-yamamoto:
In your [example](https://gist.github.com/awpr/e9aecd949d6546e88688), `indexPage` pushes two files. Is it done concurrently? Or sequentially?

If this is sequential, I would implement concurrent pushing.

awpr:
It's definitely easier to understand, but based on the anecdotal evidence that I managed to write a call-the-wrong-respond-function bug in the *one* HTTP2Application I've ever written using PushPromise, I think the conceptual overhead is justified -- that's a 100% error rate!

In `indexPage`, each call to `push` blocks just long enough to know the sender has dequeued the push `Output`, but the actual stream body is run concurrently (in its own thread).  That allows applications to adhere to the HTTP/2 spec in section [8.2.1](https://http2.github.io/http2-spec/#rfc.section.8.2.1): "The server SHOULD send PUSH_PROMISE (Section 6.6) frames prior to sending any frames that reference the promised responses."

kazu-yamamoto:
@awpr OK. Would you push 211c75c to wai repo? I will continue the review based on that.

awpr:
Done (as well as the usability changes that followed it).

kazu-yamamoto:
Thank you.

Is the definition of `HTTP2Application` still old?

Also, I would like to have the gist example updated.

awpr:
HTTP2Application isn't missing any changes that I've actually implemented -- I just haven't made the change to move PushFunc into Responder yet.

Updated the example.

AaronFriel:
> I totally agree that it's important to be able to tune the threshold of copy vs. insert for these Builders. However I don't think such a threshold should actually be represented in Warp at all; it would be a property of the code creating the Builder -- i.e. by defining at module level "myByteString = byteStringThreshold 3" in your application, and using that instead of 'byteString' to construct responses. Of course, Warp could have some canned advice about this, with experimentally determined values...

Ah, so will WAI change to hew more closely to the `Builder` spec of not copying when using `byteStringInsert`? That would be excellent and allow finer-grained control.

> I would say this again. ByteStrings are pinned objects. But global locks are taken only when large ByteString (>= 3,272 bytes on 64 bit machines) are allocated. They are called large objects in the GC terminology. No global locks are taken to allocate small ByteStrings.

I am not sure if we need to use pinned objects for WAI, ever. I understand that the current `ByteString` API contract states that all backing memory for them is pinned, but that, strictly speaking, doesn't *have* to be the case. Pinned objects cause heap fragmentation and make the garbage collector's work more difficult, but I think it's the lock more than anything else that is the issue. 

There are plenty of cases where unpinned memory works great for ByteString, and it makes sense that the default is pinned (it makes sure that complex FFI won't break because of a moved value), but one-off FFI where you just need to pass a string once don't require pinned values.

I don't think `ShortByteString` is what you want either, because there is no `Builder` for it, and that takes us back to all of the previously discussed objections to using a `ByteString` API. What you gain in not having a global lock, you lose in having to `memcpy` everything or write your own (brittle) wrapper around `ShortByteString` to act as a `Builder` (for filling buffers) and taking substrings.

> Regarding writev() and sending lazy ByteStrings to tls, I think Warp can actually do better than toLazyByteString because we also have the guarantee that the buffer is safe to reuse as soon as the write call returns. I'm sure there's still insight to be gained from its implementation, though.

I would agree with this analysis. We can't use `tls` without having to write the encrypted string somewhere, but I think we can at least be zero-copy until that point.

awpr:
> Builder spec of not copying when using byteStringInsert

There's not really any such spec -- byteStringInsert promises that running the Builder will give 'Chunk bs rest' rather than writing the contents to the given buffer, but it doesn't (and can't) guarantee anything about what the caller will do with that ByteString value.

That said I do think Warp's HTTP/2 sender (not WAI, which has no control over this) should potentially change to avoid copying ByteString chunks yielded by the builder.  This could conceivably cause poor performance for existing Builders that yield small chunks, though.  Maybe it could be a boolean setting -- "yes, I'm sure my Builders are tuned; please don't copy them" -- if there's a large performance cost for naive applications.

> I am not sure if we need to use pinned objects for WAI, ever

The send buffer already doesn't use pinned memory; it's allocated by Run.hs calling 'allocateBuffer', which is defined in Buffer.hs as 'mallocBytes', which allocates "outside of the area maintained by the Haskell storage manager".  Pinned memory is special in that it's garbage-collected but not movable; this memory is *not* garbage-collected and therefore doesn't need any special treatment to be sure it won't move.

> that, strictly speaking, doesn't have to be the case

With the current ByteString representation (namely a pointer in the form of ForeignPtr Word8), it does, because a ByteString pointing at movable memory would become a dangling reference if that memory were moved.  Changing the ByteString constructor to use ByteArray# is almost certainly a non-starter because it would prevent using memory allocated by anything other than GHC's heap allocator.

> one-off FFI where you just need to pass a string once don't require pinned values

I don't think this is true anyway, because there's nothing preventing the GC from running *during* an FFI call.  If the FFI call is marked as 'safe', then the runtime will have released the Capability before entering the call, and a collection can happen in the middle of the call.  Even if you can make it safe by marking the FFI call as 'unsafe' (which I'm not sure you can), you'd be better off using something like ShortByteString anyway.

Since Warp wraps up its buffers in ByteStrings for I/O and uses them in FFI'd calls like pread, they really do need to be immovable, so they couldn't be ShortByteString even if we wanted them to.  But I still don't see where you're getting this talk of a global lock anyway -- each HTTP/2 session gets *one* send buffer which is not even pinned heap memory anyway, and sending individual chunks of data for a stream shouldn't allocate *any* ByteStrings large or otherwise.

AaronFriel:
> > I am not sure if we need to use pinned objects for WAI, ever

> The send buffer already doesn't use pinned memory; it's allocated by Run.hs calling 'allocateBuffer', which is defined in Buffer.hs as 'mallocBytes', which allocates "outside of the area maintained by the Haskell storage manager". Pinned memory is special in that it's garbage-collected but not movable; this memory is not garbage-collected and therefore doesn't need any special treatment to be sure it won't move.

Oh, I agree. I think we're talking past each other though. The issue is that `ByteString` uses pinned memory, and `Builder` is built on `ByteString`, so when the web applications create `ByteStrings`, for example, when encoding from `Text` to `ByteString`, that creates unnecessary fragmentation.

This is why I think that it might be useful to rewrite `ByteString` in terms of `ByteArray#`, where it would use pinned memory by default (to preserve existing contracts) and an `Unpinned` module (to avoid global locking.) 

> Changing the ByteString constructor to use ByteArray# is almost certainly a non-starter because it would prevent using memory allocated by anything other than GHC's heap allocator.

I'm not sure why that would be the case? I would have to look into the GHC primops, but if we receive memory via FFI allocated outside of GHC, I think there is a way to treat it as pinned - it wouldn't be in the GHC heap anyhow. I would have to look into this, but I know `ByteArray#` is a supported type for FFI, so it would be surprising to me that this would not work.

That would require significant digging into the FFI - but I want to reiterate:

**I am not suggesting changing the default contract of `ByteString` from using pinned memory.** I want to look into improving perf by avoiding the lock when it's unnecessary. That is: when the memory is allocated by GHC and it will be sent to one-off FFI functions that do not hold any reference to it after the call completes.

> > one-off FFI where you just need to pass a string once don't require pinned values

> I don't think this is true anyway, because there's nothing preventing the GC from running during an FFI call. If the FFI call is marked as 'safe', then the runtime will have released the Capability before entering the call, and a collection can happen in the middle of the call. Even if you can make it safe by marking the FFI call as 'unsafe' (which I'm not sure you can), you'd be better off using something like ShortByteString anyway.

The manual docs specifically say that ByteArray# is FFI safe. If the GC might move the memory during the call, that seems like something that should be noted.

> ## 8.1.1. Unboxed types
> 
> The following unboxed types may be used as basic foreign types (see FFI Addendum, Section 3.2): Int#, Word#, Char#, Float#, Double#, Addr#, StablePtr# a, MutableByteArray#, ForeignObj#, and ByteArray#.

(From GHC Manual).

awpr:
I have no authority on matters of ByteString internals whatsoever, so I'm not going to spend any more time debating that change, but my reference for ByteArray#s moving during FFI calls is https://ghc.haskell.org/trac/ghc/wiki/Commentary/Rts/Storage/GC/Pinned.  I've seen some really stale information on that wiki before, so maybe it's not true anymore.

As for Builders, it's entirely possible to use Builders without ever allocating a ByteString.  Yielding a 'ByteString' chunk is an optimization for when a ByteString already exists.  Needing to use a ByteString as part of a Builder is a symptom of either a missing library function (like Text -> Builder that will encode the Text's codepoints directly into the given buffer), or some upstream API that gave you a ByteString.

AaronFriel:
> As for Builders, it's entirely possible to use Builders without ever allocating a ByteString. Yielding a 'ByteString' chunk is an optimization for when a ByteString already exists. Needing to use a ByteString as part of a Builder is a symptom of either a missing library function (like Text -> Builder that will encode the Text's codepoints directly into the given buffer), or some upstream API that gave you a ByteString.

I understand, I'm just looking for ways to avoid the `ByteString` intermediate if it's unnecessary. I'm unsure if I should file a trac ticket or where I could get clarification on the wiki article, or just bug people in #haskell on IRC until someone can clarify.

awpr:
I'm not sure why I couldn't find this on Hoogle yesterday, but I stumbled upon it browsing the 'text' Haddocks: http://hackage.haskell.org/package/text-1.2.1.3/docs/Data-Text-Encoding.html#g:6.  If you're concerned about the Text case in particular, this is your answer -- it transcodes the internal UTF-16 directly into UTF-8 in the output buffer without allocating intermediate ByteStrings.

awpr:
Well that's it, I'm switching to Stackage Hoogle -- over there it's the first result for "Text -> Builder".  On haskell.org Hoogle it doesn't seem to be indexed at all ("encodeUtf8Builder +text" gives nothing).

kazu-yamamoto:
@awpr OK. My review is done. The quality of this patch is really high. I have merged this pull request. Thank you for your contribution.

Please update TBD in the doc of `Network.WAI.HTTP2`.

AaronFriel:
@awpr Good catch. Yeah, I wasn't able to find it on Hoogle either and assumed it didn't exist. I'm glad that it does! And it would be great to see more such functions.

PR #442, opened 2015-10-05
Add custom detailed logging constructor and JSON log formatting

pseudonom:


gregwebs:
@snoyberg you might want to look at this. Main thing is that it adds some additional dependencies.

snoyberg:
I'm on the fence about the aeson dependency, the rest looks fine though. @pseudonom how would you feel about making the JSON part of this PR a separate package for now, and possibly including it in wai-extra in the future?

gregwebs:
My feeling is that almost everything made with WAI will end up depending on aeson, But I always defer to your judgement on dependencies and we can create the separate package. We should link to it from one of the wai-extra modules

snoyberg:
Like I said, I'm on the fence, not necessarily opposed. You're probably right that everyone will be using aeson. OK, consider my mind changed on this: let's include it.

gregwebs:
ok, I will commit and release this tomorrow then unless there are other objections. @pseudonom we just need a Changelog entry now.

creichert:
FWIW, I am definitely in favor of this PR and think it's a very useful addition. +1.

PR #445, opened 2015-10-07
Bump blaze-builder minimum bound to 0.4 to fix build.

### LINKED ISSUE #443: incorrect lower bound on blaze-builder in wai-3.0.4.0
[I just increased the lower bound](http://hackage.haskell.org/package/wai-3.0.4.0/revisions/) for `blaze-builder` to require `>= 0.4` because otherwise cabal may pick an install plan resulting in the following compile error:

```
[3 of 3] Compiling Network.Wai.HTTP2 ( Network/Wai/HTTP2.hs, dist/dist-sandbox-a85cdbf6/build/Network/Wai/HTTP2.o )

Network/Wai/HTTP2.hs:189:19:
    No instance for (Data.String.IsString Builder)
      arising from the literal ‘"File not found."’
    In the first argument of ‘streamBuilder’, namely
      ‘"File not found."’
    In the second argument of ‘($)’, namely
      ‘streamBuilder "File not found."’
    In the expression:
      k H.notFound404 h' $ streamBuilder "File not found."
cabal: Error: some packages failed to install:
wai-3.0.4.0 failed during the building phase. The exception was:
```

### LINKED ISSUE #444: build error in wai 3.0.4.0
Hi there,

when building wai, the compiler aborts with this error:

```
Network/Wai/HTTP2.hs:189:19:
    No instance for (Data.String.IsString Builder)
      arising from the literal ‘"File not found."’
    In the first argument of ‘streamBuilder’, namely
      ‘"File not found."’
    In the second argument of ‘($)’, namely
      ‘streamBuilder "File not found."’
    In the expression:
      k H.notFound404 h' $ streamBuilder "File not found."
```

I can't fix this myself, since I do not understand the code at that point...

awpr:
It's probably possible to make this continue to work with the older blaze-builder (instinctively I bet 0.4 was the first version to be aware of OverloadedStrings), but I haven't looked into it yet.  Should we take this version bump or change the code to avoid relying on the IsString Builder instance?

Fixes #443
Fixes #444

### DIFF
--- a/wai/Network/Wai/HTTP2.hs
+++ b/wai/Network/Wai/HTTP2.hs
@@ -54,6 +54,7 @@ module Network.Wai.HTTP2
     ) where
 
 import           Blaze.ByteString.Builder (Builder)
+import           Blaze.ByteString.Builder.ByteString (fromByteString)
 import           Control.Exception (Exception, throwIO)
 import           Control.Monad.Trans.Cont (ContT(..))
 import           Data.ByteString (ByteString)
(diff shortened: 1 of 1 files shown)


awpr:
It turns out this was a really easy fix to keep the lower bound where it currently is, so I'm doing that instead.  Closing this PR and pushing directly since the new fix should be uncontroversial.

PR #449, opened 2015-10-13
Add back http-types 0.8 support

snoyberg:
Reason: many packages do not yet support http-types 0.9. This increased
lower bound causes many imcompatibility issues on Hackage. I'd like to
get this merged and released ASAP if there are no objection.

Pinging @kazu-yamamoto


kazu-yamamoto:
What about 7d5fc7751cae2a89f04c4bd3dc199eb0004a0b28?

snoyberg:
That works, it just seems much more involved than using the original approach of a string literal. I'm OK with whatever you prefer using.

kazu-yamamoto:
Yes, my patch is vocal a bit but I prefer this to not forget the feature job.

kazu-yamamoto:
Tested both with http-types 0.9 and 0.8.6. All tests on Travis CI are passed. I have released a new version.

PR #451, opened 2015-10-15
Content length

gregwebs:
Add Network.Wai.Header.contentLength helper for reading the Content-Length header

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RequestLogger.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger.hs
@@ -41,6 +41,7 @@ import Network.Wai.Internal (Response (..))
 import Data.Default.Class (Default (def))
 import Network.Wai.Logger
 import Network.Wai.Middleware.RequestLogger.Internal
+import Network.Wai.Header (contentLength)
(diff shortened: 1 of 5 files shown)


wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -308,10 +305,7 @@ detailedMiddleware' cb ansiColor ansiMethod ansiStatusCode app req sendResponse
     collectPostParams (postParams, files) = postParams ++
       map (\(k,v) -> (k, "FILE: " <> fileName v)) files
 
-    readInt bs =
```
gregwebs: This is actually different than S8.readInteger. I was assuming that was an accident, but maybe there is a purpose?

snoyberg:
Isn't this already handled by [requestBodyLength](http://haddock.stackage.org/lts-3.9/wai-3.0.4.0/Network-Wai.html#v:requestBodyLength)?

gregwebs:
one is for the request and the other for the response. Thanks for the reminder about requestBodyLength, I will have it use that

gregwebs:
closing in favor of #453

PR #452, opened 2015-10-15
Gzip minimum

gregwebs:
This is on top of a couple commits from #451. Untested at the moment, I will test before merging, but wanted to get this PR up to see if it makes sense

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-2.17
+resolver: lts-3.9
 packages:
 - ./wai
 - ./wai-extra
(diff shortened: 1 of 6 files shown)


snoyberg:
I like, though as with #451, think we should just use [requestBodyLength](http://haddock.stackage.org/lts-3.9/wai-3.0.4.0/Network-Wai.html#v:requestBodyLength).

gregwebs:
The change to gzipping is only for the response

snoyberg:
Oh, got it. Alright, that makes more sense. Maybe it's worth including in the docs that `requestBodyLength` can handle the request header case?

gregwebs:
closing in favor of #453

Issue #376, opened 2015-05-20
can't server index at root with wai-app-static


winterland1989:
i want to server a static webpage using WaiAppStatic.Storage.Embedded like this:

``` haskell
{-# LANGUAGE TemplateHaskell, QuasiQuotes, OverloadedStrings #-}

module Static (mkEmbedded) where

import WaiAppStatic.Storage.Embedded
import Crypto.Hash.MD5 (hashlazy)
import Network.Mime (MimeType)
import qualified Data.ByteString.Lazy as BL
import qualified Data.ByteString.Base64 as B64
import qualified Data.Text as T
import qualified Data.Text.Encoding as T
import System.FilePath ((</>))

hash :: BL.ByteString -> T.Text
hash = T.take 8 . T.decodeUtf8 . B64.encode . hashlazy

staticFiles :: [(FilePath, MimeType)]
staticFiles = [
        ("index.html"         , "text/html")
    ,   ("main.js"            , "application/x-javascript")
    ,   ("html-sanitizer.js"  , "application/x-javascript")
    ,   ("ext.js"             , "application/x-javascript")
    ,   ("main.css"           , "text/css")
    ]

embedFile :: (FilePath, MimeType) -> IO EmbeddableEntry
embedFile (file, mime) = do
    f <- BL.readFile $ "static" </> file
    return $ EmbeddableEntry {
            eLocation = T.pack file
        ,   eMimeType = mime
        ,   eContent  = Left (hash f, f)
        }

mkEmbedded :: IO [EmbeddableEntry]
mkEmbedded = mapM embedFile staticFiles
```

and i create my application like this:

``` haskell
{-# LANGUAGE TemplateHaskell, OverloadedStrings #-}
module Main where

import Static (mkEmbedded)
import Network.Wai (Application)
import Network.Wai.Application.Static
import WaiAppStatic.Storage.Embedded
import Network.Wai.Handler.Warp (run)
import WaiAppStatic.Types (unsafeToPiece)

myApp :: Application
myApp = staticApp $(mkSettings mkEmbedded) {ssIndices = [unsafeToPiece "index.html"]}

main :: IO ()
main = run 3000 myApp

```

i also tried `ssRedirectToIndex = True`, but directly visit `localhost:3000` result in 404, `localhost:3000/index.html`worked as expected, is there something wrong with my code?

snoyberg:
[...]

PR #460, opened 2015-11-05
wai-app-static: honor ssIndices when used with defaultWebAppSettings

soenkehahn:
See #376.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
     - travis_retry cabal update
     - cabal unpack wai-logger
     - mv wai-logger-* wai-logger
-    - cabal install hspec doctest HTTP cabal-meta cabal-src
+    - cabal install hspec doctest HTTP temporary cabal-meta cabal-src
     - cabal-meta install --force-reinstalls
 
 script: mega-sdist --test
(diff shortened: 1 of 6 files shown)


soenkehahn:
I'm not sure if I prefer the version with or without the last commit (0c67261). @snoyberg: Any preference? Apart from that this is ready to merge from my side.

Any other feedback? Should I squash the commits?

snoyberg:
I'm confused, this looks like a completely separate goal than #376. Can you clarify?

soenkehahn:
Sorry, I was also a bit confused.

If `ssLookupFile` returned an `LRNotFound` then `wai-app-static` assumed that the looked up file is not a directory and it didn't bother to further check for index files. `ssIndices` was only ever considered when `ssLookupFile` returned an `LRFolder`. The `StaticSettings` created with `mkSettings` would not return an `LRFolder`, which is the cause for #376.

Personally I ran into a similar issue, not with `mkSettings` but with `defaultWebAppSettings`. Its `ssLookupFile` also doesn't return `LRFolder`s for performance reasons.

With my PR in the case that `ssLookupFile` returns `LRFolder` _or_ `LRNotFound` `ssLookupFile` will be called again, this time searching for index files. So `LRFolder` will not be used for looking up index files anymore. This fixes both my issue and #376. I added a test case for #376.

soenkehahn:
The builds on travis are failing but it doesn't look like this is caused by my commits. Is travis for `wai` currently broken?

snoyberg:
wai master is not broken. Can you try rebasing?

soenkehahn:
I rebased and added f088f32, now travis is happy. I hope it's the right way to fix this. (IIUC `cabal-meta` doesn't pick up test dependencies.)

soenkehahn:
Thanks for merging!

PR #461, opened 2015-11-09
Added wai-devel

urbanslug:
Updated stack.yaml and added wai-devel.


snoyberg:
I hadn't realized your intention was to include all of wai-devel in the wai repo. I don't see the motivation for doing so, can you clarify?

urbanslug:
The aim was to have all things to do with WAI under one repo. I thought that that is what this is.

snoyberg:
I don't know who's aim it was. I never stated that this should move into
the main repo.

Issue #465, opened 2015-11-17
Handle no longer auto-flushing


MHova:
`Network.Wai.Middleware.RequestLogger.Destination` of type `Handler`, which is the destination type for [default] (https://github.com/yesodweb/wai/blob/master/wai-extra/Network/Wai/Middleware/RequestLogger.hs#L83) `RequestLoggerSettings`, seems to have lost auto-flushing with this [commit] (https://github.com/yesodweb/wai/commit/441a1d776c5e323d9529802a33aac09543da5d59).

PR #466, opened 2015-11-17
Autoflush handle

MHova:
Addresses https://github.com/yesodweb/wai/issues/465

While we're here, also honor `autoFlush` for `Logger`. However I'll remove that from this PR if someone can confirm that this [commit] (https://github.com/yesodweb/wai/commit/d4e843be5b48bf0c1ac9018281281051b9e0a98e) *intentionally* always forces flushing for `Logger`.


snoyberg:
Thanks!

PR #467, opened 2015-11-18
add 404 handler

jhickner:
Adds a new setting called `ss404Handler` that specifies a `Maybe Application` to be called in case of 404 errors.


wai-app-static/WaiAppStatic/Types.hs
```diff
@@ -144,4 +144,7 @@ data StaticSettings = StaticSettings
 
       -- | Force a trailing slash at the end of directories
     , ssAddTrailingSlash :: Bool
+
+      -- | Optional `W.Application` to be used in case of 404 errors
+    , ss404Handler :: Maybe W.Application
```
snoyberg: This needs to be exposed from Network.Wai.Application.Static to be usable.

jhickner:
OK, added

PR #469, opened 2015-11-21
Make sure HEAD response bodies are empty (fixes #468)

### LINKED ISSUE #468: Warp sends response bodies for HEAD requests
Simple repro: run the following minimal Warp application:

```haskell
import Network.Wai
import Network.HTTP.Types

main :: IO ()
main = run 3000 $ \_ respond -> respond $ responseLBS status200 [] "This body should not appear for HEAD"
```

Then make a HEAD request:

    curl -i -X HEAD http://localhost:3000

Expected: the body does not appear

Actual: it does

It seems like this can be resolved easily by modifying `hasBody` (in Response.hs) to check not only the status code, but also the request method. However, the comments in that module quite clearly say that HEAD was intentionally left in. though I'm not sure what the reason is. (I also have a suspicion that I'm the one who made that decision...)

@kazu-yamamoto Can you see any objection to making the modification to `hasBody` that I just described?

snoyberg:
I've put together an initial patch to handle this. I'd still like to add a regression test to ensure that HEAD requests on ResponseFile, ResponseBuilder, and ResponseStream all return empty bodies, but I wanted to get your feedback on the approach.

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+set -eux
+
+mkdir -p $HOME/.cabal
+cat > $HOME/.cabal/config <<EOF
+remote-repo: hackage.haskell.org:http://hackage.fpcomplete.com/
+remote-repo-cache: $HOME/.cabal/packages
+jobs: \$ncpus
+EOF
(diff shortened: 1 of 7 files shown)


snoyberg:
Alright, this now has tests in place for builder, streaming, and file (both with and without range) requests.

snoyberg:
This was rebased and merged to master, closing.

Issue #459, opened 2015-10-26
Set END_STREAM on the last non-empty DATA frame when there will be no trailers.


kazu-yamamoto:
Firefox sends RST_STREAM with CANCEL when Warp sends HEADERS + DATA.

This is probably because Warp sends 0-length DATA frame with END_STREAM set when trailers are empty. In the spec, 0-length DATA is legal. But this is redundant and this is the major case.

@awpr Is it possible to not generate 0-length DATA when trailers are empty?

awpr:
I believe the spec explicitly condones this behavior.  Initially I had it
send an empty trailers block for this case, but Chrome considered that a
protocol error, so I changed it to use an empty data frame.

kazu-yamamoto:
It appeared that 0 length data frames are sent twice. One with ACK and the other without ACK. It seems to me that these logically unnecessary frames give big impact on performance.

awpr:
Interesting, it definitely shouldn't be sending two.  By ACK do you mean END_STREAM, and in the order you mentioned them?  e.g. is it sending DATA(flags=END_STREAM, "") followed by DATA(flags=0, "")?  If so that looks like a protocol error.

I'll try to find how it could end up doing this.

kazu-yamamoto:
Sorry.
s/ACK/END_STREAM/

kazu-yamamoto:
The following sequence occupies CPUs:

```
1150  sendto(13, "\0\0\0\0\0\0\0\0\3", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\1\0\0\0\3", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\0\0\0\0\5", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\1\0\0\0\5", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\0\0\0\0\7", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\1\0\0\0\7", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\0\0\0\0\t", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\1\0\0\0\t", 9, 0, NULL, 0) = 9
1150  sendto(13, "\0\0\0\0\0\0\0\0\v", 9, 0, NULL, 0) = 9
[...]

PR #472, opened 2015-11-30
No zero length data frame and HEAD fix

kazu-yamamoto:
Fixing #459.


awpr:
First three look good; left a comment on the hasBody commit.

kazu-yamamoto:
Thank you for your review.

I'm very sorry but I would like to withdraw this pull request since I would like to start over the HTTP/2 stuff due to the performance issue. I will start the work on the next Monday.

kazu-yamamoto:
This pull request is not valid anymore. Let's close.

Issue #359, opened 2015-04-03
Sophisticated file handling


kazu-yamamoto:
This is random thoughts to support multiple ranges. (#345)

`wai-app-file-cgi` (base library for `mighttpd2`) also implements `Range:` by itself. To support multiple ranges in `warp`, I would like to remove this duplications.

First, the cache of file info (size, modification time, etc) should be ported from `mighttpd2`. This would drastically improve performance for non-sophisticate WAI applications since the number disk access decreases.

The cache of file info should be visible to WAI applications since they want to log how many bytes are sent without disk access. So, it can be specified through `Settings`. If not specified, it is initialized and used only internally.

The logic of `If-Modified-Since:`, `If-Unmodified-Since:`, and `If-Range:` should be ported from `wai-app-file-cgi`. We should discuss on `ETag:`, `If-Match:`, and `If-None-Match:`.

When `Maybe FilePart` of `ResponseFile` is `Nothing`, it indicates that normal contents are being sent. The range logic should run. Otherwise, range logic should not run. This typical case is to send a status message from a user-custom file.

kazu-yamamoto:
A crazy idea on the file info cache: adding a new member to`Request`, whose type is `FilePath -> IO FileInfo`.

awpr:
HTTP2Applications introduced by #426 are capable of streaming multiple file fragments as part of their normal operation; this makes it possible to implement multipart support for HTTP/2 without copying chunks of file data.

kazu-yamamoto:
@snoyberg Can I use `vault` to disclose the new cache of file information from Warp to applications?

snoyberg:
I don't see a reason why not.

kazu-yamamoto:
Thanks. I will export `getFileInfo` as `pauseTimeout` is exported.

I'm tackling this issue and #345 on HTTP/1.1 now.

kazu-yamamoto:
I'm closing this in favor of #477.

PR #473, opened 2015-12-01
File info cache

kazu-yamamoto:
These patches implement the file information cache.
This is the first step for #359.

@snoyberg Please review and merge this pull request.


warp/Network/Wai/Handler/Warp.hs
```diff
@@ -219,10 +223,26 @@ setManager x y = y { settingsManager = Just x }
 -- Enabling this cache results in drastic performance improvement for file
 -- transfers.
 --
--- Default value: since 3.0.13, default value is 0, was previously 10
+-- Default value: 0, was previously 10
```
snoyberg: Perhaps should include the comment from below:

> 0 means that the cache mechanism is not used.

warp/Network/Wai/Handler/Warp.hs
```diff
@@ -371,3 +391,10 @@ setHTTP2Disabled y = y { settingsHTTP2Enabled = False }
 -- Since 3.0.10
 pauseTimeout :: Request -> IO ()
 pauseTimeout = fromMaybe (return ()) . Vault.lookup pauseTimeoutKey . vault
+
+-- | Getting file information from the inside cache.
+--   If file information is not available, an IO error is thrown.
+--
+-- Since 3.1.10
+getFileInfo :: Request -> FilePath -> IO FileInfo
+getFileInfo = fromMaybe (\_ -> throwIO (userError "getFileInfo")) . Vault.lookup getFileInfoKey . vault
```
snoyberg: Wouldn't it be better to distinguish between "you're not actually using warp and therefore `getFileInfoKey` isn't populated" and "I couldn't find that file in the cache"

warp/Network/Wai/Handler/Warp/Request.hs
```diff
@@ -278,3 +281,7 @@ checkCR bs pos = if pos > 0 && 13 == S.index bs p then p else pos -- 13 is CR
 pauseTimeoutKey :: Vault.Key (IO ())
 pauseTimeoutKey = unsafePerformIO Vault.newKey
 {-# NOINLINE pauseTimeoutKey #-}
+
+getFileInfoKey :: Vault.Key (FilePath -> IO FileInfo)
```
snoyberg: By my recommendation above, this would return an `IO (Maybe FileInfo)`

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -237,13 +238,18 @@ runServeSettingsConnectionMakerSecure :: Settings
...
-            withTimeoutManager $ \tm ->
-                acceptConnection set getConnMaker serveConn dc fc tm counter
+    withII $ acceptConnection set getConnMaker serveConn counter
   where
+    withII action =
+        D.withDateCache $ \dc ->
+        F.withFdCache fdCacheDurationInSeconds $ \fc ->
+        I.withFileInfoCache fdFileInfoDurationInSeconds $ \get ->
+        withTimeoutManager $ \tm -> do
+            let ii0 = InternalInfo undefined tm fc get dc -- fixme: undefined
```
snoyberg: What's the fixme for?

warp/Network/Wai/Handler/Warp.hs
kazu-yamamoto: This explanation is included in the beginning of the manual, isn't it?

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: I would like to get rid of `undefined`. If we provide `defaultHandle` or something, this can be achieved. However, we need to introduce one more `unsafePerformIO`.

If you think that `undefined` is a necessary evil, let's leave it as it is.

warp/Network/Wai/Handler/Warp.hs
kazu-yamamoto: If warp is not used, applications cannot use `getFileInfoKey`. In other words, applications using `getFileInfo` or `pauseTimeout` depend on warp.

If you don't like these dependency, wai should provide them by itself. What do you think?

warp/Network/Wai/Handler/Warp.hs
snoyberg: Yup, scratch that, the diff display simply wasn't showing the earlier part of the comment.

warp/Network/Wai/Handler/Warp.hs
snoyberg: What I'm saying is that there are really two different ways this can fail:

1. The `Vault.lookup getFileInfoKey` fails. In this case, it's because Warp isn't being used, and I'm OK with that throwing an exception, since it's a clear misuse of the API
2. The lookup function itself fails because the file in question isn't in the cache. That's the case that I'd prefer not throw an exception, but instead return a `Nothing` value.

warp/Network/Wai/Handler/Warp/Run.hs
snoyberg: I just misread, ignore this.

warp/Network/Wai/Handler/Warp.hs
kazu-yamamoto: For (2), I think that you misunderstand.

`getFileInfo` tries to obtain the file information of the target file. This is main behavior. 

Using the cache or registering to the cache is just for performance purpose only.

In other words, this is just a high-performance `stat(2)` syscall. So, it's strange to have `Maybe` for its semantics.

snoyberg: If that's the case, I think the comments here are misleading. It says that it will check inside the cache, and if file information is not available, it will throw an exception. Perhaps it should read:

> Getting file information from the inside cache.
> If information on the given file is not yet in the cache, information will be looked up via a `stat` syscall.
> If this function is used an a `Request` generated by a WAI backend besides Warp, it will throw an exception.

kazu-yamamoto: Thank you for your advice. I pushed a new comment. I think that this pull request is ready to merge.

PR #477, opened 2015-12-04
Conditional requests

kazu-yamamoto:
These commits implement conditional requests discussed in #359. The behavior is described in the manual of `sendResponse`. The logic is imported from wai-app-file-cgi.

- ETag stuff is pending
- Probably, I will not implement multiple byte ranges because there is no clear benefit at this moment.

@snoyberg Please review.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/File.hs
+++ b/warp/Network/Wai/Handler/Warp/File.hs
@@ -52,13 +52,13 @@ conditionalRequest finfo hs0 reqidx = case condition of
 ----------------------------------------------------------------
 
 ifModifiedSince :: IndexedHeader -> Maybe HTTPDate
-ifModifiedSince reqidx = reqidx ! idxIfModifiedSince >>= parseHTTPDate
+ifModifiedSince reqidx = reqidx ! fromEnum ReqIfModifiedSince >>= parseHTTPDate
 
 ifUnmodifiedSince :: IndexedHeader -> Maybe HTTPDate
-ifUnmodifiedSince reqidx = reqidx ! idxIfUnmodifiedSince >>= parseHTTPDate
+ifUnmodifiedSince reqidx = reqidx ! fromEnum ReqIfUnmodifiedSince >>= parseHTTPDate
 
 ifRange :: IndexedHeader -> Maybe HTTPDate
-ifRange reqidx = reqidx ! idxIfRange >>= parseHTTPDate
+ifRange reqidx = reqidx ! fromEnum ReqIfRange >>= parseHTTPDate
 
 ----------------------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


warp/Network/Wai/Handler/Warp/File.hs
```diff
@@ -0,0 +1,173 @@
...
+data RspFileInfo = WithoutBody H.Status
+                 | WithBody H.Status H.ResponseHeaders Integer Integer
+                 deriving (Eq,Show)
+
+----------------------------------------------------------------
+
+conditionalRequest :: I.FileInfo
+                   -> H.ResponseHeaders -> IndexedHeader
+                   -> RspFileInfo
+conditionalRequest finfo hs0 reqidx = case fromJust mx of
```
snoyberg: I'd rather not include a `fromJust` here. My best preference is to make the call total by making `uncoditional` a fallback (since it appears that it always returns a `Just` value). My second-best preference would be to pattern match on the `Maybe` and use an explicit `error` call that states that `conditionalRequest` failed, since getting a `fromJust` error from a program is maddening.

warp/Network/Wai/Handler/Warp/Header.hs
```diff
@@ -19,27 +19,35 @@ indexRequestHeader hdr = traverseHeader hdr requestMaxIndex requestKeyIndex
 
 idxContentLength,idxTransferEncoding,idxExpect :: Int
 idxConnection,idxRange,idxHost :: Int
-idxContentLength    = 0
```
snoyberg: Not specifically relevant to this change, but I wonder if we can get the same performance with easier to maintain code if we create an enum type here instead of using the `Int`s directly.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -351,6 +285,76 @@ sendRsp conn _ _ _ _ (RspRaw withApp src tickle) = do
     send bs = connSendAll conn bs >> tickle
 
 ----------------------------------------------------------------
+
+-- Sophisticated WAI applications.
+-- We respect s0. s0 MUST be a proper value of either 200 or 206.
```
snoyberg: Why can't we have, for example, a 404 response that streams part of a file? I can't picture actually needing it, but this seems like a strange limitation.

kazu-yamamoto: Ah. 404 is also OK. For instance, a HTML file for 404 is specified. This should be read "s0 MUST be a proper value". I will fix.

warp/Network/Wai/Handler/Warp/Header.hs
kazu-yamamoto: I don't know how performance will change. Does GHC change `fromEnum TransferEncoding` to a constant value when compiling?

warp/Network/Wai/Handler/Warp/File.hs
kazu-yamamoto: A good point. Fixed in the first way.

warp/Network/Wai/Handler/Warp/Header.hs
kazu-yamamoto: OK. I confirmed. Consider the following code:

```haskell
data X = A | B | C | D | E deriving Enum

main :: IO ()
main = print $ 1234 + fromEnum E
```

`ghc -O -ddump-prep` generates 1238 (ie 1234 + 4)!

kazu-yamamoto:
Resolved all issues. Please merge.

Issue #479, opened 2015-12-07
Too weak lower bound on http-types


snoyberg:
```
[23 of 35] Compiling Network.Wai.Handler.Warp.File ( Network/Wai/Handler/Warp/File.hs, .stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/Network/Wai/Handler/Warp/File.o )

/Users/michael/Documents/wai/warp/Network/Wai/Handler/Warp/File.hs:143:37:
    Not in scope: ‘H.hContentRange’
    Perhaps you meant one of these:
      ‘H.hContentLength’ (imported from Network.HTTP.Types.Header),
      ‘H.hContentLength’ (imported from Network.HTTP.Types),
      ‘H.hContentType’ (imported from Network.HTTP.Types.Header)

/Users/michael/Documents/wai/warp/Network/Wai/Handler/Warp/File.hs:160:41:
    Not in scope: ‘H.hAcceptRanges’

--  While building package warp-3.1.9 using:
      /Users/michael/.stack/setup-exe-cache/setup-Simple-Cabal-1.22.4.0-x86_64-osx-ghc-7.10.2 --builddir=.stack-work/dist/x86_64-osx/Cabal-1.22.4.0 build lib:warp --ghc-options " -ddump-hi -ddump-to-file"
```

@kazu-yamamoto Like previously, I'd recommend avoiding usage of the new values, and instead just using the raw strings.

kazu-yamamoto:
I'm OK with acc8f8c710352ba2a35c1bc458ed05f3eb34b088.

snoyberg:
OK, I'll cherry-pick to master.

kazu-yamamoto:
When https://travis-ci.org/yesodweb/wai/builds/95300667 shows all greens, we should release new warp and warp-tls. May I ask you to do so?

snoyberg:
I'd like to wait until I get my new Stack/Cabal split Travis setup going too, any objection? I'll merge that in without your review if you're OK with it.

kazu-yamamoto:
No objection. I'm not in hurry. I guessed you are in hurry because of yesod. You can release whenever you like.

For new Stack/Cabal, I'm OK if the three major GHCs are supported.

snoyberg:
No, there's no hurry from my side either actually, I already put a workaround into yesod-bin for this, so there's no pressure. I just happen to have time to work on this now :)

kazu-yamamoto:
[...]

PR #480, opened 2015-12-07
Switch over to using Stack for building Travis

snoyberg:
There are three reasons for this:

* The build script is significantly simpler
* Build times should go down significantly due to precompiled tools (as
  opposed to building mega-sdist) and caching
* We'll catch bugs that arise due to older versions of libraries (like
  #479)

Downsides:

* It won't always check against the latest versions of dependencies,
  just the latest in Stackage Nightly
* We can't test any GHC before 7.8 with it

It's possible to modify this further to have a hybrid of Stack and cabal
for building.

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -2,6 +2,19 @@
 
 set -eux
 
+# For simplicity, installing cabal-meta and Stack for both the Cabal and Stack
+# build plans, even though they're never both necessary.
+
+mkdir -p ~/.local/bin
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
I don't care downside (1).

But for the downside (2), I would like to support three major versions of GHC and GHC head.

snoyberg:
You're right, I agree. I'll put together some conditional Travis stuff.

Issue #486, opened 2015-12-10
Does not handle correctly multiple Cookie headers in HTTP/2 requests


abailly:
I have started using warp-tls-3.1.3 and encounter an issue with Cookies. When a user-agent send multiple `Cookie` headers, only the first one gets parsed and supernumerary headers are dropped. 

AFAICT this is indeed conformant to [RFC625](http://tools.ietf.org/html/rfc6265#section-5.4). But according to [HTTP/2](https://http2.github.io/http2-spec/#CompressCookie) specification, multiple Cookie headers are allowed. I have googled a bit and found a similar issue in [netty](https://github.com/netty/netty/issues/4457) where I got my understanding of this issue. 

Am I missing something? Is there a way to instruct browser to use HTTP/1.1?

kazu-yamamoto:
Multiple cookie values are concatenated into a single one. If you don't see multiple values in a single cookie, it's a bug. Please give us examples.

abailly:
Tentative unit tests in #487

kazu-yamamoto:
@abailly Can you try https://github.com/yesodweb/wai/tree/http2-starting-over? This branch has efficient code to make multiple cookies to a single on HTTP/2. It will be merged to the master soon but it breaks backward compatibility.

abailly:
Sure. Are the tests I provide meaningful and correct?

abailly:
I tried to run the given unit tests against branch `http2-starting-over` but got same issue.

kazu-yamamoto:
First of all, your test case for HTTP/2 uses HTTP/2, not HTTP/1.1.

Which do you talk about, HTTP/1.1 or HTTP/2?

- HTTP/1.: Multiple cookies are not allowed. Warp does not concatenate multiple cookies into a single one.

- HTTP/2: Clients are still able to send a single cookie. But HTTP/2 layer divided it into multiple so that HPACK works efficiently. HTTP/2 code of Warp concatenates multiple cookies into a single one and passes it to WAI applications.

abailly:
I assume you meant it uses HTTP/1.1 instead of HTTP/2. How can I trigger
use of HTTP/2 ? Sorry I am not familiar with HTTP/2 yet...

I triggered this issue because I have seen Firefox sending multiple cookie
headers so not sure I understand your second comment. I will fix first test
case to make it pass, I agree current behavior is compliant with standards.



Le mercredi 16 décembre 2015, Kazu Yamamoto <notifications@github.com> a
[...]

PR #487, opened 2015-12-14
added test for multiple cookies concatenation

abailly:
Not sure this test is extremely meaningful as it seams the RFC for HTTP/1.1 as noted in #486 disallow multiple cookies. Furthermore I observe this behaviour only when using warp-tls but it seems to be hard to unit test warp-tls as there are currently no tests in this package.


kazu-yamamoto:
This cannot be merged anymore.

Issue #462, opened 2015-11-12
warp-tls: Status code 200 for insecure request denial


michael-swan:
It's really a minor detail, but perhaps a different status code such as 403 would be more fitting for a request denial due to an insecure request being made.

See https://github.com/yesodweb/wai/blob/master/warp-tls/Network/Wai/Handler/WarpTLS.hs#L465

reiddraper:
I believe the correct status code is HTTP 426. See [RFC 2817](http://tools.ietf.org/html/rfc2817#section-4.2).

snoyberg:
Sounds like a good change to me, can you send a PR?

kazu-yamamoto:
:+1:

kazu-yamamoto:
The current code hard-codes "HTTP/1.1" and "TLS/1.0". What if "HTTP/2" and "TLS/1.1" or later?

I discussed this issue with my friends. I seems to us that RFC 2817 recommends 426 status code to redirect from 80 port to 443 port. But in this case, WarpTLS is listening to 443 port.

I would suggest another error status code for HTTP/1.1 to avoid this hard coding. I believe that `Content-Length:` is also necessary.

If we wish to distinguish HTTP/1.1 and HTTP/2 based on the HTTP/2 prefix, `GOAWAY +  INADEQUATE_SECURITY` would be suitable for HTTP/2. But I don't know it's worth implementing since this is a rare error case.

PR #490, opened 2015-12-16
Draft of solution for #462

tolysz:
It still needs to get the appropriate upgrade string. Please advice if we are going to hard-code it or build according to servers capabilities.
https://github.com/yesodweb/wai/issues/462


kazu-yamamoto:
I have merged this into the `wai-3.0-warp-3.1` branch. 

We need to support HTTP/2, too, in the future.

PR #494, opened 2015-12-21
wai-handler-launch: Improvements to ping's javascript script.

mihaigiurgeanu:
-   Avoid using the browser's cache for the response. 
  
  Because the ping request uses GET method, IE returns the response from
  cache and does not send the ping request to the server causing the server 
  to shut down.
-   Make the javascript use the asynchronous ajax request instead of
  synchronous request. 
  
  This fixes the warning displayed in Chrome: `Synchronous XMLHttpRequest on the
  main thread is deprecated because of its detrimental effects to the end
  user's experience. For more help, check 
  http://xhr.spec.whatwg.org/.`


snoyberg:
Sorry for the delay on this, it slipped through the cracks somehow. Thank you!

Issue #503, opened 2016-01-06
Handling of missing host header field


sir-murray:
http://tools.ietf.org/html/rfc7230#section-5.4

> A server MUST respond with a 400 (Bad Request) status code to any
> HTTP/1.1 request message that lacks a Host header field and to any
> request message that contains more than one Host header field or a
> Host header field with an invalid field-value.

To test I used this:

```haskell
{-# LANGUAGE OverloadedStrings     #-}
{-# LANGUAGE QuasiQuotes           #-}
{-# LANGUAGE TemplateHaskell       #-}
{-# LANGUAGE TypeFamilies          #-}
import Yesod
import Data.Text.Encoding (decodeUtf8)
import Network.Wai (requestHeaders)

data Test = Test

mkYesod "Test" [parseRoutes|
/ HomeR GET
|]

instance Yesod Test

getHomeR :: Handler Html
getHomeR = do
    r <- waiRequest
    let host = maybe "[None]" decodeUtf8 (lookup "host" . requestHeaders $ r)
    defaultLayout [whamlet|#{host}|]

main :: IO ()
main = warp 3000 Test
```

Missing Host:
```
λ nc localhost 3000
GET / HTTP/1.1

HTTP/1.1 200 OK
Date: Wed, 06 Jan 2016 10:28:57 GMT
Server: Warp/3.1.3.1 + Yesod/1.4.15 (core)
Content-Length: 76
Content-Type: text/html; charset=utf-8
Set-Cookie: _SESSION=zXJ8hMJRzjdLbu0FUm6NcbUFNycmuMf7OUbOnspmtYz4L1OwTiyfGhTiUWLhGuAomC7EK6lrO/0Tx03nHUIn4IVTMk7G/jaVmeX1CaWbaFsUu2RxmcJA2LSOUatYQtkiquK+dQ9BOWI=; Path=/; Expires=Wed, 06-Jan-2016 12:28:49 GMT; HttpOnly
Vary: Accept, Accept-Language

<!DOCTYPE html>
<html><head><title></title></head><body>[None]</body></html>
```

More than one host header:
```
λ nc localhost 3000
GET / HTTP/1.1
Host: dave
Host: lisa

HTTP/1.1 200 OK
Date: Wed, 06 Jan 2016 10:33:25 GMT
Server: Warp/3.1.3.1 + Yesod/1.4.15 (core)
Content-Length: 74
Content-Type: text/html; charset=utf-8
[...]

PR #504, opened 2016-01-06
warp: validate request host header field

sir-murray:
See #503

### DIFF
--- a/warp/Network/Wai/Handler/Warp/RequestHeader.hs
+++ b/warp/Network/Wai/Handler/Warp/RequestHeader.hs
@@ -12,6 +12,7 @@ import qualified Data.ByteString.Char8 as B (unpack)
 import Data.ByteString.Internal (ByteString(..), memchr)
 import qualified Data.CaseInsensitive as CI
 import Data.Word (Word8)
+import Data.List (foldl')
 import Foreign.ForeignPtr (withForeignPtr)
 import Foreign.Ptr (Ptr, plusPtr, minusPtr, nullPtr)
 import Foreign.Storable (peek)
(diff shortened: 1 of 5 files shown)


warp/Network/Wai/Handler/Warp/RequestHeader.hs
```diff
@@ -35,8 +35,10 @@ parseHeaderLines [] = throwIO $ NotEnoughLines []
 parseHeaderLines (firstLine:otherLines) = do
     (method, path', query, httpversion) <- parseRequestLine firstLine
     let path = H.extractPath path'
-        hdr = map parseHeader otherLines
-    return (method, path', path, query, httpversion, hdr)
+        (hdr, hostCount) = foldr parseValidHeader ([], 0) otherLines
+    if hostCount == 1
```
meteficha: Now that the result of fold is being used strictly, perhaps it should be changed into a `foldl'`?

meteficha:
Very nice! :+1: :shipit:

warp/Network/Wai/Handler/Warp/RequestHeader.hs
sir-murray: I don't know, I'm fairly new to Haskell so if you say so I'll change it.

sir-murray:
@meteficha Changed it to foldl'

kazu-yamamoto:
We need to be very careful about this kind of change. This patch introduce one more traversal of headers with the very heavy operation (i.e. (==)). Moreover, this patch does not cover HTTP/2.

- First, we should continue to discuss whether or not we implement this feature in #503.
- If we agree, we need to find the best place for this traversal. I guess `traverseHeader` is a candidate for HTTP/1.1.
- Confirm that this change does not introduce performance regression.

sir-murray:
ok, will close this for now.

PR #505, opened 2016-01-09
Fixed a typo for windows version, changed h to hdl

reygoch:
There was an error / typo which prevented warp 3.2.1 from compiling on windows.

It would be nice if you could merge this soon, since I have some project to finish and I want to have it all nice and tight without manually installing cabal packages.

    Network\Wai\Handler\Warp\HTTP2\Sender.hs:299:14: Not in scope: `h'


kazu-yamamoto:
Merged. Thanks.

PR #508, opened 2016-01-11
compress favicon.ico

knupfer:
This reduces complaints about compression from performance testing websites for wai using websites (especially yesod powered).


snoyberg:
Thanks!

PR #513, opened 2016-02-23
warp: make .ghci work with stack

soenkehahn:
This change makes it possible to load the test-suite of `warp` into `ghci` while using `stack`. This is currently broken.

I don't expect this to be merged as is. Here's some questions:

- Do we want a `.ghci` file that works both with `stack` and `cabal-install`? Or can we just go with `stack`? Or `cabal-install`?
- It would be nice to have a way to make `ghc` find the `Paths_warp` module without relying on the processor architecture, the operating system and the cabal version. I imagine if any of those change, this'll break. Is there a way to do that?

### DIFF
--- a/warp/.ghci
+++ b/warp/.ghci
@@ -1,1 +1,1 @@
-:set -itest -idist/build/autogen -optP-include -optPdist/build/autogen/cabal_macros.h
+:set -itest -idist/build/autogen -i.stack-work/dist/x86_64-linux/Cabal-1.22.5.0/build/autogen


snoyberg:
I don't actually have any input on this, as I hardly ever use GHCi (yes, I know, I should). @kazu-yamamoto do you have any thoughts?

kazu-yamamoto:
I don't know how to combine GHCi and stack. This `.ghci` works perfectly for me without stack. I would prefer to not change it.

soenkehahn:
After doing `stack init ; stack test` once in the project you can just say `stack exec ghci` inside the warp directory. That's what I do.

But I don't want to convince you to use `stack`. :) The problem is that the current `.ghci` does not work for me because I'm using `stack`. If I tried to come up with a files that works for both `cabal-install` and `stack` would you be willing to merge that?

kazu-yamamoto:
> If I tried to come up with a files that works for both cabal-install and stack would you be willing to merge that?

Of course!

soenkehahn:
Updated this PR. Worked locally with both a cabal sandbox and stack.

kazu-yamamoto:
Why did you remove `-optP-include`?

soenkehahn:
Two reasons:

1) `cabal_macros.h` doesn't seem to be used in `warp`.
2) The include makes `stack` choke as it doesn't create the file at this path.

kazu-yamamoto:
OK. Let's merge this and see what will happen.

soenkehahn:
Thanks!

PR #516, opened 2016-03-02
warp: add withApplication and openFreePort

soenkehahn:
@snoyberg: You mentioned that this could be implemented using `async`. I used it in one place now, but I don't see how anything else could be replaced by operations from `async`. But I'm also not too familiar with that package.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/WithApplication.hs
+++ b/warp/Network/Wai/Handler/Warp/WithApplication.hs
@@ -3,6 +3,7 @@ module Network.Wai.Handler.Warp.WithApplication (
   withApplication,
   testWithApplication,
   openFreePort,
+  withFreePort,
 ) where
 
 import           Control.Concurrent
@@ -16,8 +17,7 @@ import           Network.Wai.Handler.Warp.Types
 data App
   = App {
     appThread :: ThreadId,
-    appKilled :: Waiter (),
-    appPort :: Int
+    appKilled :: Waiter ()
   }
 
 -- | Runs the given 'Application' on a free port. Passes the port to the given
@@ -26,25 +26,26 @@ data App
 withApplication :: IO Application -> (Port -> IO a) -> IO a
 withApplication mkApp action = do
   app <- mkApp
-  bracket (acquire app) free (\ runningApp -> action (appPort runningApp))
+  withFreePort $ \ (port, sock) -> do
+    bracket (start app sock) stop $ \ _ ->
+      action port
   where
-    acquire :: Application -> IO App
-    acquire app = do
-      start <- mkWaiter
+    start :: Application -> Socket -> IO App
+    start app sock = do
+      started <- mkWaiter
       killed <- mkWaiter
       thread <- forkIO $ do
-        (port, sock) <- openFreePort
         let settings =
               defaultSettings{
-                settingsBeforeMainLoop = notify start port
+                settingsBeforeMainLoop = notify started ()
               }
         runSettingsSocket settings sock app
           `finally` notify killed ()
-      port <- waitFor start
-      return $ App thread killed port
+      waitFor started
+      return $ App thread killed
 
-    free :: App -> IO ()
-    free runningApp = do
+    stop :: App -> IO ()
+    stop runningApp = do
       killThread $ appThread runningApp
       waitFor $ appKilled runningApp
 
@@ -92,3 +93,7 @@ openFreePort = do
   listen s 1
   port <- socketPort s
   return (fromIntegral port, s)
+
+-- | Like 'openFreePort' but closes the socket before exiting.
+withFreePort :: ((Port, Socket) -> IO a) -> IO a
+withFreePort = bracket openFreePort (close . snd)
--- a/warp/Network/Wai/Handler/Warp/WithApplication.hs
+++ b/warp/Network/Wai/Handler/Warp/WithApplication.hs
@@ -1,6 +1,7 @@
 
 module Network.Wai.Handler.Warp.WithApplication (
   withApplication,
+  testWithApplication,
   openFreePort,
 ) where
 
@@ -16,15 +17,12 @@ data App
   = App {
     appThread :: ThreadId,
     appKilled :: Waiter (),
-    appExceptionMVar :: MVar (Maybe SomeException),
     appPort :: Int
   }
 
 -- | Runs the given 'Application' on a free port. Passes the port to the given
 -- operation and executes it, while the 'Application' is running. Shuts down the
 -- server before returning.
---
--- Handy for e.g. testing 'Application's over a real network port.
 withApplication :: IO Application -> (Port -> IO a) -> IO a
 withApplication mkApp action = do
   app <- mkApp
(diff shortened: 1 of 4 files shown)


snoyberg:
* I take back my comments on async. For this level of usage (just async and cancel), there's no real advantage over using forkIO and killThread, and that avoids the extra dependency
* Are you looking into the Travis failures?

Thanks!

soenkehahn:
Yep, I'm on it.

soenkehahn:
I removed the use of `async` and fixed the PR for older ghcs. The current travis failure is not related to my changes, as far as I can tell.

snoyberg:
I'm a little concerned about the exception catching from the application itself. The exception is being put into an `MVar`, but:

* The exception doesn't seem to be used
* It gets in the way of warp's standard exception handling behavior, which could be confusing
* It's quite a separate concern from the rest of the code (meaning: you could write a general-purpose exception catching mechanism)
* It seems like the second exception will result in a blocked indefinitely on MVar exception

soenkehahn:
@snoyberg: Thanks a lot for looking into this.

> The exception doesn't seem to be used

It's being rethrown here: https://github.com/yesodweb/wai/pull/516/files#diff-7c5504af069c10f4614395e3a2978e70R56

> It gets in the way of warp's standard exception handling behavior, which could be confusing

Yes, very good point.

Some context: This behavior is motivated by my primary use-case: testing wai applications. If an `Application` throws an exception, ideally the test-case would fail with that exception. If we don't catch and rethrow the exception in the executing thread the test-case may fail with a less specific error message. The actual exception will be printed to `stderr`, but might get lost in the rest of the output of the test-suite. (Or might even be silenced with `silently`.)

But I completely get your point. This is not at all what we want or expect from a function meant for production. So what about putting `withApplication` into a test module (`Network.Wai.Handler.Warp.Test` or maybe even `Test.Warp`) and adding a comment about the exception behavior and that it's not suitable for production use? (My preferred solution so far.)

Alternatively we could revert to `warp`'s exception behavior. But personally I don't see a reason to use this function outside of testing, so it'd be strictly less useful for me.

> It's quite a separate concern from the rest of the code (meaning: you could write a general-purpose exception catching mechanism)

As I understand you can't with the usual exception handling mechanisms (`catch`, etc.), since the `warp` runner is being forked. Can you?

> It seems like the second exception will result in a blocked indefinitely on MVar exception

Not sure, what you mean. What is the second exception? Which MVar operation would throw the `blocked indefinitely` exception?

snoyberg:
For the test case, why can't you do something like this:

```haskell
tid <- myThreadId
let app req send = origApp req send `catch` \e -> do
        throwTo tid (e :: SomeException)
        throwIO e
```

And then use `app` in place of `origApp`?

snoyberg:
For the test case, why can't you do something like this:

```haskell
tid <- myThreadId
let app req send = origApp req send `catch` \e -> do
        throwTo tid (e :: SomeException)
        throwIO e
```

And then use `app` in place of `origApp`?

soenkehahn:
Ah, yes, this is how this could be implemented with rather common primitives.

I've thought about this some more. The current behavior regarding exceptions (in my version of `withApplication`) is what I want in all my use-cases. If we make `withApplication` have `warp`'s normal exception behavior I'd probably end up not using it. So I guess there's two options:

- We put it in a test module as proposed above. What do you think about that?
- I put it in some other package. (Any suggestions welcome.) I might even whip up a new package, though I'm trying to avoid that.

Any suggestions for different solutions are also very welcome.

snoyberg:
I'd be open to exposing this functionality from the main module, if:

* It was under a name that made it clear that it is for testing, or that it crashes on an exception
* There is the non-exception version available too to avoid confusion

I actually have very concrete use cases for the non-exception version of this, such as providing a statistics-serving endpoint for an application. As an end user, I would be confused if the `withApplication` name implied automatically shutting down my entire application if a handler threw an exception.

soenkehahn:
Ok, I'll create two versions of `withApplication` (tomorrow, hopefully).

Are you talking about something similar to http://hackage.haskell.org/package/ekg ? That makes sense...

snoyberg:
Something similar, yes, but targeted at developer-defined statistics.

soenkehahn:
Done.

warp/Network/Wai/Handler/Warp/WithApplication.hs
```diff
@@ -0,0 +1,94 @@
...
+withApplication mkApp action = do
+  app <- mkApp
+  bracket (acquire app) free (\ runningApp -> action (appPort runningApp))
+  where
+    acquire :: Application -> IO App
+    acquire app = do
+      start <- mkWaiter
+      killed <- mkWaiter
+      thread <- forkIO $ do
+        (port, sock) <- openFreePort
```
snoyberg: Is there a reason that this is done inside the forked thread instead of the parent thread? I'm worried about the situation where `openFreePort` throws an exception.

warp/Network/Wai/Handler/Warp/WithApplication.hs
```diff
@@ -0,0 +1,94 @@
...
+        let settings =
+              defaultSettings{
+                settingsBeforeMainLoop = notify start port
+              }
+        runSettingsSocket settings sock app
+          `finally` notify killed ()
```
snoyberg: There's a slight async exception hole here (besides the `openFreePort` exception I mentioned above). How about using `forkFinally` instead, e.g.:

```haskell
(port, sock) <- openFreePort
let settings = defaultSettings {...}
thread <- forkFinally
  (runSettingsSocket settings sock app)
  (const (notify killed ())
```

snoyberg: And this reminded me of what I'd meant originally about using async for this:

```haskell
app <- mkApp
(port, sock) <- openFreePort
let settings = defaultSettings {...}
res <- race (runSettingsSocket settings sock app) (action port)
case res of
  Left () -> error "Unexpected: runSettingsSocket exited"
  Right x -> return x
```

And after writing _that_ out, I realized: the `openFreePort` call actually needs to be wrapped in a `bracket` to close the Socket afterwards.

soenkehahn:
Good points. I've introduced `withFreePort` that makes sure the socket is closed afterwards. And I'm using `race` now, which felt a bit counter-intuitive to me, but it looks much cleaner now, I think.

snoyberg:
Awesome, thanks!

soenkehahn:
Thanks for the thorough review and for merging!

soenkehahn:
Could we get a release for this, please? (@snoyberg)

snoyberg:
@kazu-yamamoto any objection with me cutting a release of warp from master?

kazu-yamamoto:
I would also like to include de909fd20345e445b8b3c56ccca443f481566d34 which is a terrible bug.

@snoyberg Please go ahead if https://travis-ci.org/yesodweb/wai/builds/119141853 displays all green.

snoyberg:
Uploaded. The OS X failures are due to FastCGI no longer being available
from Homebrew. I think I'll just disable FastCGI in our Travis test suite.

soenkehahn:
Great, thanks a lot!

PR #518, opened 2016-03-10
Case insensitive multipart request header lookup

kellymclaughlin:
Treat `Content-Type` and `Content-Disposition` header names as case
insensitive when parsing the parts of multipart requests. These names
were previously treated as case sensitive which causes parsing problems
when attempting to parse requests that may use all lower case header
names.

I hit this problem yesterday when trying to process a multipart request
composed with a library that sends all the part header field names as 
lower case. Most libraries use the expected camel case for the header names, but 
RFC 2616 and RFC 7230 both say the header field names are case insensitive so 
I made this small change to address it.


snoyberg:
LGTM, thanks!

PR #519, opened 2016-03-12
wai-app-static: use cryptonite instead of cryptohash

myfreeweb:
`cryptohash` is deprecated in favor of `cryptonite`. Not marked as such on Hackage yet, only [on GitHub](https://github.com/vincenthz/hs-cryptohash). But `tls` uses `cryptonite` already, so these two packages already clash (they export the exact same API).

Also, `memory` instead of `base64-bytestring` for Base64 encoding (`memory` is a dependency of `cryptonite`).

Yes, `conduit` is also dropped, because `cryptohash-conduit` wasn't updated in a long time, and wasn't ported to `cryptonite`. I think it's perfectly fine to use lazy I/O here, for the sake of fewer dependencies.

### DIFF
--- a/wai-app-static/WaiAppStatic/Storage/Filesystem.hs
+++ b/wai-app-static/WaiAppStatic/Storage/Filesystem.hs
@@ -23,10 +23,9 @@ import WaiAppStatic.Listing
 import Network.Mime
 import System.PosixCompat.Files (fileSize, getFileStatus, modificationTime, isRegularFile)
 import Data.Maybe (catMaybes)
-import qualified Crypto.Hash.Conduit (hashFile)
-import Data.Byteable (toBytes)
-import Crypto.Hash (MD5, Digest)
-import qualified Data.ByteString.Base64 as B64
+import Data.ByteArray.Encoding
+import Crypto.Hash (hashlazy, MD5, Digest)
+import qualified Data.ByteString.Lazy as BL (readFile)
 import qualified Data.Text as T
 
 -- | Construct a new path from a root and some @Pieces@.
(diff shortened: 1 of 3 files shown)


wai-app-static/WaiAppStatic/Storage/Filesystem.hs
```diff
@@ -120,8 +119,8 @@ webAppLookup hashFunc prefix pieces =
 -- exists.
 hashFile :: FilePath -> IO ByteString
 hashFile fp = do
-    h <- Crypto.Hash.Conduit.hashFile fp
-    return $ B64.encode $ toBytes (h :: Digest MD5)
+    f <- BL.readFile fp
```
snoyberg: This has the potential to leak FDs. A better approach would be `withBinaryFile`, which will guarantee that the FD is closed.

snoyberg:
The Travis build did not pass, can you resolve that before merge? Otherwise, the change seems good to me.

wai-app-static/WaiAppStatic/Storage/Filesystem.hs
myfreeweb: Lazy ByteString's `readFile` closes the FD ("The Handle will be held open until EOF is encountered"). I've tried `withBinaryFile`, it results in `hGetBufSome: illegal operation (handle is closed)`

myfreeweb:
Works fine now with lower dependency bounds.

wai-app-static/WaiAppStatic/Storage/Filesystem.hs
snoyberg: The correct way to do this is something like:

```haskell
withBinaryFile fp ReadMode $ \h -> do
    lbs <- L.hGetContents h
    ...
```

The problem with `readFile` rears its head in the case of async exceptions. It's an uncommon case to be sure, but I try to avoid the lazy I/O functions in libraries anyway to prevent confusing behavior.

myfreeweb:
Oh – it was closed because the whole function was lazy.

Updated to use `BangPatterns`, I think it's correct now.

snoyberg:
Thanks!

PR #520, opened 2016-03-12
Add nix dependencies to stack.yaml

chris-martin:
This enables building on NixOS with `stack --nix`!

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -13,3 +13,8 @@ packages:
 - ./auto-update
 extra-deps:
 - http2-1.5.0
+nix:
+  enable: false
+  packages:
+  - fcgi
+  - zlib


snoyberg:
Unfortunately this change breaks non-Nix usage of wai (see the Travis build failures). How about including the changes in a commented-out section of the stack.yaml file?

chris-martin:
Whoops, I didn't think that section would make any difference unless Nix support was explictily enabled with `nix.enable` or the `--nix` flag. Is there really no better way to do that? :frowning:

snoyberg:
Maybe you can set nix.enable to false?

chris-martin:
Trying that, waiting on the build now.

chris-martin:
Looks like that worked!

PR #529, opened 2016-04-10
fix doctests when 'network-data' and 'iproute' are installed

trofi:
Before the patch tests failed for me as:

  Network/Wai/Handler/Warp/Run.hs:22:8:
    Ambiguous module name ‘Data.IP’:
      it was found in multiple packages:
      network-data-0.5.3@netwo_G5ThtIF3NpS8U3d3JAsmlG iproute-1.7.0@iprou_5PoZLY0zggKF8zyZ5Xj1ys

Patch explicitly lists all the packages for ghci/doctest run.

When added dependencies found missing 'template-haskell' depend.

Signed-off-by: Sergei Trofimovich <siarheit@google.com>

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -1,9 +1,10 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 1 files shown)


snoyberg:
How about just using PackageImports for the problematic imports?

trofi:
> How about just using PackageImports for the problematic imports?

Sounds even better! Added `PackageImports` import in updated patch.

snoyberg:
Thanks!

PR #531, opened 2016-04-12
Add {test,}withApplicationSettings

utdemir:
`withApplication` was using `defaultSettings` implicitly, this patch implements `withApplicationSettings` and `testWithApplicationSettings` functions that run `Application`'s with given `Settings`.

Without this patch, I was unable to increase Warp's timeout on `testWithApplication`.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/WithApplication.hs
+++ b/warp/Network/Wai/Handler/Warp/WithApplication.hs
@@ -1,7 +1,9 @@
 
 module Network.Wai.Handler.Warp.WithApplication (
   withApplication,
+  withApplicationSettings,
   testWithApplication,
+  testWithApplicationSettings,
   openFreePort,
   withFreePort,
 ) where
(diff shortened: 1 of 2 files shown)


warp/Network/Wai/Handler/Warp/WithApplication.hs
```diff
@@ -21,13 +23,18 @@ import           Network.Wai.Handler.Warp.Types
 --
 -- @since 3.2.4
 withApplication :: IO Application -> (Port -> IO a) -> IO a
-withApplication mkApp action = do
+withApplication = withApplicationSettings defaultSettings
+
+-- | 'withApplication' with given 'Settings'
```
snoyberg: It's worth including a note that any port provided in the `Settings` value would be ignored.

Also, can you add a `@since` comment for version 3.2.7 to this and the function added below, and update the `ChangeLog.md` file?

utdemir:
@snoyberg I updated the PR, can you check again if everything is okay?

warp/Network/Wai/Handler/Warp/WithApplication.hs
```diff
@@ -61,7 +77,7 @@ testWithApplication mkApp action = do
 
 data Waiter a
   = Waiter {
-    notify :: a -> IO (),
+    notify  :: a -> IO (),
```
snoyberg: I think this part of the diff was accidentally added.

snoyberg:
LGTM, just waiting on Travis before merging.

snoyberg:
Thanks for the quick turnaround on updating the PR.

warp/Network/Wai/Handler/Warp/WithApplication.hs
utdemir: That was stylish-haskell-mode, fixed now. Thanks.

snoyberg:
Thanks!

kazu-yamamoto:
Warp 3.2.7 has been released.

PR #534, opened 2016-04-19
Updated Mime.hs for less (css) support

tommyschnabel:


snoyberg:
Do you have a source for `text/less` being the correct mime type for Less?

tommyschnabel:
Well it seems like I've been doing things wrong for about a year, since according to the [less website](http://lesscss.org/) its `text/css`. I updated my commit to correct my error

snoyberg:
I don't see any mention of mime types on their site, just the `@type` attribute of the `<link>` tag, which may follow different rules.

tommyschnabel:
Ah yeah, I guess their website doesn't have a specification for less files. I have `mime-types` installed on my machine with this commit and it server up less files correctly, but I understand that that is no replacement for official specs. 

This is really just for development, nobody uses uncompiled less files in production, so its not a big deal if this doesn't get merged.

snoyberg:
The reason this caught my eye is that it's surprising to see a non-standard file format get a non-x-prefixed mimetype. I would have expected text/x-less or something like that. I don't really have any opinion against this, but since it's a non-standard file format, I'd like at least some source, even semi-official, to reference as the justification for the decision here.

tommyschnabel:
I don't know if you would consider this semi-official or not, but here there are [67 votes (as of right now) on Stack Overflow](http://stackoverflow.com/questions/7319555/how-to-add-less-to-iis-7-0) for an answer recommending `text/css` for less files.

snoyberg:
Good enough for me. Mind updating this PR with a comment providing a link to that SO question and this discussion? Then I'm fine with this

MaxGabriel:
Is that SO answer specific to Microsoft IIS? One of the commenters on that answer also said:

> Please note that your webserver may return HTTP error code 406 - not acceptable if you define the mime type as text/css. You should use text/less as the mimetype.

It just seems really strange to me that LESS would have the same MIME type as CSS.

tommyschnabel:
Since all of this is pretty up in the air, I went ahead and tried it out with `text/less`, and Firefox didn't like it

tommyschnabel:
@snoyberg I'm not sure what you mean, did you mean in the commit message or the PR description, or something else?

snoyberg:
I meant a comment in the source code itself

tommyschnabel:
Oh yeah that makes a lot more sense. Updated my changes

snoyberg:
Perfect, thanks!

PR #540, opened 2016-04-22
Use hAcceptEncoding from http-types

psibi:


snoyberg:
Requiring such a new http-types is restrictive, as evidenced by the Travis failure. Is this new a version actually necessary?

psibi:
> Is this new a version actually necessary?

Yes, it's version `0.9` of `http-types`. 

I'm closing the PR. This can be fixed by CPP, but that just makes the problem worse IMO.

Issue #537, opened 2016-04-21
wai-handler-launch should not launch a browser if server startup failed ?


simonmichael:
In https://github.com/simonmichael/hledger/issues/226, if server startup fails, eg because of the port being in use, I don't want to show users a useless browser window. 

I think we could propagate exceptions from the forked server thread back to the parent, which would prevent the launch. And/or, we could expose the monitored server starting and the cross-platform browser launching as separate features, allowing more control.

snoyberg:
Sounds like a good change, either setOnBeforeMainLoop or the new
withApplication function may be useful. Interested in sending a PR?

simonmichael:
Here's one that seems to work. I hope adding async is acceptable..

snoyberg:
Resolved by PR #541

PR #541, opened 2016-04-22
don't launch if server fails; kill server on exit #537

simonmichael:
Make the server thread and launch/monitor thread more aware of each
other's failures, using async. Now if the server fails to start we won't
launch a browser. Also when the main thread terminates (eg from ctrl-c
in GHCI), we make sure the server thread does too. There is now a 0.1s
delay before launching the browser, which also helps ensure the server
is ready to serve.

### DIFF
--- a/wai-handler-launch/Network/Wai/Handler/Launch.hs
+++ b/wai-handler-launch/Network/Wai/Handler/Launch.hs
@@ -34,9 +34,9 @@ import Data.Streaming.Blaze (newBlazeRecv, defaultStrategy)
 import qualified Data.Streaming.Zlib as Z
 
 ping :: IORef Bool -> Middleware
-ping  var app req sendResponse
+ping  active app req sendResponse
     | pathInfo req == ["_ping"] = do
-        liftIO $ writeIORef var True
+        liftIO $ writeIORef active True
         sendResponse $ responseLBS status200 [] ""
     | otherwise = app req $ \res -> do
         let isHtml hs =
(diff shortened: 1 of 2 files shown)


snoyberg:
I don't like relying on timing assumptions to make this work. Warp has a setting to run some code after initialization but before the main loop, which would allow handling this more robustly

simonmichael:
Great! PR updated.

wai-handler-launch/Network/Wai/Handler/Launch.hs
```diff
@@ -192,20 +193,41 @@ runUrlPort = runHostPortUrl "*4"
...
+    withAsync
+      -- server thread, will be killed if we terminate
+      (Warp.runSettings settings (ping active app))
+      (\a -> do
+          -- launch & monitor thread, reraises exceptions from server thread
+          link a
+
+          -- launch the browser, after waiting until the server
+          -- signals that it's ready (or raises a startup exception)
+          let wait = do
```
snoyberg: Why use an IORef here? If you use an MVar, then you can get blocking behavior and not need to worry about polling.

simonmichael: Unfamiliarity/doubt about exception handling. It seems to work, PR updated...

wai-handler-launch/Network/Wai/Handler/Launch.hs
```diff
@@ -192,20 +193,36 @@ runUrlPort = runHostPortUrl "*4"
...
-    loop x
+    ready <- newEmptyMVar
+    active <- newIORef True
+    let settings =
+          Warp.setPort port $
+          Warp.setOnException (\_ _ -> return ()) $
+          Warp.setHost (fromString host) $
+          Warp.setBeforeMainLoop (putMVar ready ()) $
+          Warp.defaultSettings
+    withAsync
```
snoyberg: Sorry to do these piece-meal, but I only notice these things after each successive issue has been addressed.

I think this `withAsync`+`link` is identical in behavior to `fmap (either id id) $ race (Warp.runSettings ...) (takeMVar ready ...)`

simonmichael:
Nice! I appreciate the review.

snoyberg:
Alright, I'm out of ideas :). I'll wait for Travis to check before merging, but LGTM.

PR #542, opened 2016-04-23
Minor improvements to auto-update docs

MaxGabriel:
General changes:
* Adds example code at the top of the module
* Where possible, uses single quotes to create hyperlinks to individual functions, and double quotes for modules
* (Minor) Uses `@since` Haddock syntax

Specific changes are marked in the PR

I'm working on making similar changes for Control.Reaper, but coming up example code for that module is much trickier.

### DIFF
--- a/auto-update/Control/AutoUpdate.hs
+++ b/auto-update/Control/AutoUpdate.hs
@@ -1,10 +1,17 @@
--- | In a multithreaded environment, running actions on a regularly scheduled background thread can dramatically improve performance. 
--- For example, web servers need to return the current time with each HTTP response. 
--- For a high-volume server, it's much faster for a dedicated thread to run every second, and write the current time to a shared 'IORef', than it is for each request to make its own call to 'getCurrentTime'.
+-- | In a multithreaded environment, running actions on a regularly scheduled
+-- background thread can dramatically improve performance.
+-- For example, web servers need to return the current time with each HTTP response.
+-- For a high-volume server, it's much faster for a dedicated thread to run every
+-- second, and write the current time to a shared 'IORef', than it is for each
+-- request to make its own call to 'getCurrentTime'.
 --
--- But for a low-volume server, whose request frequency is less than once per second, that approach will result in /more/ calls to 'getCurrentTime' than necessary, and worse, kills idle GC.
+-- But for a low-volume server, whose request frequency is less than once per 
+-- second, that approach will result in /more/ calls to 'getCurrentTime' than 
+-- necessary, and worse, kills idle GC.
 --
--- This library solves that problem by allowing you to define actions which will either be performed by a dedicated thread, or, in times of low volume, will be executed by the calling thread.
+-- This library solves that problem by allowing you to define actions which will
+-- either be performed by a dedicated thread, or, in times of low volume, will 
+-- be executed by the calling thread.
 --
 -- Example usage:
 --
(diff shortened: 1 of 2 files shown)


auto-update/Control/AutoUpdate.hs
```diff
@@ -1,21 +1,33 @@
...
+-- | In a multithreaded environment, running actions on a regularly scheduled background thread can dramatically improve performance. 
+-- For example, web servers need to return the current time with each HTTP response. 
+-- For a high-volume server, it's much faster for a dedicated thread to run every second, and write the current time to a shared 'IORef', than it is for each request to make its own call to 'getCurrentTime'.
 --
--- This library allows you to define actions which will either be
--- performed by a dedicated thread or, in times of low volume, will be
--- executed by the calling thread.
+-- But for a low-volume server, whose request frequency is less than once per second, that approach will result in /more/ calls to 'getCurrentTime' than necessary, and worse, kills idle GC.
+--
+-- This library solves that problem by allowing you to define actions which will either be performed by a dedicated thread, or, in times of low volume, will be executed by the calling thread.
```
MaxGabriel: So I'm not super confident this new description is better, but I thought the old one was tricky to understand if you weren't familiar with the problem space:

* Reader needs to know why web requests need the current time
* Reader needs to be familiar with concept of 1 thread writing to a shared value, and others reading it

New version:

* Clarifies that auto-updated is meant for multithreaded programs
* Explains the requirement to return the time with each HTTP response (it should probably be more explicit and say HTTP response headers)
* Is more verbose :(

(Again, not certain the new version is better)

auto-update/Control/AutoUpdate.hs
```diff
@@ -1,21 +1,33 @@
...
+-- For more examples, <http://www.yesodweb.com/blog/2014/08/announcing-auto-update see the blog post introducing this library>.
 module Control.AutoUpdate (
       -- * Type
       UpdateSettings
     , defaultUpdateSettings
       -- * Accessors
+    , updateAction
     , updateFreq
     , updateSpawnThreshold
-    , updateAction
```
MaxGabriel: I moved `updateAction` to the top of the accessors, since that's the most important one. `updateSpawnThreshold` is at the bottom since it's unused.

auto-update/Control/AutoUpdate.hs
```diff
@@ -39,44 +51,44 @@ defaultUpdateSettings = UpdateSettings
 
 -- | Settings to control how values are updated.
 --
--- This should be constructed using @defaultUpdateSettings@ and record
+-- This should be constructed using 'defaultUpdateSettings' and record
 -- update syntax, e.g.:
 --
 -- @
--- let set = defaultUpdateSettings { updateAction = getCurrentTime }
+-- let settings = 'defaultUpdateSettings' { 'updateAction' = 'Data.Time.Clock.getCurrentTime' }
```
MaxGabriel: `settings` is a more clear variable name, since `set` is usually used as a verb in programming

auto-update/Control/AutoUpdate.hs
psibi: @MaxGabriel It would be easy to see the changes if the comments were around 80 line limit like it was previously there.

MaxGabriel: Personally I'm against 80 line limits, because they make prose really hard to edit, but I made that change

snoyberg:
This is great, thanks!

PR #544, opened 2016-04-25
don't use deprecated CRT functions on Windows

hreinhardt:
Due to changes in the linker in GHC 8 these deprecated functions are not available anymore.

See:
https://github.com/ghc/ghc/commit/90538d86af579595987826cd893828d6f379f35a
https://msdn.microsoft.com/en-us/library/ms235384(v=vs.80).aspx


snoyberg:
Thanks!

Issue #536, opened 2016-04-21
wai-extra Gzip doesn't appear to compress output


axman6:
I'm building a simple web app which produces highly compressible data as CSVs using the servant-cassava package (uncompressed size is over 6MB so should definitely be compressed). I'm finding that when using the gzip middleware, our content is marked as being encoded as gzip, but is not reduced in size at all. The content type of the result is `text/csv;charset=utf-8`, which appears to be covered by the `text/` check.

Also the documentation in the module is very unhelpful, it took reading the source code to find out that it was actually possible to set the values of `GzipSettings` - just export `GzipSettings{..}` instead of exporting the accessors separately.

Sadly this app isn't likely to be open source for quite a while, but I may be able to make a minimal example if necessary.

axman6:
Looking at the code, the `GzipCompress` constructor of `GzipFiles` isn't actually used anywhere, what is it even supposed to do?

snoyberg:
Documentation improvements would definitely be accepted as a PR.

As for behavior of `GzipCompress`, my guess is that through a refactoring at some point, that functionality was removed. I don't see any reason to avoid adding it back, though.

psibi:
One way of solving this would be by changing the default instance of `GzipSettings`. Another way would be to override the function `gzip`  in a way such that no matter what value is passed in `gzipFiles` for the record `GzipSettings` , we set it to `GzipCompress`.

Another way would be to change the signature to `gzip :: Bool -> Middleware`, like it was done previously. But that will break backward compatibility. I like the second solution. Any thoughts ?

axman6:
None of this seems to deal with the fact the data isn't actually being compressed. Apparently we're not alone though, I'm seeing the same behaviour from varnish or nginx (we're not sure which just yet) the first time we request data from a specific URL, but not from subsequent requests.

snoyberg:
> Looking at the code, the GzipCompress constructor of GzipFiles isn't actually used anywhere, what is it even supposed to do?

I reviewed the code again, and in fact `GzipCompress` _is_ used via the fall-through on line 92. But the code is very difficult to parse. I'm going to test a theory...

snoyberg:
I just confirmed that compressing files works as expected:

```haskell
import Network.Wai
import Network.Wai.Middleware.Gzip
import Network.Wai.Handler.Warp
import Network.HTTP.Types

main :: IO ()
main = run 3000 $ gzip def { gzipFiles = GzipCompress }
[...]

PR #546, opened 2016-04-26
Enable gzip compression when gzip is used

psibi:


snoyberg:
What's the motivation behind making this change? There's a significant downside to always compressing files: it breaks the ability to use the sendfile optimization.

psibi:
I think this is one of the way to fix #536. More details here: https://github.com/yesodweb/wai/issues/536#issuecomment-213684604

snoyberg:
My understanding is that, right now:

* The default behavior is to not compress files
* There is an option available to compress files
* But the option available doesn't actually compress files

This PR changes the default behavior without fixing the fact that files are not compressed.

But I just realized from rereading the code that some assumptions over there are incorrect, I'll comment there.

psibi:
Ok, my understanding was that the default behavior was to compress files. Probably we need to update the docs:

* http://www.yesodweb.com/book/web-application-interface#web-application-interface_middleware
* Hackage doc for `gzip` function

snoyberg:
I'm in favor of updating the docs for clarity. Let's wait and see though that #536 is in fact resolved.

psibi:
Closing the PR as #536 has been resolved.

Issue #535, opened 2016-04-20
What's the reason `updateAction` API doesn't allow modifiction?


nanonaren:
Is there any particular reason for `Control.AutoUpdate.updateAction` not providing an API to update the existing value? Seems rather a useful thing to have to me.

snoyberg:
Can you clarify what kind of API you're looking for, ideally with a type signature?

nanonaren:
So, I was looking for something like the following
```
updateAction :: Maybe a -> IO a
```
That way, when run run the first time it is called with `updateAction Nothing` and at later times as `updateAction (Just a)` where the current value is supplied as argument.

snoyberg:
I'm not actually seeing how such a function would work. Are you missing a parameter? Is this intended to be a field in a data type?

MaxGabriel:
@nanonaren wants something like this I think:

```haskell
foobar :: IO (IO Text)
foobar = mkAutoUpdate defaultUpdateSettings
  { updateAction = \maybePreviousValue -> foo -- Potentially use previous value to calculate the new value
  , updateFreq = 24 * 60 * 60 * 1000000 -- Update once per day
  }
```

@nanonaren It would probably help clarify things if you provided an example use case

nanonaren:
The example by @MaxGabriel is what I'm looking for. I have a structure in memory that gets updated periodically. Specifically, it maintains a vector of computed values based on recent data in the data. So it pulls in records after the last update time and then drops very old items. This is why I need to update the previous value rather than refresh it with a completely new one.

snoyberg:
OK, that makes sense. To provide for maximum performance and not break backwards compatibility, I'd imagine we could add a new field:

```haskell
updateActionModify :: Maybe (a -> IO a)
```

The idea would be:

* `updateAction` would always be used for the first action
* If provided, `updateActionModify` would be used for subsequent actions
* If not provided, current behavior continues

MaxGabriel:
@nanonaren Are you interested in implementing this?

nanonaren:
Sure, I'll do it.

nanonaren:
^ anyone want to check?

snoyberg:
LGTM. I didn't review this earlier since Github doesn't send notifications for commits pinging issues/PRs.

PR #547, opened 2016-05-04
Provide updateActionModify API in AutoUpdate

nanonaren:
Fixes yesodweb/wai#535

### DIFF
--- a/auto-update/Control/AutoUpdate.hs
+++ b/auto-update/Control/AutoUpdate.hs
@@ -6,12 +6,12 @@
 -- second, and write the current time to a shared 'IORef', than it is for each
 -- request to make its own call to 'getCurrentTime'.
 --
--- But for a low-volume server, whose request frequency is less than once per
--- second, that approach will result in /more/ calls to 'getCurrentTime' than
+-- But for a low-volume server, whose request frequency is less than once per 
+-- second, that approach will result in /more/ calls to 'getCurrentTime' than 
 -- necessary, and worse, kills idle GC.
 --
 -- This library solves that problem by allowing you to define actions which will
--- either be performed by a dedicated thread, or, in times of low volume, will
+-- either be performed by a dedicated thread, or, in times of low volume, will 
 -- be executed by the calling thread.
 --
 -- Example usage:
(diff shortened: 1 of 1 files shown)


snoyberg:
Travis builds failed due to (at least) missing imports.

nanonaren:
How did that happen? Why would it compile for me?

snoyberg:
You're probably using a newer version of GHC.

nanonaren:
Right. I am using 7.10.3 and the Travis build doesn't show the missing import for `<*>` in that version but does in 7.8.4 and 7.6.3. How do I fix this?

nanonaren:
I guess I'll just not use `<*>`. Probably the simplest thing to do.

snoyberg:
By adding the missing import statement.

nanonaren:
Adding the import statement raises the complaint
```
/home/narens/Documents/wai/auto-update/Control/AutoUpdate.hs:42:1: Warning:
    The import of ‘Control.Applicative’ is redundant
      except perhaps to import instances from ‘Control.Applicative’
    To import instances alone, use: import Control.Applicative()
```

Is that okay?

snoyberg:
It's OK. The usual solution is to use CPP to make the import conditional.

nanonaren:
Sorry, I could only find one example of `#ifdef GHC_7_4` in the code. Would this be fine
```
#ifndef GHC_7_10
import           Control.Applicative     ((<*>))
#endif
```

snoyberg:
I'm on my phone right now so I can't find an example myself, but there
should be lots of usages of MIN_VERSION scattered around

auto-update/Control/AutoUpdate.hs
```diff
@@ -90,6 +92,11 @@ data UpdateSettings a = UpdateSettings
     -- Default: does nothing.
     --
     -- @since 0.1.0
+    , updateActionModify   :: Maybe (a -> IO a)
+    -- ^ Optional action to be performed to get the current value
+    -- and updating it if necessary.
```
MaxGabriel: This sentence switches tense—I think you want "to get the current value and update it if necessary."

auto-update/Control/AutoUpdate.hs
```diff
@@ -90,6 +92,11 @@ data UpdateSettings a = UpdateSettings
     -- Default: does nothing.
     --
     -- @since 0.1.0
+    , updateActionModify   :: Maybe (a -> IO a)
```
MaxGabriel: This function needs to be exported in the export list above, btw

MaxGabriel:
Looks like there are some auto-update related compiler errors in the latest Travis results, btw: https://travis-ci.org/yesodweb/wai/jobs/127737868

nanonaren:
It looks like this change cannot be made without breaking the API. So, this error
```
Network/Wai/Handler/Warp/Date.hs:31:27:

    Couldn't match type ‘()’ with ‘ByteString’

    Expected type: Control.AutoUpdate.UpdateSettings GMTDate

      Actual type: Control.AutoUpdate.UpdateSettings ()

    In the expression: defaultUpdateSettings

    In the first argument of ‘mkAutoUpdate’, namely

      ‘defaultUpdateSettings

         {updateAction = formatHTTPDate <$> getCurrentHTTPDate}’
```

happens because in the use above `updateAction` is updated to the type `IO GMTDate` but `updateActionModify` remains unchanged. It will have to be explicitly set.

snoyberg:
That's unfortunate. I don't see a way of making this happen then.

snoyberg:
There's another approach that could be taken I think: add a new function `mkAutoUpdateWithModify` that takes the modification value as a separate argument.

MaxGabriel:
I think this is good to merge I think

snoyberg:
Thanks!

PR #551, opened 2016-05-17
Make warp benchmarks build with modern criterion

RyanGlScott:
I had to make some changes to get the `warp` benchmarks to build:

* I had to list some missing dependencies in the `benchmark` stanza
* I had to update the code a little for `criterion >= 1`

See also https://github.com/iu-parfunc/sc-haskell/issues/7


snoyberg:
I'm going to ignore the Travis error, it appears to be a cabal solver issue. Thanks!

PR #553, opened 2016-05-18
accept(2) error handling

reiddraper:
This PR contains three commits which deal with an issue we've been having with our production Warp services. Currently, Warp treats all exceptions that can be raised by `accept` as unrecoverable (other than its special handling of `ResourceExhausted`, which I've removed for reasons described in 3011cade3b7a6eb9dab124b1f6f6c4fc15efa121). However, some errors from the `accept` syscall _are_ recoverable, and do not imply the entire listening socket is closed or in a bad state. Specifically, accept(2) and accept4(2) will return ECONNABORTED if a connection request is aborted between entering the socket's listen queue, and actually being accepted. In this case, a subsequent call to `accept` will simply return the next socket on the queue.

Since this was treated as a fatal error before, when this would happen to us in production, the webserver would begin its graceful shutdown routine, but since we use long-lived connections, graceful shutdown would wait indefinitely for HTTP requests that will never finish (we leave an HTTP connection open and stream data). This manifested itself as the entire application hanging indefinitely. In aebd8d4081297a31726bcfeb947d525bfd816437, I've added an option to the Warp `Settings` for an optional graceful shutdown timeout, so that if there is a fatal error raised by `accept`, the webserver won't simply hang forever, waiting to shut down.

I've kept this in three commits, so that it's easier to review, and if you only want some of this functionality, it can be cherry-picked.

### DIFF
--- a/warp/Network/Wai/Handler/Warp.hs
+++ b/warp/Network/Wai/Handler/Warp.hs
@@ -247,6 +247,8 @@ getOnException :: Settings -> Maybe Request -> SomeException -> IO ()
 getOnException = settingsOnException
 
 -- | Get the graceful shutdown timeout
+--
+-- Since 3.2.8
 getGracefulShutdownTimeout :: Settings -> Maybe Int
 getGracefulShutdownTimeout = settingsGracefulShutdownTimeout
 
@@ -356,6 +358,8 @@ setLogger lgr y = y { settingsLogger = lgr }
 -- | Set the graceful shutdown timeout. A timeout of `Nothing' will
 -- wait indefinitely, and a number, if provided, will be treated as seconds
 -- to wait for requests to finish, before shutting down the server entirely.
+--
+-- Since 3.2.8
 setGracefulShutdownTimeout :: Maybe Int
                            -> Settings -> Settings
 setGracefulShutdownTimeout time y = y { settingsGracefulShutdownTimeout = time }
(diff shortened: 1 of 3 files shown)


reiddraper:
The failure is a timeout on the `BUILD=cabal GHCVER=7.6.3 CABALVER=1.16 HAPPYVER=1.19.5 ALEXVER=3.1.7` build. The other builds all pass.

warp/Network/Wai/Handler/Warp.hs
```diff
@@ -347,6 +353,13 @@ setLogger :: (Request -> H.Status -> Maybe Integer -> IO ())
           -> Settings -> Settings
 setLogger lgr y = y { settingsLogger = lgr }
 
+-- | Set the graceful shutdown timeout. A timeout of `Nothing' will
+-- wait indefinitely, and a number, if provided, will be treated as seconds
+-- to wait for requests to finish, before shutting down the server entirely.
```
MaxGabriel: It'd be good to add `-- @since x.x.x` or `-- Since x.x.x` like the other APIs have. I think the next version would be 3.2.8 (Current is 3.2.7, and this PR adds new APIs. Not sure if the `ResourceExhausted` thing would be considered a breaking change, though)

reiddraper: Added in 504fa37c5eb47645422ba263092ab54f4cee2086.

reiddraper:
I've force-pushed fce31039a0ccb6f99758755eafe772c1932d9882 to address the formatting issue in the documentation.

snoyberg:
Looks fine to me. @kazu-yamamoto do you want to review before merge?

kazu-yamamoto:
I support this changes.

snoyberg:
Thanks!

kazu-yamamoto:
Warp 3.2.7 has been released.

PR #554, opened 2016-05-20
Allow process-1.4

cocreature:
Builds fine with GHC 8 and process 1.4. It would be great if you could make a release after merging this.


snoyberg:
Thanks!

PR #556, opened 2016-05-22
Add dh-params setting to wrap-tls

tolysz:
Currently there is no way of specifying DHParams in `warp-tls` this patch adds configuration.
DH.Params are not exported from `tls` but they live in `cryptonite`.


snoyberg:
LGTM. @kazu-yamamoto any objections?

kazu-yamamoto:
No objections. 

If the low level libraries provide https://tools.ietf.org/html/draft-ietf-tls-negotiated-ff-dhe-10, it would be great.

tolysz:
Thanks

kazu-yamamoto:

Issue #555, opened 2016-05-20
Safe exception handling with Warp?


mkscrg:
Is it safe to catch all exceptions at the top level of a `wai`/`warp` app? [1]
- If safe and correct, let's document this on `warp`. Catching all exceptions is usually an antipattern, etc etc
- If not safe, is it reasonable to use something like `enclosed-exceptions`? I'd worry about that affecting performance, though I haven't tried it.

FWIW, this question came up because we'd like access to the `Request` when constructing a 500 response. `warp`'s current `onExceptionResponse` doesn't pass the `Request`.

[1] Like so:
```haskell
myApp :: Application
myApp req respond = respond . response500 req `handle` theRestOfMyApp

response500 :: Request -> SomeException -> Response
response500 req someEx = -- ...
```

snoyberg:
@kazu-yamamoto do you have thoughts on this? I'm not sure how the HTTP2 code affects things. AFAIK, it's safe to catch exceptions, though if you catch a `TimeoutThread` exception you should probably rethrow it.

kazu-yamamoto:
I believe all exception should be caught by the code to run Warp regardless of HTTP/1 and HTTP/2.
If it is not the case, it's a bug of Warp.

snoyberg:
Addressed by #557

PR #557, opened 2016-05-24
warp: add example of normal exception handling to Warp.hs haddocks

mkscrg:
As clarified in #555, it's safe to catch `SomeException` in an `Application` and respond with a 500 status. (Thanks @kazu-yamamoto @snoyberg) This is a rare exception to rule that catching `SomeException` is a [bad thing to do](https://www.schoolofhaskell.com/user/snoyberg/general-haskell/exceptions/catching-all-exceptions).

Here I've added some example code to `Warp.hs`. Is this the right place for it?


snoyberg:
Thanks!

PR #558, opened 2016-06-11
Gzip: add note about file compression

thumphries:
As far as I can tell, the `Data.Default` instance for `GzipSettings` does nothing. More than half of the Github results appear to be [using this accidentally](https://github.com/search?q=language%3Ahaskell+%22gzip+def%22&ref=simplesearch&type=Code&utf8=%E2%9C%93), and getting no compression as a result. I had to look at the source after testing my application to realise this instance was not useful.

Could also fix via improved documentation (or providing some `gzipCompress :: GzipSettings` that is easy to discover). However, I am struggling to think of cases where you'd import `Network.Wai.Middleware.Gzip` and actually want a no-op middleware, so I am suggesting a change to the instance.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/Gzip.hs
+++ b/wai-extra/Network/Wai/Middleware/Gzip.hs
@@ -50,12 +50,23 @@ data GzipSettings = GzipSettings
     , gzipCheckMime :: S.ByteString -> Bool
     }
 
-data GzipFiles = GzipIgnore | GzipCompress | GzipCacheFolder FilePath
+-- | Gzip behavior for files.
+data GzipFiles
+    = GzipIgnore -- ^ Do not compress file responses.
+    | GzipCompress -- ^ Compress files. Note that this may counteract
+                   -- zero-copy response optimizations on some
+                   -- platforms.
+    | GzipCacheFolder FilePath -- ^ Compress files, caching them in
+                               -- some directory.
     deriving (Show, Eq, Read)
 
+-- | Use default MIME settings; /do not/ compress files.
 instance Default GzipSettings where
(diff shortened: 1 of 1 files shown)


thumphries:
Apologies if I've misunderstood something :)

snoyberg:
The default middleware right now is not a no op. It simply doesn't compress files, which is a sensible default since file compression disabled the send file optimization

thumphries:
Ah, that makes sense. I was only checking the compression of static JS resources.

I'd like to add a comment to that effect, will update this PR.

thumphries:
Updated docs. Let me know if anything is wrong.

erikd:
:+1: LGTM!

wai-extra/Network/Wai/Middleware/Gzip.hs
```diff
@@ -50,12 +50,23 @@ data GzipSettings = GzipSettings
...
+                   -- platforms.
+    | GzipCacheFolder FilePath -- ^ Compress files, caching them in
+                               -- some directory.
     deriving (Show, Eq, Read)
 
+-- | Use default MIME settings; /do not/ compress files.
 instance Default GzipSettings where
     def = GzipSettings GzipIgnore defaultCheckMime
 
+-- | MIME types that will be compressed by default:
```
snoyberg: This code will also compress anything that starts with `text/`.

snoyberg:
Looks really good, I only had one minor comment.

thumphries:
> This code will also compress anything that starts with text/.

I truly am blind. Fixed

snoyberg:
Awesome, thank you!

PR #559, opened 2016-06-21
warp-tls: Handle recv failure on mkConn

utdemir:
That `safeRecv` is no different than a `recv` except on Windows.

Without this patch, a simple connect scan (nmap -sT <ip> -p <port>)
leaks an open socket, because the first recv fails with `recv:
connection vanished`.

You can easily reproduce the bug with sending that `nmap` command
to any warp-tls application.


utdemir:
IMO this is pretty important, since it makes `warp-tls` pretty much unusable in production. I think its worth a minor release on Hackage.

aycanirican:
@kazu-yamamoto ping

kazu-yamamoto:
LGTM. Merged. Thank you for your contribution.

kazu-yamamoto:

aycanirican:
@kazu-yamamoto thank you very much for the quick response

Issue #560, opened 2016-06-24
Memory exhaustion with tempFileBackend?


hce:
Hi, I haven't tested this, but by looking at the type signature, it seems that while files are indeed written to temporary files, what is there to stop me from declaring, say 1 GB of data not as a file but as a parameter? That would still be stored in memory in a ByteString. Since parameters are also sent inside the HTTP Body when doing a file transfer, it is not possible to limit the size of the parameters inside the reverse proxy.

Is there any way to limit the amount of memory the non-file parameters can take per request?

snoyberg:
No, and that's a good point. Interested in sending a PR to add a variant of `tempFileBackend` that allows the caller to provide an arbitrary byte limit to the size of each parameter, and/or the total bytes used by all parameters?

hce:
Yup! Lemme give it a try.

hce:
Looking at the code, it seems that the backend cannot control the number of files and doesn't even have access to the non-file parameters. I considered two options:

a) Add to the backend interface a way to intercept any parameter/file and give it a chance to reject it;
b) Add another variant of parseRequestBody that directly accepts more config options.

IMHO option b) is easier and also more efficient, because it will not be used at all for the normal parseRequestBody call.

The options I plan to add are:
 * Maximum key length (file/parameter name) (default: Just 32)
 * Maximum number of files (default: Just 10)
 * Maximum filesize per file (default: Nothing)
 * Maximum total filesize (default: Nothing)
 * Maximum number of parameters (default: Just 32)
 * Maximum size per parameter (default: Just 8192)
 * Maximum total parameter value size (default: Just 65536)

Sounds ok?

snoyberg:
Yes, sounds good.

snoyberg:
Fixed by your PR, thanks!

PR #562, opened 2016-07-02
Add various limits like filesize

hce:
Okay, here is my first attempt at fixing issue #560 .

Unfortunately, because the Network.HTTP.Types.parseSimpleQuery function does not allow to limit the number of parameters etc., I've only added a generic restriction on total non-parameter size.

I've also added a way to limit the number of files, the length of the filenames, the size of an individual file, and the total filesize. The number of header files in each mime entry can also be limited, but I'll still have to do another patch to limit individual line length.

Some refining is probably still necessary, but I'd be interested to hear if the general approach is ok.

Unfortunately, I had to duplicate a bit of code to keep the code compatible interface wise...

### DIFF
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -51,8 +51,6 @@ import Data.IORef
 import Network.HTTP.Types (hContentType)
 import Data.CaseInsensitive (mk)
 
-import Debug.Trace
-
 breakDiscard :: Word8 -> S.ByteString -> (S.ByteString, S.ByteString)
 breakDiscard w s =
     let (x, y) = S.break (== w) s
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -586,7 +586,7 @@ wrapTillBound bound src max' = do
             WTBDone _ -> return S.empty
             WTBWorking front -> do
                 bs <- readSource src
-                cur <- atomicModifyIORef sref $ \ cur ->
+                cur <- atomicModifyIORef' sref $ \ cur ->
                     let new = cur + fromIntegral (S.length bs) in (new, new)
                 case max' of
                     Just max'' | cur > max'' -> error "Maximum size exceeded"
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -45,6 +45,7 @@ import System.Directory (removeFile, getTemporaryDirectory)
 import System.IO (hClose, openBinaryTempFile)
 import Network.Wai
 import qualified Network.HTTP.Types as H
+import Control.Applicative ((<$>))
 import Control.Monad (when, unless)
 import Control.Monad.Trans.Resource (allocate, release, register, InternalState, runInternalState)
 import Data.IORef
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -219,8 +219,9 @@ clearMaxHeaderLineLength p = p { prboMaxHeaderLineLength=Nothing }
 -- maximum files: 10; filesize unlimited; maximum
 -- size for parameters: 64kbytes; maximum number of header
 -- lines: 32 bytes (applies only to headers of a mime/multipart message);
--- maximum header line length: 8190 bytes (applies only to headers of
--- a mime/multipart message)
+-- maximum header line length: Apache's default for that is 8190 bytes
+-- (http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestline)
+-- so we're using that here as well.
 defaultParseRequestBodyOptions :: ParseRequestBodyOptions
 defaultParseRequestBodyOptions = ParseRequestBodyOptions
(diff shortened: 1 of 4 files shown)


hce:
Sorry for the noise, but I was a bit too eager with the pull request earlier on. Should be ok now.

wai-extra/Network/Wai/Parse.hs
```diff
@@ -19,6 +19,8 @@ module Network.Wai.Parse
     , File
     , FileInfo (..)
     , parseContentType
+    , ParseRequestBodyOptions (..)
```
snoyberg: By exposing the data constructor here, we make it impossible to add new fields without breaking backwards compatibility. I'd rather go for a setter-based approach, which is generally more extensible.

wai-extra/Network/Wai/Parse.hs
```diff
@@ -115,6 +118,35 @@ tempFileBackEndOpts getTmpDir pattern internalState _ _ popper = do
...
+    , -- | The maximum filesize per file.
+      prboMaxFileSize           :: Maybe Int64
+    , -- | The maximum total filesize.
+      prboMaxFilesSize          :: Maybe Int64
+    , -- | The maximum size of the sum of all parameters values
+      prboMaxParmsValueSize     :: Int
+    , -- | The maximum header lines per entry
+      prboMaxHeaderLines        :: Int }
+
+instance Default ParseRequestBodyOptions where
```
snoyberg: Most of the community has been moving away from the `Default` typeclass, how about just having a `defaultParseRequestBodyOptions` value?

wai-extra/Network/Wai/Parse.hs
```diff
@@ -213,24 +288,56 @@ conduitRequestBody _ UrlEncoded rbody add = do
...
+                     -> RequestBodyType
+                     -> IO S.ByteString
+                     -> (Either Param (File y) -> IO ())
+                     -> IO ()
+conduitRequestBodyEx o _ UrlEncoded rbody add = do
+    -- NOTE: in general, url-encoded data will be in a single chunk.
+    -- Therefore, I'm optimizing for the usual case by sticking with
+    -- strict byte strings here.
+    size <- newIORef 0
+    let loop front = do
```
snoyberg: Is there any reason to use the mutable reference approach here? How about just including `cursize` as a parameter to `loop`?

snoyberg:
Overall this looks good, but I'm worried about the code duplication. Would it be possible to express the old non-throwing function in terms of the new throwing function?

hce:
Thank you very much for the detailed feedback. I'll address the issues you mentioned. I think it's safe to reduce the code duplication, but I will have to do more extensive testing to ensure existing functionality isn't broken by it.

wai-extra/Network/Wai/Parse.hs
```diff
@@ -172,76 +306,128 @@ parseContentType a = do
...
+conduitRequestBodyEx o backend (Multipart bound) rbody add =
+    parsePiecesEx o backend (S8.pack "--" `S.append` bound) rbody add
+
 
-takeLine :: Source -> IO (Maybe S.ByteString)
-takeLine src =
-    go id
+-- | Take one header or subheader line.
+-- It makes sense to limit the maximum line length.
+-- Apache's default is 8190 (http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestline)
```
snoyberg: Perhaps this comment should be moved above to the actual usage of `8190` in the code?

hce: Oops, definitely!

wai-extra/Network/Wai/Parse.hs
```diff
@@ -351,23 +567,32 @@ data WTB = WTBWorking (S.ByteString -> S.ByteString)
...
+                return (y, siz)
 
-    go ref = do
+    go ref sref = do
         state <- readIORef ref
         case state of
             WTBDone _ -> return S.empty
             WTBWorking front -> do
                 bs <- readSource src
+                cur <- atomicModifyIORef sref $ \ cur ->
```
snoyberg: Probably want to either use a bang pattern on `cur` or switch to `atomicModifyIORef'` to avoid a memory leak in the `Nothing` case of `max'`.

snoyberg:
This looks really good. I've added a few more minor comments, but this looks just about ready for merge. Would you be able to add a ChangeLog entry as well about the new functionality?

hce:
Of course, will do. Which version number should I use for the changelog entry? 3.0.16.0?

snoyberg:
Yes, perfect

hce:
Do you want me to create another branch for merging with the current master? Or is it easier for you to do the merge on your side?

snoyberg:
I just merged this onto a branch and opened up a new PR: https://github.com/yesodweb/wai/pull/563. The conflicts were trivial, but I've created the separate branch/PR so that Travis can confirm this all works before merging to master. Thanks for the work on this and the very responsive changes to comments :)

hce:
Thanks! :) I really enjoyed working on this.

snoyberg:
Awesome, glad to hear it! You're more than welcome to contribute again any time.

PR #564, opened 2016-07-05
Fix the way the header length is determined

hce:
Hi,
I'm really sorry, but during further testing of #563 I discovered a small but important bug that will cause the determined header length to be much bigger than the actual header length.

### DIFF
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -425,7 +425,7 @@ takeLine maxlen src =
     go front = do
         bs <- readSource src
         case maxlen of
-            Just maxlen' -> when (S.length front + S.length bs > maxlen') $
+            Just maxlen' -> when (S.length front > maxlen') $
                 error "Header line length exceeds allowed maximum."
             Nothing -> return ()
         if S.null bs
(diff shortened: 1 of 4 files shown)


wai-extra/wai-extra.cabal
```diff
@@ -1,5 +1,5 @@
 Name:                wai-extra
-Version:             3.0.16.0
+Version:             3.0.16.1
```
snoyberg: Good version bump 👍

wai-extra/Network/Wai/Parse.hs
```diff
@@ -425,7 +425,7 @@ takeLine maxlen src =
     go front = do
         bs <- readSource src
         case maxlen of
-            Just maxlen' -> when (S.length front + S.length bs > maxlen') $
+            Just maxlen' -> when (S.length front > maxlen') $
```
snoyberg: I should've caught that too, good fix

snoyberg:
Looks like a good fix, except now a test case is failing:

```
      test/Network/Wai/ParseSpec.hs:139: 
      1) Network.Wai.Parse.parseRequestBody exceeding header line size
           did not get expected exception: ErrorCall
```

hce:
Oh sorry, I didn't notice that the tests were failing when running them locally. I'm going to push two commits now that fix the tests, however: Please don't merge yet, as there seems to be another problem. When the header of the 2nd file is too long, it doesn't get recognized as a header but is treated as file contents. I'll look into this and get back to you.

hce:
Okay, the problem was with my tests, not the Parse.hs code. I've pushed another fix for the tests and the parser has also been tested in a real world scenario, so things should work correctly now.

PR #567, opened 2016-07-13
Catch `ConnectionClosed` exceptions when attempting to close a stream

TGOlson:
Ignore `ConnectionClosed` exceptions that may be thrown when attempting to close a stream using `WS.close`. Sometimes a stream may already be closed (normally due to some error) before we attempt this cleanup action. In those cases, we don't want the middleware to error.


snoyberg:
LGTM, thanks!

PR #570, opened 2016-07-23
Make ssIndices work with ssLookupFile and trailing slashes

soenkehahn:


wai-app-static/test/WaiAppEmbeddedTest.hs
```diff
@@ -32,7 +32,7 @@ embSpec = do
                              { requestHeaders = [("If-None-Match", "Etag 1")] }
             assertStatus 304 req
 
-        it "ssIndices works" $ do
+        it "ssIndices work" $ do
```
snoyberg: Why did you change this? I think the former grammar makes more sense (`ssIndices` is a singular noun as its own configuration value)

snoyberg:
Minor grammar comment, but not worth holding up the PR on it. Thanks!

wai-app-static/test/WaiAppEmbeddedTest.hs
soenkehahn: I thought it sounded better. But here's the reverting PR: https://github.com/yesodweb/wai/pull/574

soenkehahn:
Thanks for merging!

PR #572, opened 2016-07-26
warp: add HTTP2 CONNECT support

### LINKED ISSUE #561: feature request: HTTP2 CONNECT method support
HTTP2 spec defines `CONNECT method` in [section 8.3](https://http2.github.io/http2-spec/#CONNECT), and it's most commonly used as tunnel for HTTPS traffic over a proxy server.

A `WAI` application with CONNECT method requires connection takeover and could be implemented with `ResponseRaw`. Unfortunately `ResponseRaw` is not supported in current HTTP2 implementation of `Warp`. In order to run such WAI application, the implementation of HTTP2 `ResponseRaw` is necessary in Warp. I understand that it might be difficult to be implemented in a straightforward way due to HTTP's multiplexing nature. But if it's only about sending/receiving `DATA` frame in single stream, could we make a special case and and leave a dedicated thread for `ResponseRaw` streaming?

bjin:
As per section 8.3 of HTTP2 spec, ":scheme" and ":path" should be
omitted and ":authority" is mandatory for HTTP2 CONNECT request.

Treat CONNECT request as a special case for headers check. And also
for CONNECT request, use ":authority" as unparsed path, so that
CONNECT method has consistent behavior compare to HTTP 1.0.

Closes #561

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2/HPACK.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/HPACK.hs
@@ -83,13 +83,21 @@ hpackDecodeHeader hdrblk Context{..} = do
 {-# INLINE checkRequestHeader #-}
 checkRequestHeader :: ValueTable -> Bool
 checkRequestHeader reqvt
-  | getHeaderValue tokenStatus     reqvt /= Nothing     = False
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
I have merged this pull request. Thank you for your contribution.

PR #575, opened 2016-08-15
Feat/extend rewrite middleware

### LINKED ISSUE #571: Rewrite middleware contradicts guidance on rawPathInfo
Hello,

I’ve been using [Network.Wai.Middleware.Rewrite](https://github.com/yesodweb/wai/blob/master/wai-extra/Network/Wai/Middleware/Rewrite.hs) to convert a URL based on compulsory query parameters to one based on paths, in order to place a site I’m upgrading behind a reverse proxy. It’s a bit of a workaround, but the pattern may be useful more generally.

I’d like to add helper functions and some more examples to this module, but I noticed the two functions explicitly update [`rawPathInfo`](https://github.com/yesodweb/wai/blob/master/wai/Network/Wai/Internal.hs#L35) despite that function’s documentation advising against it.

So before I submit the PR, should my functions modify `rawPathInfo` and `rawQueryString` (as the functions in Network.Wai.Middleware.Rewrite), or not (as the documentation for the raw accessors states)?

---

Brief summary of the PR (to follow pending guidance): 
-   Factors out modification of request from generation of middleware (for use with `WPRModifiedRequest` of the `http-reverse-proxy` library)
-   Extend this to use query parameters in addition to path segments
-   Expose new functions to modify requests based on query parameters

This would add extra exports, but not break current exports from the module.

dbaynard:
Fixes #571 

-   The module preserves existing semantics of 'rewrite' and 'rewritePure' but adds warnings.
-   Four new functions are exported.
-   The documentation is expanded, including explanation of the changes.

I wanted to add tests, but couldn’t think of any that didn’t involve deriving `Arbitrary` instances for `Request`.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/Rewrite.hs
+++ b/wai-extra/Network/Wai/Middleware/Rewrite.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE CPP #-}
 
 module Network.Wai.Middleware.Rewrite
     ( rewrite
@@ -16,6 +17,11 @@ import Data.Functor.Identity (Identity(..))
 import qualified Data.Text.Encoding as TE
 import qualified Data.Text as T
 import Network.HTTP.Types as H
+
+-- GHC ≤ 7.10 does not export Applicative functions from the prelude.
+#if __GLASGOW_HASKELL__ <= 710
+import Control.Applicative
+#endif
 
 -- | A tuple of the path sections as '[Text]' and query parameters as
 -- 'H.QueryText'.
@@ -85,7 +91,7 @@ rewriteRequestPure convert req =
   in  runIdentity $ rewriteRequestRawM convertPure req
 
 -- | This helper function factors out the common behaviour of rewriting requests.
-rewriteRequestM :: Monad m
+rewriteRequestM :: (Applicative m, Monad m)
                 => (PathsAndQueries -> H.RequestHeaders -> m PathsAndQueries)
                 -> Request -> m Request
 rewriteRequestM convert req = do
@@ -98,7 +104,7 @@ rewriteRequestM convert req = do
 -- which the rewrite functions modify the `rawPathInfo` parameter. Note
 -- that this has not been extended to modify the `rawQueryInfo` as
 -- modifying either of these values has been deprecated.
-rewriteRequestRawM :: Monad m
+rewriteRequestRawM :: (Applicative m, Monad m)
                     => (PathsAndQueries -> H.RequestHeaders -> m PathsAndQueries)
                     -> Request -> m Request
 rewriteRequestRawM convert req = do
@@ -108,7 +114,7 @@ rewriteRequestRawM convert req = do
 
 -- | Produce a function that works on 'PathsandQueries' from one working
 -- only on paths.
-pathsOnly :: Monad m
+pathsOnly :: (Applicative m, Monad m)
           => ([Text] -> H.RequestHeaders -> m [Text])
           -> PathsAndQueries -> H.RequestHeaders -> m PathsAndQueries
 pathsOnly convert paths headers = (,[]) <$> convert (fst paths) headers
--- a/wai-extra/Network/Wai/Middleware/Rewrite.hs
+++ b/wai-extra/Network/Wai/Middleware/Rewrite.hs
@@ -1,21 +1,32 @@
+{-# LANGUAGE TupleSections #-}
+
 module Network.Wai.Middleware.Rewrite
-    ( rewrite, rewritePure
+    ( rewrite
+    , rewritePure
+    , rewriteWithQueries
+    , rewritePureWithQueries
+    , rewriteRequest
+    , rewriteRequestPure
     ) where
 
 import Network.Wai
 import Control.Monad.IO.Class (liftIO)
 import Data.Text (Text)
+import Data.Functor.Identity (Identity(..))
 import qualified Data.Text.Encoding as TE
 import qualified Data.Text as T
 import Network.HTTP.Types as H
 
+-- | A tuple of the path sections as '[Text]' and query parameters as
+-- 'H.QueryText'.
+type PathsAndQueries = ([Text], H.QueryText)
 
 -- | rewrite based on your own conversion rules
 rewrite :: ([Text] -> H.RequestHeaders -> IO [Text]) -> Middleware
 rewrite convert app req sendResponse = do
-  newPathInfo <- liftIO $ convert (pathInfo req) (requestHeaders req)
-  let rawPInfo = TE.encodeUtf8 $ T.intercalate "/" newPathInfo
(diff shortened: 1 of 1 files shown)


dbaynard:
The third commit was because I realised it was probably better to treat query parameters as data rather than text, and handle conversion on a per-use case.

dbaynard:
Think the build failure is because of the import of `Data.Functor.Identity` on GHC < 7.10. I’m not sure how best to handle this.

snoyberg:
This is the problematic bit:

```
Network/Wai/Middleware/Rewrite.hs:280:3: Not in scope: ‘pure’
Network/Wai/Middleware/Rewrite.hs:292:3: Not in scope: ‘pure’
Network/Wai/Middleware/Rewrite.hs:304:41: Not in scope: ‘<$>’
```

You'll need an import from Control.Applicative

dbaynard:
*slaps forehead* forgot about AMP. Thanks! Is the rest ok?

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+import Data.Functor.Identity (Identity(..))
 import qualified Data.Text.Encoding as TE
 import qualified Data.Text as T
 import Network.HTTP.Types as H
 
+-- $howto
+-- This module provides 'Middleware' to rewrite URL paths. It also provides
+-- functions that will convert a 'Request' to a modified 'Request'.
+-- Both operations require a function that takes url parameters and
+-- headers, and returns new url parameters. Parameters are pieces of url
```
MaxGabriel: URL should be capitalized

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+-- type of the new function is different the compiler will indicate where
+-- this change must be made.
+--
+-- The 'rewriteRequest' and 'rewriteRequestPure' functions use the new
+-- semantics, too.
 
--- | rewrite based on your own conversion rules
--- Example convert function:
+-- $takeover
+-- Let’s say we want to replace a website written in php with one written
```
MaxGabriel: PHP should be capitalized

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+-- > pathFromPhp (pieces, queries) _ = piecesConvert pieces queries
+-- >     where
+-- >         piecesConvert :: [Text] -> H.Query -> PathsAndQueries
+-- >         piecesConvert ["index.php"] qs@(join . lookup "page" -> Just page) =
+-- >             ( ["index", TE.decodeUtf8With TE.lenientDecode page]
+-- >             , delete ("page", pure page) qs
+-- >             )
+-- >         piecesConvert ps qs = (ps, qs)
+--
+-- On the other side, will use 'rewriteRequestPure' to rewrite outgoing
```
MaxGabriel: Should be we will use, I think

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+--
+-- /?bar=baz/ would look like
+--
+-- > [("bar", Just "baz")] :: QueryText
+--
+-- Together,
+--
+-- /\/foo?bar=baz/ would look like
+--
+-- > (["foo"],[("bar", Just "baz")]) :: PathsAndQueries
```
MaxGabriel: This section was pretty useful for understanding the sample code above. Maybe it should go first, I'm not sure

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+  newReq <- rewriteRequestM convert req
+  let rawPInfo = TE.encodeUtf8 . T.intercalate "/" . pathInfo $ newReq
+  pure newReq { rawPathInfo = rawPInfo }
+{-# WARNING rewriteRequestRawM [
+          "This modifies the 'rawPathInfo' field of a 'Request'."
+        , " This is not recommended behaviour; it is however how"
+        , " this function has worked in the past."
+        , " Use 'rewriteRequestM' instead"] #-}
+
+-- | Produce a function that works on 'PathsandQueries' from one working
```
MaxGabriel: "And" in `PathsAndQueries` should be capitalized

wai-extra/Network/Wai/Middleware/Rewrite.hs
```diff
@@ -1,32 +1,305 @@
...
+--   where
+--    piecesConvert [] = ["static", "html", "pages.html"]
+--    piecesConvert route@("pages":_) = "static":"html":route
+-- @
+--
+-- Instead, use this function, supplied to 'rewritePureWithQueries'.
+--
+-- @
+-- staticConvert :: PathsAndQueries -> H.RequestHeaders -> PathsAndQueries
+-- staticConvert pathsAndQueries _ = first (piecesConvert pathsAndQueries)
```
MaxGabriel: Is there a difference between this approach and `pathsOnly`?

MaxGabriel:
Really nice documentation, nice work @dbaynard

snoyberg:
There's a build failure on Travis from older GHCs missing `pure` and `<$>`, can you address those? For example, https://travis-ci.org/yesodweb/wai/jobs/152288546.

dbaynard:
Yes, I haven’t had the chance — didn’t realise it was an urgent issue — but I’ll do it now, and I’ll make Max’s changes too.

snoyberg:
Not urgent from my side, just a blocker for merging. I was triaging the emails I received over the past week of vacation so got to this now :)

wai-extra/Network/Wai/Middleware/Rewrite.hs
dbaynard: Yes. Also there’s a typo 👍 .

To clarify, previous users of the library will have code that looks like `staticConvert'`; this was the example given in the old formulation of the library. That code will still work. However, to migrate to using the new functions they just need to add the single function, `first :: ([Text] -> [Text]) -> PathsAndQueries -> PathsAndQueries` (I’ll add the specialised type to the docs). `pathsOnly` has to handle the `RequestHeaders` parameter too; first does not do this.

dbaynard:
Is it OK to have build warnings from the redundant `Control.Applicative` import for ghc ≥ 7.10, or would you prefer CPP statements?

snoyberg:
It's better to avoid the warnings.

dbaynard:
Woohoo, travis is happy! I can’t wait until there are stackage snapshots covering all the supported compiler versions…

Thank you all for the work you’ve done on this library and others. I figured the least I could do was document the functions I proposed to change.

snoyberg:
Thanks @dbaynard.

@MaxGabriel since you've done a more thorough doc review than I, do you think this is good to go?

MaxGabriel:
I haven't looked at the actual code, but yes the docs are good

PR #576, opened 2016-08-17
Import monoid to fix errors with latest optparse-applicative 0.13.*

charleso:
Apologies, not sure if this is the correct means of keep the compatibility with as much as possible.

`optparse-applicative` has just released a new version which no longer exports `<>`, which breaks `wai-app-static`.

https://github.com/pcapriotti/optparse-applicative/pull/216/files#diff-4191720afbea61cacc0c18ca54118978L52

Let me know if there's a more appropriate fix.


snoyberg:
Thanks!

PR #580, opened 2016-08-31
Gzip pre compressed

tolysz:
There are situations where we want to have some pre-compressed version of our files, maybe they are too big to compress on the fly or we want to use some very expensive gzip algorithm. Or we send a lot of them and want to keep CO2 low.

The solution could be an additional constructor for `gzip` it would allow serve compressed version if we have it. If we do not have it it will check what action it should do by using the value stored in its constructor.
I would even go as far as `GzipPreCompressed GzipIgnore` as a default value... but it adds extra file check so maybe a better way would be to configure it at the point of use.

It fixes the double `Last-Modified` headers on some static file responses as well.

(http://c1.kio.sx/twss/marcin) end result running with `keter`
 the file in question `http://c1.kio.sx/twss/all.json` would be too big to compress on the fly.

### DIFF
--- a/wai-extra/Network/Wai/Middleware/Gzip.hs
+++ b/wai-extra/Network/Wai/Middleware/Gzip.hs
@@ -58,6 +58,8 @@ data GzipFiles
                    -- platforms.
     | GzipCacheFolder FilePath -- ^ Compress files, caching them in
                                -- some directory.
+    | GzipPreCompressed GzipFiles -- ^ If we use compression use the file with ".gz"
+                                  -- appended to the name requested file if it exists
     deriving (Show, Eq, Read)
 
 -- | Use default MIME settings; /do not/ compress files.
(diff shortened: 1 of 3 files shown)


snoyberg:
Would it be possible to separate out the Warp changes to a separate PR? I'd like Kazu to be able to review them in isolation. (It seems like there's no dependency between that change and the rest of this.)

tolysz:
Ok, I will try. This commit will have only the Gzip changes. I will open one more with "last-modified"

wai-extra/Network/Wai/Middleware/Gzip.hs
```diff
@@ -92,13 +94,23 @@ gzip set app env sendResponse = app env $ \res ->
...
                 then
-                    case (res, gzipFiles set) of
-                        (ResponseFile s hs file Nothing, GzipCacheFolder cache) ->
-                            case lookup hContentType hs of
-                                Just m
-                                    | gzipCheckMime set m -> compressFile s hs file cache sendResponse
-                                _ -> sendResponse res
-                        _ -> compressE set res sendResponse
+                    let runAction x = case x of
+                            (_ , GzipPreCompressed (GzipPreCompressed _)) -> sendResponse res
```
snoyberg: This isn't what I would have expected. I'd have expected another clause after the `ResponseFile, GzipPreCompressed` clause that looks like:

```haskell
(res, GzipPreCompressed nextAction) -> runAction (res, nextAction)
```

tolysz: `GzipPreCompressed (GzipPreCompressed ..)` I tried to be "clever" and cut a possible loop. But it is not needed.  
Maybe it would be a bit more transparent to have an explicit list of things we try in order, they will be singletons for the current behavior? Right now I just have a way to try one extra check, without touching any other type.

wai-extra/Network/Wai/Middleware/Gzip.hs
```diff
@@ -92,13 +94,23 @@ gzip set app env sendResponse = app env $ \res ->
...
                 then
-                    case (res, gzipFiles set) of
-                        (ResponseFile s hs file Nothing, GzipCacheFolder cache) ->
-                            case lookup hContentType hs of
-                                Just m
-                                    | gzipCheckMime set m -> compressFile s hs file cache sendResponse
-                                _ -> sendResponse res
-                        _ -> compressE set res sendResponse
+                    let runAction x = case x of
+                            (ResponseFile s hs file Nothing, GzipPreCompressed nextAction) ->
```
snoyberg: Why would we ignore the `nextAction` entirely for responses besides `ResponseFile`? It seems like we would want to check that value for `ResponseBuilder` and `ResponseStream`

tolysz: It is not ignored :) we already are in `ResponseFile` constructor, this is just a specialization for places where we have precompressed files handy. `nextAction` can change the `Gzip`setting but not the response type.

snoyberg: What happens if `x ~ (ResponseBuilder ..., GzipPreCompressed GzipIgnore)`?

tolysz:
Travis failed :( I need to see why. And You are right the other constructors are missing.

wai-extra/Network/Wai/Middleware/Gzip.hs
tolysz: I updated the code above.
But `_ -> compressE set res sendResponse`  where `compressE` igonores `gzipFiles`

snoyberg: That's not the problem I was getting at. Presumably, in the case of `GzipPreCompressed GzipIgnore`, for non-`ResponseFile` responses the behavior should be identical to today's `GzipIgnore`. However, since the pattern match on `GzipIgnore` will fail, it will fall through to the default case and use `compressE`, which is not what is expected.

tolysz: I do not think the current (pre patch) code cares about `gzipFiles` setting for non `ResponseFile` cases.
The code checks only the request header, then if there is no `"gzip"` is shout-circuit to send response otherwise 
(as it already dissmised `GzipIgnore`) `compressE set res sendResponse` everything (but a `cacheFolder` case)

PR #581, opened 2016-09-03
Fixed doc typo

bitemyapp:


bitemyapp:
Nuts, this is actually wrong, `responseLBS` is a `Network.Wai.Response` and not an `Application` when all params are applied. Not sure what `respondLBS` was supposed to be, but I can't find it anywhere in `wai`.

bitemyapp:
This should do it now.

If I might express a preference, I think an `examples` dir of small modules that get compiled in CI is better than sticking code in Haddocks.

wai-websockets/Network/Wai/Handler/WebSockets.hs
```diff
@@ -49,7 +49,7 @@ isWebSocketsReq req =
 --         'WS.sendTextData' conn ("Hello, client!" :: 'Data.Text.Text')
 --
 --     backupApp :: 'Wai.Application'
---     backupApp = 'Wai.respondLBS' 'Network.HTTP.Types.status400' [] "Not a WebSocket request"
+--     backupApp _ response = response $ 'Wai.responseLBS' 'Network.HTTP.Types.status400' [] "Not a WebSocket request"
```
snoyberg: Typical nomenclature here would be `respond` instead of `response`, you OK with that change?

bitemyapp: @snoyberg quite alright and makes more sense given it's a verb.

PR #582, opened 2016-09-24
ForceSSL: preserve port number when redirecting to https.

bollmann:
Previously, the port number was dropped from the request's host header when redirecting a request to https.

Now, the port from the request's host header is kept as is.


snoyberg:
Thanks!

PR #585, opened 2016-10-17
Add basicAuth'

phadej:


phadej:
Travis fails because of `setup: The program 'hsc2hs' is required but it could not be found`, seems unrelated

wai-extra/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 3.0.19
+
+* Add a new function basicAuth', which passes request to the CheckCreds variant.
```
snoyberg: I'm not sure what "variant" means in this case, can you clarify? Perhaps this should be "argument"?

wai-extra/Network/Wai/Middleware/HttpAuth.hs
```diff
@@ -40,7 +40,15 @@ type CheckCreds = ByteString
 basicAuth :: CheckCreds
           -> AuthSettings
           -> Middleware
-basicAuth checkCreds AuthSettings {..} app req sendResponse = do
+basicAuth checkCreds = basicAuth' (\_ -> checkCreds)
+
+-- | Like 'basicAuth', but also passes a request to the authentication function.
+--
+-- Since 3.0.19
```
snoyberg: Can you use `@since` syntax instead? I know that the library still uses `Since` in many places, but I'm trying to update it bit by bit.

snoyberg:
I just pushed an updated Travis config to master, hopefully that will fix the hsc2hs issue. I'm not sure what's been causing it to pop up recently.

phadej:
Changed all annotations to use `@since`. Also other comment addressed.

wai-extra/Network/Wai/Middleware/HttpAuth.hs
```diff
@@ -36,11 +36,19 @@ type CheckCreds = ByteString
...
-basicAuth checkCreds AuthSettings {..} app req sendResponse = do
+basicAuth checkCreds = basicAuth' (\_ -> checkCreds)
+
+-- | Like 'basicAuth', but also passes a request to the authentication function.
+--
+-- @since 3.0.19
+basicAuth' :: (Request -> CheckCreds)
+           -> AuthSettings
+           -> Middleware
+basicAuth' checkCreds AuthSettings {..} app req sendResponse = do
```
snoyberg: Sorry, but just noticed: this function isn't exported from this module.

phadej: Ah true, it's used by `basicAuth` so didn't trigger a warning. Exported now.

snoyberg:
Thanks!

PR #586, opened 2016-10-28
Fix spaceleak in sendResponse

edsko:
Consider the following simple WAI application that uses `wai-conduit` to generate a streaming body:

``` haskell
{-# OPTIONS_GHC -fno-full-laziness #-}
module Main (main) where

import Data.ByteString.Builder (Builder, int32BE, word8)
import Data.Conduit
import Data.Int (Int32)
import Network.HTTP.Types
import Network.Wai.Handler.Warp (run)
import Network.Wai.Conduit

main :: IO ()
main = run 3000 $ \_req k -> k $ responseSource ok200 [] (testSource 1000000)

testSource :: Int32 -> Source IO (Flush Builder)
testSource 0  = return ()
testSource !i = do
    yield (Chunk $ int32BE i)
    yield (Chunk $ word8 10)
    yield Flush
    testSource (i - 1)
```

We'd expect this to run in constant space, but sadly it suffers from a space leak. In fact, it's another dreaded of-type-`Pipe` space leak :-( It seems I'm [spending a significant chunk of my time tracking those down lately](http://www.well-typed.com/blog/2016/09/sharing-conduit/)... I wish we had a better way to deal with these.

Anyway, the problem is not with the conduit per se; connecting it directly to a sink that writes all values to standard output runs in constant space. Nor is the problem in `wai-conduit`; manually executing the `StreamingBody` generated by `responseSource` works just fine. In the end I traced this down to a bug in `sendResponse` in `warp` itself; hopefully the commit is clear enough: we're hanging on to `response` longer than we need, which is causing the space leak. 

Though I realize that that is somewhat hand-wavy, once I realized that the problem was here I didn't investigate in more detail the exact nature of this spaceleak. Note however that for this test case, disabling full laziness is necessary even with the bugfix applied (without the bugfix there is a spaceleak with full laziness enabled or not).

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -136,6 +136,8 @@ sendResponse :: Settings
 sendResponse settings conn ii req reqidxhdr src response = do
     hs <- addServerAndDate hs0
     if hasBody s then do
+        -- Make sure we don't hang on to 'response' (avoid space leak)
+        ret' <- E.evaluate ret
         -- The response to HEAD does not have body.
         -- But to handle the conditional requests defined RFC 7232 and
         -- to generate appropriate content-length, content-range,
@@ -147,7 +149,7 @@ sendResponse settings conn ii req reqidxhdr src response = do
             Nothing         -> return ()
             Just realStatus -> logger req realStatus mlen
         T.tickle th
-        return ret
+        return ret'
       else do
         _ <- sendRsp conn ii ver s hs RspNoBody
         logger req s Nothing


snoyberg:
Ouch, sorry you keep hitting these 😢 

@kazu-yamamoto I see no downside to this change, how about you?

edsko:
Haha :) Not your fault, obviously. Just makes me wonder if there isn't something more fundamental that we can do about these problems. I'd much prefer neither library authors nor library users would need to think quite so hard about this kind of memory behaviour of libraries like conduit. Oh well. Wishful thinking right now. Maybe on day :)

edsko:
By the way, I don't think there can be a downside, because this just forces the evaluation of `response :: `Response` to whnf, but the next line calls `sendRsp` which is strict in its `Rsp` argument, and that `Rsp` argument is constructed by looking at the top-level of the `Response` argument; so that evaluation of `response` to whnf will happen immediately anyway.

edsko:
@kosmikus points out that my comment about full laziness is somewhat unclear, so let me clarify: with the bug fix enabled, the _test case_ still needs to be compiled with full laziness disabled (otherwise that `testSource 1000000` expression gets floated to the top, becomes a CAF, and we get an -- uninteresting -- space leak for that reason). Full laziness does not need to be disabled in `warp` itself, as far as I know.

kazu-yamamoto:
I'm not a big fun of lazy evaluation. So, I welcome strictness.
What about using BangPatterns instead of `evaluate`?

```
     !ret = ...
```

edsko:
I guess that would work as well; I used `evaluate` because it makes it a lot clearer when exactly the evaluation happens. If you have a big function definition with a lot of `where`-bound variables and a bang on one of them, it's harder to see. (Also, I think having the bang _does_ in fact change the semantics of the function, as it now forces the evaluation to happen in both cases of the conditional. Don't think that actually matters but it might.) Another alternative might be to move that definition of `ret` up and then use

``` haskell
let !ret = ...
...
```

that would also work, and still be clear. Whatever the preferred style is :)

kazu-yamamoto:
I have merged your pull request and changed the style to use BangPatterns at `where`.
I like this style because of the case where other valuables need BangPatterns.

edsko:
Ok, sure. Thanks!

kazu-yamamoto:
A new version has been released. Thank you for your contribution!

PR #589, opened 2016-11-18
bytestring-builder is not used

nponeccop:


snoyberg:
It is used when compiling with older versions of bytestring, unless I'm mistaken.

nponeccop:
This information is for future reference, if you decide to drop support in the future:

`wai` has `bytestring >= 0.10` constraint, and `builder` is required before 0.10.4, so the only versions affected are 0.10.0.0, 0.10.0.1, 0.10.0.2, 0.10.2.0.

In terms of GHC versions it is 7.6.1 and 7.6.2

PR #594, opened 2016-11-25
URL now seems to be squatted by spam redirector

supermario:
Noticed the URL reference for `WaiAppStatic.Listing.defaultListing` was redirecting to some rather dodgy sites.

Updated URL to point to the file on Happstack GH repo instead, pinned to most recent SHA change to the file (happens to coincide with `blaze-html 0.5` release) prior to this comment being added here on `2012-05-22`.


snoyberg:
Thanks!

PR #596, opened 2016-12-02
Add docs example to setInstallShutdownHandler

4e6:
- add usage exmple to `setInstallShutdownHandler` docs

I've spent some time recently trying to figure out how to use `setInstallShutdownHandler`. I've also found that it is a quite frequent question on the internet. 

The confusing part is that you don't call lambda argument (which closes socket) directly, and should use signal handler instead. On the other hand, this is exactly what the docs said. After some considerations, I thought that wording in the docs is fine, but it lacks an example.


snoyberg:
Thanks!

PR #598, opened 2016-12-07
Update slightly confusing description.

alexanderkjeldaas:
As TLS is just a name for SSL to avoid a lawsuit from Netscape, and has no technical meaning in itself, I think the description should be qualified a bit.

At least this confused me, and I can see in the issues that others have been similarly confused.


snoyberg:
Any objection @kazu-yamamoto?

kazu-yamamoto:
I believe that this issue was already fixed by me. Please read the current version of `warp-tls.cabal`.

snoyberg:
Awesome 👍

PR #600, opened 2016-12-09
build a docker image with the warp executable

ketzacoatl:
`stack image container` will now build a docker image suitable for
static file serving out of the box and with no additional futzing.

Two questions I have:
* could travis run `stack image container` to build a docker image for commits on `master`?
* does `yesodweb` have a docker repo this could use?

...or should this not be integrated with CI and just be available for users to create and host on their own (which is totally reasonable too)?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -23,3 +23,9 @@ nix:
   packages:
   - fcgi
   - zlib
+image:
+  containers:
+    - base: "fpco/pid1"
+      name: "yesodweb/warp"
+      executables:
+        - warp


snoyberg:
I like the idea of a yesodweb org on Docker Hub, so I've created it. Automating the pushing of images like this is painful; we could have Travis build an executable, upload to S3, and then have an automated build that pulls that executable into an image, but it will still require manual running because of timing issues. IMO, it's easier to just have a manual `stack image container && docker push yesodweb/warp` process.

snoyberg:
OK, `yesodweb/warp` is now pushed to Docker Hub.

ketzacoatl:
OK, with the exception of travis not doing much to help, that sounds great!

PR #605, opened 2017-01-22
Better async exception behavior for timeouts #602

### LINKED ISSUE #602: Expected behaviour when catching TimeoutThread?
In an application I am currently debugging a response handler catches all exceptions via 

```haskell
`catch` \(e :: SomeException) ->
```

and then logs the expection and continues to do it's work. While I know that this is bad coding and should be fixed right there, I was wondering what the expected behaviour of warp for this kind of situation would be?

snoyberg:
This fixes #602.

Pinging @kazu-yamamoto and @agrafix

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2/Worker.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/Worker.hs
@@ -236,7 +236,8 @@ worker :: Context -> S.Settings -> Application -> Responder -> T.Manager -> IO (
 worker ctx@Context{inputQ,controlQ} set app responder tm = do
     sinfo <- newStreamInfo
     tcont <- newThreadContinue
-    E.bracket (T.registerKillThread tm) T.cancel $ go sinfo tcont
+    let timeoutAction = return () -- cannot close the shared connection
+    E.bracket (T.registerKillThread tm timeoutAction) T.cancel $ go sinfo tcont
   where
     go sinfo tcont th = do
         setThreadContinue tcont True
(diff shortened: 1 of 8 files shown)


warp/Network/Wai/Handler/Warp/HTTP2/Worker.hs
```diff
@@ -236,7 +236,8 @@ worker :: Context -> S.Settings -> Application -> Responder -> T.Manager -> IO (
 worker ctx@Context{inputQ,controlQ} set app responder tm = do
     sinfo <- newStreamInfo
     tcont <- newThreadContinue
-    E.bracket (T.registerKillThread tm) T.cancel $ go sinfo tcont
+    let timeoutAction = return () -- cannot close the shared connection
```
snoyberg: @kazu-yamamoto I'm not sure if this is correct, or if there is in fact some way to close the connection when the timeout is triggered.

kazu-yamamoto:
I will try to understand this on Friday since I will join the QUIC interim meeting in Tokyo from Tuesday to Thursday.

Before that, it would be nice if we can fix the CI failures above.

snoyberg:
Hmm, having trouble reproducing locally, investigating.

snoyberg:
OK, it reproduces reliably on Linux, not at all on OS X. It seems that registering the socket close causes the problem, but I can't see why. Investigation continues.

snoyberg:
I've narrowed this down to a double-free error in `connClose` with the call to `freeBuffer`. However, looking at some tracing output I've added I have no idea how it ever gets called twice. The only other thing I can think of is that the buffer is somehow being freed in a different way elsewhere, such as via GC. If that's the case, I just uncovered a bug in the library somewhere. But this is just conjecture, more likely is that I'm doing something dumb and not realizing it.

snoyberg:
Alright, thanks to @lehins finding the bug, this PR is now passing Travis, and (thankfully) no longer segfaulting :)

kazu-yamamoto:
LGTM. Merged.
I will release warp and warp-tls soon.

PR #606, opened 2017-01-27
Update type of Application in README.md

iporsut:
I am trying to learn Wai. But see type of Application in document be not like in GHCI.


snoyberg:
Thanks!

PR #611, opened 2017-02-16
catch exceptions "escaping" from mkConn (connection maker) so they are properly handled by a user provided handler 

erickg:
Hi,

while using the warp-tls module in a project I am working on, I noticed that exceptions thrown "early" during "connection making" as you seem to call it, are not being passed to the user provided handler (or the default handler for that matter). The worker thread is simply terminated and the exception gets displayed on stderr by the default Haskell exception handler (and in most server applications, nobody is monitoring stderr so in end effect this becomes pretty much a silent failure).

This small change simply makes sure that if an exception makes it all the way to the top of the forked thread, it will still get passed to the exception handling function provided in settings.

Cheers!

Erick


erikd:
+1000 for the general principle. Don't know enough about he actual code to comment o that though.

snoyberg:
LGTM. @kazu-yamamoto want to sanity-check this as well before merging?

Thanks @erickg!

PR #612, opened 2017-02-18
Tidy up docstring of `MethodOverride` middleware

prophile:
This tidies the grammar and brings the structure of the first sentence in line with that used elsewhere.

Not a _dramatic_ improvement, more as a mild decrease in surprise factor when browsing the documentation.


snoyberg:
Thanks!

PR #616, opened 2017-02-23
Move exception handling to top of thread (fixes #613)

### LINKED ISSUE #613: Regression in warp 3.2.10: Timeout reaper is triggered if socket is kept open
When building the program below against warp *3.2.9*, everything works as expected. However, using *3.2.10* or *3.2.11*  `Thread killed by Warp's timeout reaper` is printed to stderr when I open [http://localhost:8080](http://localhost:8080) in a browser (I tried firefox and chrome). This error does not occur using curl (even when copying the curl command provided by the browser) and it also doesn’t appear if I click fast enough on `Close idle sockets` on `chrome://net-internals/#sockets`. This strongly suggests that the problem is caused by browsers keeping sockets open for potential reuse. I don’t have the time right atm to bisect and find the exact commit that triggered this bug.

```
module Main where

import Network.HTTP.Types.Status
import Network.Wai
import qualified Network.Wai.Handler.Warp as Warp

main :: IO ()
main = do
  Warp.runSettings (Warp.setTimeout 1 $ Warp.setPort 8080 Warp.defaultSettings) $ \req respond ->
    respond $ responseLBS status200 [] "Hello World"
```

snoyberg:
This is a long-standing bug, but hardly manifested until
3eb586636b7f0a9a8e4b4ad456c16efe23847c9e landed. With that change, the
bracket function calls outside of the unmask were likely to receive an
async timeout exception. Without a wrapping call to settingsOnException,
these would be printed to the console, which isn't desired.

Moving the handle call to the top of the thread ensures that any
exception which occurs inside the thread has the handler applied.

### DIFF
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -1,3 +1,8 @@
+## 3.2.11.1
+
+* Move exception handling to top of thread (fixes
+  [#613](https://github.com/yesodweb/wai/issues/613))
+
 ## 3.2.11
 
 * Fixing 10 HTTP2 bugs pointed out by h2spec v2.
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -263,51 +263,53 @@ fork :: Settings
      -> Counter
      -> InternalInfo0
      -> IO ()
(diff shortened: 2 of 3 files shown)


snoyberg:
Looks good @kazu-yamamoto!

PR #621, opened 2017-05-24
Relax HTTP2 headers check.

lucasdicioccio:
HTTP/2-draft17 (e.g., used in gRPC) does not mandate an `:authority` pseudo header.
In fact, this pseudo header MUST be omitted in some HTTP/1.1 translation
cases.

More details at https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-8.1.2.3 .


lucasdicioccio:
Note that the only Travis failing build was failing with:
```The job exceeded the maximum time limit for jobs, and has been terminated.```

psibi:
I have restarted that build.

kazu-yamamoto:

Issue #439, opened 2015-09-30
Collaborating with http-types


charleso:
[Both](https://github.com/yesodweb/wai/search?utf8=%E2%9C%93&q=hContentRange) [warp/Network/Wai/Handler/Warp/Types.hs](https://github.com/yesodweb/wai/blob/c5001a0fb8100a425e68b09a38c32b83bd69384f/warp/Network/Wai/Handler/Warp/Types.hs#L36) and [wai/Network/Wai/Internal.hs](https://github.com/yesodweb/wai/blob/ac96dc569708a5a3c7df9906ddd2a9f58ff30a2d/wai/Network/Wai/Internal.hs#L167) define  `hContentRange`. This leads to the following:

```
Network/Wai/Handler/Warp/Response.hs:107:43:
    Ambiguous occurrence ‘parseByteRanges’
    It could refer to either ‘Network.Wai.Handler.Warp.RequestHeader.parseByteRanges’,
                             imported from ‘Network.Wai.Handler.Warp.RequestHeader’ at Network/Wai/Handler/Warp/Response.hs:54:48-62
                             (and originally defined
                                at Network/Wai/Handler/Warp/RequestHeader.hs:138:1-15)
                          or ‘Network.Wai.Internal.parseByteRanges’,
                             imported from ‘Network.Wai.Internal’ at Network/Wai/Handler/Warp/Response.hs:58:1-27

Network/Wai/Handler/Warp/Response.hs:394:25:
    Ambiguous occurrence ‘hAcceptRanges’
    It could refer to either ‘Network.Wai.Handler.Warp.Types.hAcceptRanges’,
                             imported from ‘Network.Wai.Handler.Warp.Types’ at Network/Wai/Handler/Warp/Response.hs:57:1-37
                             (and originally defined
                                at Network/Wai/Handler/Warp/Types.hs:40:1-13)
                          or ‘Network.Wai.Internal.hAcceptRanges’,
                             imported from ‘Network.Wai.Internal’ at Network/Wai/Handler/Warp/Response.hs:58:1-27

Network/Wai/Handler/Warp/Response.hs:406:39:
    Ambiguous occurrence ‘hContentRange’
    It could refer to either ‘Network.Wai.Handler.Warp.Types.hContentRange’,
                             imported from ‘Network.Wai.Handler.Warp.Types’ at Network/Wai/Handler/Warp/Response.hs:57:1-37
                             (and originally defined
[...]

PR #623, opened 2017-05-30
Using http-types 0.9.1 only

kazu-yamamoto:
10 months have passed since http-types v0.9.1 is released.
It's time to remove the support old http-types to clean up the code.
See https://github.com/yesodweb/wai/issues/439.


snoyberg:
Looks good to me, merge when ready.

kazu-yamamoto:
This PR breaks tls-3 and others.
Should we support them?

Related question: why don't we support 8.0.2 instead of 8.0.1?

snoyberg:
I've just created #625 to try and fix these failures.

Issue #593, opened 2016-11-24
Thread killed by timeout reaper when sending large HTTP/2 response


ip1981:
With Warp 3.2.8 I'm getting error `Thread killed by Warp's timeout reaper` when sending large (GiB) bodies from backend servers. This looks similar to https://github.com/yesodweb/wai/issues/351, but happens only on HTTP/2 (HTTP/1.1 to the backend).

[The docs say](https://github.com/yesodweb/wai/blob/7a961288b997c2c485957743763d10d7644a61d6/warp/Network/Wai/Handler/Warp/Internal.hs)
> Every time data is successfully sent to the client, the timeout is tickled.

The data is definitely being sent to the client. Also [this commit](https://github.com/yesodweb/wai/commit/8655714bbb1bde6e24736ca712bc30134ad79be4) suggests that timeout should be disabled _when_ request body is consumed, that I believe is the case.

[Source code](https://github.com/ip1981/sproxy2/blob/master/src/Sproxy/Application.hs):

```haskell
forward :: BE.Manager -> W.Application
forward mgr req resp = do
  let beReq = BE.defaultRequest
        { BE.method = W.requestMethod req
        , BE.path = W.rawPathInfo req
        , BE.queryString = W.rawQueryString req
        , BE.requestHeaders = modifyRequestHeaders $ W.requestHeaders req
        , BE.redirectCount = 0
        , BE.decompress = const False
        , BE.requestBody = case W.requestBodyLength req of
            W.ChunkedBody   -> requestBodySourceChunkedIO (sourceRequestBody req)
            W.KnownLength l -> requestBodySourceIO (fromIntegral l) (sourceRequestBody req)
        }
      msg = unpack (BE.method beReq <> " " <> BE.path beReq <> BE.queryString beReq)
  Log.debug $ "BACKEND <<< " ++ msg ++ " " ++ show (BE.requestHeaders beReq)
  BE.withResponse beReq mgr $ \res -> do
        let status = BE.responseStatus res
            headers = BE.responseHeaders res
            body = mapOutput (Chunk . fromByteString) . bodyReaderSource $ BE.responseBody res
            logging = if statusCode status `elem` [ 400, 500 ] then
                      Log.warn else Log.debug
        logging $ "BACKEND >>> " ++ show (statusCode status) ++ " on " ++ msg ++ " " ++ show headers ++ "\n"
        resp $ responseSource status (modifyResponseHeaders headers) body
```

kazu-yamamoto:
[...]

PR #624, opened 2017-05-30
Wrap recvN in HTTP2 to tickle the timer.

lucasdicioccio:
This patch addresses a bug preventing long HTTP2 connections.  See #593
for an initial report, as well as
https://github.com/yesodweb/wai/issues/593#issuecomment-304382746 for a
reproducible example -- which this patch addresses (the actual root
cause for #593 might differ but seem closely related).


kazu-yamamoto:
I have merged this PR with rebasing and modifications.
Thank you for your contribution.

kazu-yamamoto:

PR #629, opened 2017-06-20
Allow people to manage session state between calls to runSession.

fisx:
We want to keep state between calls to runSession in our tests, but this is not possible without this patch because ClientState is not exported.

In our opinion this is fine for merging, but we are happy to follow any instructions that make it land.

thanks!

### DIFF
--- a/wai-extra/Network/Wai/Test.hs
+++ b/wai-extra/Network/Wai/Test.hs
@@ -5,7 +5,7 @@ module Network.Wai.Test
     ( -- * Session
       Session
     , runSession
-    , runSessionWith, ClientState(..), initState
+    , runSessionWith, initState
       -- * Client Cookies
     , ClientCookies
     , getClientCookies
@@ -40,6 +40,7 @@ import Data.Monoid (mempty, mappend)
 
 import Network.Wai
 import Network.Wai.Internal (ResponseReceived (ResponseReceived))
+import Network.Wai.Test.Internal
 import Control.Monad.IO.Class (liftIO)
 import Control.Monad.Trans.Class (lift)
 import qualified Control.Monad.Trans.State as ST
--- a/wai-extra/Network/Wai/Test.hs
+++ b/wai-extra/Network/Wai/Test.hs
@@ -105,12 +105,21 @@ deleteClientCookie cookieName =
   modifyClientCookies
     (Map.delete cookieName)
 
+-- |
+--
+-- Since 3.0.20.0
 initState :: ClientState
 initState = ClientState Map.empty
 
 runSession :: Session a -> Application -> IO a
 runSession session app = ST.evalStateT (runReaderT session app) initState
 
+-- | Like 'runSession', but if allows you to hand in cookies and get
+-- the updated cookies back.  One use case for this is writing tests
+-- that address the application under test alternatingly through rest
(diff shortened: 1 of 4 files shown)


wai-extra/Network/Wai/Test.hs
```diff
@@ -5,6 +5,7 @@ module Network.Wai.Test
     ( -- * Session
       Session
     , runSession
+    , runSessionWith, ClientState(..), initState
```
snoyberg: I'd rather not expose the constructor and fields. Would it work to just export `ClientState`?

snoyberg: Please add `@since` comments with a new minor version bump for all newly exposed identifiers, as well as a modified .cabal file and updated ChangeLog.

wai-extra/Network/Wai/Test.hs
```diff
@@ -110,6 +111,9 @@ initState = ClientState Map.empty
 runSession :: Session a -> Application -> IO a
 runSession session app = ST.evalStateT (runReaderT session app) initState
 
+runSessionWith :: ClientState -> Session a -> Application -> IO (a, ClientState)
```
snoyberg: Can you add a Haddock comment here explaining why you'd use this vs `runSession`?

wai-extra/Network/Wai/Test.hs
fisx: good idea, will do.  is 3.0.20.0 the right next version?

snoyberg: Correct. FYI, you may be interested in: http://www.snoyman.com/blog/2017/06/how-to-send-me-a-pull-request

fisx: About `ClientState`: what i want to do is get the state out of the result of `runSessionWith`, parse it, and feed it to the backend handlers when talking to the database directly.  (I know that this style of writing test has its opponents, but I like them because it lets you check invariants in the backend without having to expose them via rest api.)

I think what would also work for me is just exporting the type, the record field selector, and `initState`, then I could read and write the cookies using `s { clientCookies = ... }` (that works without having the constructor, right?), but I am not sure that's any better for you.

Instead, I propose c3221d7a60968ef18e4d13ad03eb8d620a4980b6.

fisx: thanks for the link.  sorry you had to point it out, again...  :-)

fisx:
I think I found a better solution, could you take another look?  Please select whether you would like me to:

- [ ] keep (and fix) c3221d7a60968ef18e4d13ad03eb8d620a4980b6, and revert ed575fb2f0aa9fba37c32e93bebff2566e9a2dd4
- [ ] squash away detour commits
- [ ] other: ...

Thanks!

snoyberg:
I _think_ the answer is checkbox (1). To be clear: I'd rather the constructors be exposed from a `.Internal` module. I don't care about squashing commits, I can always do that while merging if necessary.

fisx:
Ok, done.  I also moved `runSessionWith` to Internal, as it does not make sense without access to the `ClientState` type.  This means I had to pull in a few more definitions.

Anything else?

snoyberg:
Looks good, thanks!

PR #630, opened 2017-06-29
TLS session manager

kazu-yamamoto:
I would like to include an in-memory TLS session manager to `wai` rather than to `tls` because it requires `auto-update` and it is used in `warp-tls`.

I'm implementing this, rather than the mechanism of stateless session tickets, based on [this analysis](https://github.com/tlswg/tls13-spec/issues/1001) and [Single-Use Tickets](https://github.com/tlswg/tls13-spec/pull/1034). In short, I would like to provide anti-replay features.
Please read the documentation of this package (i.e. the beginning of the `SesssionManagers.hs`).

This package can be used for TLS 1.2 or earlier, hence this PR includes the modification of `warp-tls`. This is also used for TLS 1.3 which is implemented currently as one of my topic branch.

Any suggestions are welcome.


kazu-yamamoto:
@ocheron You might be interested in this.

kazu-yamamoto:
@snoyberg Please hold merging until a new version of `psqueues` is released.
See https://github.com/bttr/psqueues/issues/18

ocheron:
> You might be interested in this.

Session resumption is certainly desirable and the analysis you reference is really interesting.

If I understand correctly, security concerns come only when using session tickets or 0-RTT, which we don't have currently in tls. Session manager API could have two functions `sessionResume` and `sessionResumeOnlyOnce`, so that sessions are resumed multiple times when safe.

kazu-yamamoto:
@ocheron Thank you for your suggestions.
That's exactly what I'm wondering now.
I will consider this today.

kazu-yamamoto:
@ocheron Your idea is really cool. I will implement `sessionResumeOnlyOnce` in my TLS 1.3 branch for `tls`.

tls-session-manager/Network/TLS/SessionManager.hs
```diff
@@ -0,0 +1,128 @@
...
+type DB = OrdPSQ SessionID UTCTime Value
+type Item = (SessionID, UTCTime, Value, Operation)
+
+data Operation = Add | Del
+data Use = SingleUse | MultipleUse
+data Availability = Fresh | Used
+
+----------------------------------------------------------------
+
+-- | Creating a in-memory session manager.
```
snoyberg: s/a/an

tls-session-manager/Network/TLS/SessionManager.hs
```diff
@@ -0,0 +1,128 @@
...
+    return SessionManager {
+        sessionResume     = resume reaper MultipleUse
+      , sessionEstablish  = establish reaper lifetime
+      , sessionInvalidate = invalidate reaper
+      }
+
+cons :: Int -> Item -> DB -> DB
+cons lim (k,t,v,Add) db
+  | Q.size db == lim    = case Q.minView db of
+      Nothing          -> Q.insert k t v Q.empty -- not happens, just in case
```
snoyberg: How about adding an `assert False` here?

tls-session-manager/Network/TLS/SessionManager.hs
```diff
@@ -0,0 +1,128 @@
...
+      Just (_,_,_,db') -> Q.insert k t v db'
+  | otherwise           = Q.insert k t v db
+cons _   (k,_,_,Del) db = Q.delete k db
+
+clean :: DB -> IO (DB -> DB)
+clean olddb = do
+    currentTime <- getCurrentTime
```
snoyberg: It may be better to use a monotonic clock from the `clock` package to avoid clock skew issues.

(And yes, that same advice probably applies to some of my reaper logic elsewhere like http-client.)

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -129,6 +131,15 @@ data TLSSettings = TLSSettings {
     -- Default: Nothing
     --
     -- Since 3.2.2
+  , tlsSessionManagerConfig :: Maybe SM.Config
+    -- ^ Configuration for in-memory TLS session manager.
+    -- If Nothing, 'TLS.noSessionManager' is used.
```
snoyberg: Why wouldn't we default to using some basic configuration? Is there a downside to turning on a manager by default?

tls-session-manager/Network/TLS/SessionManager.hs
kazu-yamamoto: Thanks. Fixed.

warp-tls/Network/Wai/Handler/WarpTLS.hs
kazu-yamamoto: - It changes the current behavior of warp-tls (i.e. allowing session resumptions).
- It requires more memory than the current.

So, I think that users should explicitly set this parameter.

tls-session-manager/Network/TLS/SessionManager.hs
kazu-yamamoto: Do you refer to `Control.Exception.assert`?
If so, what the second argument should be?

snoyberg: Yes, that's what I meant. I'd leave the second argument as the current right-hand side of the `->` as a fair fallback.

kazu-yamamoto: Thanks and done.

tls-session-manager/Network/TLS/SessionManager.hs
kazu-yamamoto: What constructor of `Clock` should I choose for `getTime` for this purpose?
`Realtime`?

snoyberg: I've typically used `Monotonic`

kazu-yamamoto:
@snoyberg A new `psqueues` has been released.
I fixed all issues and rebased this branch for merging.
Please review this again and merge this PR.

kazu-yamamoto:
I have removed the patch for warp-tls so that all tests should pass in CI.

kazu-yamamoto:
Thank you!

snoyberg:
A pleasure :)

kazu-yamamoto:
`sessionResumeOnlyOnce` is being implemented in the `tls13-18` branch of my `tls` repo.

PR #631, opened 2017-07-20
Document headers required for SSE through Nginx

mschristiansen:
Add documentation for HTTP headers required when running SSE behind Nginx. This is for instance required when running on AWS and probably for other proxy servers.

I was looking at updating the `app` example, but it is rather badly broken. Does it need to be removed or should I add it to the cabal file and fix it?


snoyberg:
Adding the executable to the cabal file (behind a flag that disables it by default) would be great. We could have the flag turned on in the stack.yaml file so that CI notices when the example stops building, without causing people installing from Hackage/Stackage to have to build the example.

But I'll merge in this doc change now.

PR #633, opened 2017-07-24
GHC 7.10 or later

kazu-yamamoto:
GHC 8.2 has been released.
Let's remove `#if` as much as possible.


kazu-yamamoto:
Thanks.
I will do it after I fix the check on Travis CI.

kazu-yamamoto:
- The travis ci problem has been fixed.
- wai and warp-tls also supports GHC 7.10 or later only.
- base now requires '>= 4.8'.
- bytestring-builder was removed.
- This branch was rebased onto master.

Now I think this PR is ready for merging.

PR #638, opened 2017-08-07
Update SSE example in wai-extra. Fixes #635

### LINKED ISSUE #635: Broken app example in wai-extra
The [example](https://github.com/yesodweb/wai/blob/master/wai-extra/app/Main.hs) in `wai-extra` needs an update. I was looking at doing it and got some advice / guidelines from @snoyberg in #631 (see below), since it is also not in the cabal file.

> Adding the executable to the cabal file (behind a flag that disables it by default) would be great. We could > have the flag turned on in the stack.yaml file so that CI notices when the example stops building, without > causing people installing from Hackage/Stackage to have to build the example.

It isn't an urgent issue and I'm happy self-assign and fix once I'm a bit ahead on my current project.

mschristiansen:
Some updates to the app example in `wai-extra`, which wasn't building.

- Include example in cabal behind flag
- Add `README.md` with app instructions
- Removed `eventSourceAppSource` endpoint (was from before wai-eventsource merged with wai-extra
- Add "fake" favicon to `index.html` to avoid favicon requests causing error responses

Haven't changed the `travis` config, but would be nice to have app example build as part of CI.


wai-extra/wai-extra.cabal
```diff
@@ -81,6 +81,11 @@ extra-source-files:
   ChangeLog.md
   README.md
 
+flag build-app
+  description:        Build app executable.
+  manual:             True
+  default:            False
```
snoyberg: How about enabling this flag in the `stack.yaml` so that CI will build the executable?

mschristiansen: Sure. Will do that.

Now that I am making changes what do you think about calling it `example` instead of `app` and merge the `app` README with the `wai-extra` README?

I'll rename the flag and directory from `app` to `example` and by adding the README info to the project info the example will at least be visible to people not browsing the source. (The current project README is *minimal*.)

snoyberg: That all sounds great, thank you!

mschristiansen:
@snoyberg Think that should be it.

wai-extra/README.md
```diff
@@ -1,3 +1,16 @@
...
+
+## Example using Server-Sent Events ##
+
+There is a small example using Server-Sent Events (SSE) in the
+`./example` directory.
+
+Run the commands below to start the server on http://localhost:8080
+
+```
+$ cabal configure -f build-example
```
snoyberg: Let's stick to Stack recommendations here, which will deal with local packages and not drop people into dependency solving problems.

    stack build .
    stack exec example

mschristiansen:
Changed to Stack. One build failure, which seems to be something Ruby and unrelated to this PR.

snoyberg:
Thank you!

PR #641, opened 2017-08-15
warp-tls: ignore socket errors while sending close_notify

### LINKED ISSUE #640: Trivial server throws exception when run with warp-tls (but not with warp)
I have a trivial warp (and warp-tls) server that just responds with 200 to all
requests: Here's the version that serves over HTTP (compile and run with
lts-9.0):

```haskell
{-# LANGUAGE LambdaCase          #-}
{-# LANGUAGE OverloadedStrings   #-}
{-# LANGUAGE ScopedTypeVariables #-}

module Main where

import Control.Concurrent
import Control.Concurrent.Async
import Control.Exception
import Network.HTTP.Types.Status
import Network.Wai
import Network.Wai.Handler.WarpTLS
import System.Environment

main :: IO ()
main = do
    [port_str] <- getArgs
    let
      port = read port_str

      exc_handler _ exc =
        putStrLn ("exception happened: " ++ show exc)

      ready = putStrLn "ready"

      settings =
        setPort port $
        setOnException exc_handler $
        setBeforeMainLoop ready $
        defaultSettings

      tls_settings = tlsSettings "cert.pem" "key.pem"

    warp_done <- newEmptyMVar

    -- withAsync (runTLS tls_settings settings app `finally` tryPutMVar warp_done ()) $ \server_thr -> do
    withAsync (runSettings settings app `finally` tryPutMVar warp_done ()) $ \server_thr -> do
      readMVar warp_done
      poll server_thr >>= \case
        Just (Left (exc :: SomeException))
          | Just (_ :: IOException) <- fromException exc
          -> putStrLn ("IOException: " ++ show exc)
          | otherwise
          -> putStrLn ("Exception: " ++ show exc)
        other -> print other

app :: Application
app _req resp = do
    -- threadDelay (10*1000*1000)
    resp (responseLBS status200 [] "Done")

```

I can test this with curl:

```
$ curl http://127.0.0.1:1234
Done
```

This version works fine. However, if I comment out the `runSettings` line and
enable the line before that (`runTLS` line), after each request I get this line
printed in the server:

(test with `curl https://127.0.0.1:1234 --insecure`)

```
Warp: Client closed connection prematurely
```

This is printed after every single request. I added a print statement in
warp-tls to get a better feeling of what may be causing this, and it looked like
this:

```
Network.Socket.sendBuf: resource vanished (Broken pipe)
Warp: Client closed connection prematurely
```

It seems like the other end is closing the TCP connection as expected, but warp
(or warp-tls, or tls) is trying to send more stuff after that.

I'm not sure if this is critical (a bug) or not, but if it's not maybe this
shouldn't be printed at all.

(Note: this line isn't printed by `exc_handler` function)

osa1:
(fixes #640)

RFC 5246 section 7.2.1 says this about closure alerts:

> The client and the server must share knowledge that the connection is ending
> in order to avoid a truncation attack.  Either party may initiate the
> exchange of closing messages.
>
> [...]
>
> Unless some other fatal alert has been transmitted, each party is required to
> send a close_notify alert before closing the write side of the connection.
> The other party MUST respond with a close_notify alert of its own and close
> down the connection immediately, discarding any pending writes.  It is not
> required for the initiator of the close to wait for the responding
> close_notify alert before closing the read side of the connection.

This `close_notify` message is sent by `TLS.bye`, and it used to generate an
exception when the receiving side closes the connection before waiting for
`close_notify`. But this scenario is perfectly legal according to the RFC, so
we should ignore `ConnectionClosedByPeer` exception that happens during
`TLS.bye`.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -47,9 +47,9 @@ module Network.Wai.Handler.WarpTLS (
 import Control.Applicative ((<$>))
 #endif
 import Control.Applicative ((<|>))
-import Control.Exception (Exception, throwIO, bracket, finally, handle, fromException, try, IOException, onException, SomeException(..))
+import Control.Exception (Exception, throwIO, bracket, finally, handle, fromException, try, IOException, onException, SomeException(..), handleJust)
 import qualified Control.Exception as E
-import Control.Monad (void)
+import Control.Monad (void, guard)
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as L
 import Data.Default.Class (def)
@@ -348,8 +348,16 @@ httpOverTls TLSSettings{..} s bs0 params = do
         sendfile fid offset len hook headers =
             readSendFile writeBuf bufferSize sendall fid offset len hook headers
 
-        close' = void (tryIO $ TLS.bye ctx) `finally`
+        close' = void (tryIO sendBye) `finally`
                  TLS.contextClose ctx
+
+        sendBye =
+          -- It's fine if the connection was closed by the other side before
+          -- receiving close_notify, see RFC 5246 section 7.2.1.
+          handleJust
+            (\e -> guard (e == ConnectionClosedByPeer) >> return e)
+            (const (return ()))
+            (TLS.bye ctx)
 
         -- TLS version of recv with a cache for leftover input data.
         -- The cache is shared with recvBuf.


kazu-yamamoto:
@snoyberg Is it possible to distinguish "closed by me" and "closed by peer" in the Warp level?

Sorry but I'm very busy in August due to NetPL 2017. So, I cannot take time to consider this issue deeply at this moment.

snoyberg:
I don't think that ability exists, but this approach (ignoring the exception) seems fine to me. Do you have an objection to me reviewing/merging this @kazu-yamamoto?

Good luck at NetPL :)

kazu-yamamoto:
No objection!
And thank you!

snoyberg:
Thanks!

PR #653, opened 2017-11-07
Remove unused language extensions.

iporsut:
Try to remove extension and build again and compile successful.
So I think it's necessarily use that extensions.


snoyberg:
Thanks!

PR #654, opened 2017-11-10
[wai-extra] Set `requestBodyLength` for `srequest`, fixes #652

### LINKED ISSUE #652: wai-extra: srequest doesn't properly set requestBodyLength
Consider the following contrived example:

```

import Network.Wai
import Network.Wai.Test

import Control.Monad

main :: IO ()
main = void $ runSession sess app

sess = srequest (SRequest defaultRequest "test")

app :: Application
app req _ = do
  print $ requestBodyLength req
  return $ error "ResponseReceived"
```
The code will print `KnownLength 0`, even though the request body has been set to "test".

jbransen:


snoyberg:
Could you make a patch version bump to the cabal file and update the ChangeLog with a description of the change?

jbransen:
Done. Some tests seem to be failing though, it is not clear to me why.

snoyberg:
The tests are definitely failing for me too, which will prevent this patch from being accepted.

jbransen:
I have fixed this in a better way, which should make all tests succeed again. The tests are not so nicely designed though, my previous fix was also fine, but the tests depend on the logger, which only logs when the content body length is known.

snoyberg:
Thanks! If you have a recommendation for a better way to write the tests, please let me know.

snoyberg:
I'm actually now getting downstream test failures because of this change, specifically in yesod-core, which is expecting to have direct control of the reported content-length. I understand your motivation here, but given that this is a silently breaking change, I'm going to have to revert it. A better path forward would be a new function that automatically sets the reported length.

PR #658, opened 2017-11-17
unsetting http?_proxy for testing.

kazu-yamamoto:
If we set the "http_proxy" env var, `curl` connects to it instead of localhost, resulting test failures. This patch ensures that `curl` does not use proxies.

After submitting your PR:


snoyberg:
Makes sense to me

PR #659, opened 2017-11-17
Add OverloadedStrings to default-extensions in test-suite

mschristiansen:
Busy adding tests for gzip middleware in preparation for some refactoring and adding overloadedstrings to the test suite seemed quite obvious.


snoyberg:
I don't think it makes sense to add this. Most of the packages I've seen (and definitely the ones I maintain) prefer to keep the pragmas inside the modules themselves. Do you have a specific motivation for this change?

mschristiansen:
Motivation: 1 addition, 7 deletions. Also prefer extensions in module, except when same in every single module :-)

snoyberg:
I understand your request, but it contradicts the current style/approach for the project.

PR #660, opened 2017-11-24
Mega imports

kazu-yamamoto:
We have been suffering from warnings ("the import is redundant") due to GHC versions (e.g. `Control.Applicative`). Recently, a magic solution has been found. The solution is that we provide a *mega* import module. Other modules import this and if at least one difinition is used in a module, no warning displays. This magically works in many cases. No `#ifdef` is necessary.

We have already stopped support of GHC 7.8. So, we are not suffering anymore. But the same things would happen in the future. Thus, I would suggest to take this solution.

- The first commit implements this.
- The second commit adopts hlint's suggestions since `Data.Maybe` is imported by default.
- The third commit uses `NonEmpty` which is also imported by default.


kazu-yamamoto:
Merged and pushed. Thank you for reviewing!

PR #662, opened 2017-12-17
Export Network.Wai.Parse.noLimitParseRequestBodyOptions

3noch:


wai-extra/Network/Wai/Parse.hs
```diff
@@ -22,6 +22,7 @@ module Network.Wai.Parse
     , parseContentType
     , ParseRequestBodyOptions
     , defaultParseRequestBodyOptions
+    , noLimitParseRequestBodyOptions
```
snoyberg: Good catch. Since this is newly exposed, can I ask you to:

1. Add a minor version bump to the .cabal file
2. Add a `ChangeLog` entry
3. Add a `@since` Haddock comment to this function

3noch: Sure. `@since` is already there but I suppose it's wrong since it wasn't ever *actually* exposed.

snoyberg: Yup, that sounds like a mistake on my part. If you could change the `@since` comment, that would be great.

3noch: Done.

PR #666, opened 2018-01-15
Adds support for upcoming streaming-commons 0.2 and resourcet 1.2

snoyberg:


kazu-yamamoto:
@snoyberg I have merged this.

I rebased this PR onto master to make the history linear. The advantage of this policy is make the maintenance much easier. The disadvantage is that PR is not marked with MERGED.

PR #672, opened 2018-01-31
Add call stacks to assertion functions in wai-test

cloudhead:
wai-test: Use `HasCallStack` on assertion functions so we get stack traces on errors.


cloudhead:
Will update to support GHC 7.10.x

PR #674, opened 2018-02-15
Add support for windows thread block hack and closeOnExec to TLS.

ondrap:

(the exposed functions are exposed because

On Windows the TLS web server cannot be killed; this is especially problematic if you run it as windows service and it cannot be stopped by normal means. It seems that some things that are done when opening the normal `runSettings` from `warp` are not performed in `runTLS`. This patch sets CloseOnExec flag on the socket and uses the windows hack to let the thread to be killed.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -293,7 +293,12 @@ alpn xs
 
 getter :: TLS.TLSParams params => TLSSettings -> Socket -> params -> IO (IO (Connection, Transport), SockAddr)
 getter tlsset@TLSSettings{..} sock params = do
+#if WINDOWS
+    (s, sa) <- windowsThreadBlockHack $ accept sock
+#else
     (s, sa) <- accept sock
+#endif
+    setSocketCloseOnExec s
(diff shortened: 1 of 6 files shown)


kazu-yamamoto:
@ondrap This patch is really nice. Thank you!

@snoyberg I dislike that this kind of PRs includes version-bump. What do you think?

snoyberg:
I know that opinions on this go both ways, but I've encouraged contributors to include version bumps, see the second bullet on https://www.snoyman.com/blog/2017/06/how-to-send-me-a-pull-request. I'm fine with running Warp differently, we should probably include that in a CONTRIBUTING.md file.

kazu-yamamoto:
@snoyberg I read it. But I don't see why you like it. What if multiple PRs come in a short period?

snoyberg:
In my experience, it hasn't been an issue. U tend to release very quickly
after merging a PR, and can always revise an unnecessary version bump if I
haven't released yet. On the other hand, there's a very real possibility of
forgetting to add comments to new identifiers indicating the version they
were added.

kazu-yamamoto:
I understand. Let's go with your way.

And this PR is perfectly OK with me.

snoyberg:
Thanks!

PR #689, opened 2018-04-20
Atomically recv() wrt. capturing errno

mitchellwrosen:
Hi there, consider this a (basic) question with some code attached.

Concurrent Haskell code that manually checks `errno` after a foreign call seems broken to me, since `errno` it is shared among all threads on an OS thread (though I believe GHC does carry it around in the TSO for migrating threads).

For example, the code in `recvloop` reads:

```haskell
do bytes <- c_recv
   -- (1)
   if bytes == -1
     then do
       -- (2)
       errno <- getErrno
       if errno == eAGAIN
         then loop
         else throw errno
   else success
```   

It's possible that between `(1)` and `(2)`, some other foreign call is scheduled and clobbers our `errno`.

So, in this patch I've simply swapped out `recv` for a custom version that returns `-errno` instead of `-1` on error. That way, the foreign call "atomically" captures `errno` since no Haskell thread can be sneakily scheduled on the same OS thread between `(1)` and `(2)` below, both in C:

```c
my_recv() {
  result := recv()
  // (1)
  if (result == -1) {
    // (2)
    return -errno
  } else {
    return result
  }
```

Am I right about this? Thanks for looking.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Recv.hs
+++ b/warp/Network/Wai/Handler/Warp/Recv.hs
@@ -12,7 +12,7 @@ module Network.Wai.Handler.Warp.Recv (
 import qualified Control.Exception as E
 import qualified Data.ByteString as BS
 import Data.IORef
-import Foreign.C.Error (eAGAIN, getErrno, throwErrno)
+import Foreign.C.Error (Errno(Errno), eAGAIN, errnoToIOError)
 import Foreign.C.Types
 import Foreign.ForeignPtr (withForeignPtr)
 import Foreign.Ptr (Ptr, castPtr, plusPtr)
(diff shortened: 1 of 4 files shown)


mitchellwrosen:
I found this ancient thread in which Simon Marlowe assures us there's actually no race condition: http://haskell-cafe.haskell.narkive.com/FbVXpWBG/errno-handling-in-concurrent-haskell

snoyberg:
Given that thread from Simon, any reason to still consider this PR, or should it be closed?

mitchellwrosen:
It can be closed, I was just leaving it up in case you had any comment. Errno is green-thread-local in haskell according to that forum post and this code:

```haskell
import Control.Concurrent
import Foreign.C.Error

foreign import ccall unsafe "junk"
  junk :: Int -> IO ()

main :: IO ()
main = do
  baton1 <- newEmptyMVar
  baton2 <- newEmptyMVar
  done <- newEmptyMVar

  forkOn 1 $ do
    putStrLn "thread 1: setting errno to 5"
    junk 5
    putMVar baton2 ()
    takeMVar baton1
    putStrLn "thread 1: checking errno"
    getErrno >>= \(Errno n) -> print n
    putMVar done ()

  forkOn 1 $ do
    takeMVar baton2

    putStrLn "thread 2: checking errno"
    getErrno >>= \(Errno n) -> print n

    putStrLn "thread 2: setting errno to 10"
    junk 10
    putMVar baton1 ()
    putMVar done ()

  takeMVar done
  takeMVar done
```
```c
#include <errno.h>
void junk(int n) {
  errno = n;
}
```
```
// Output
thread 1: setting errno to 5
thread 2: checking errno
0
thread 2: setting errno to 10
thread 1: checking errno
5
```

PR #696, opened 2018-07-08
Don't make me use Data.Default

bitemyapp:

After submitting your PR:

- [X] Update the Changelog.md file with a link to your PR


wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -81,12 +84,19 @@ data RequestLoggerSettings = RequestLoggerSettings
     , destination :: Destination
     }
 
+-- | @defaultRequestLoggerSettings@ is the original 'Default' instance
+--   factored out into a top-level declared variable.
+--   @since 3.0.22.2
```
MaxGabriel: This is just from memory so could be wrong, but does the @since need to be separated by an empty comment to work?

Vlix:
`Network.Wai.Middleware.RequestLogger` has a `defaultRequestLoggerSettings` now, which means you don't need to use `Data.Default`, so I'll just close this PR.

Issue #701, opened 2018-07-26
setTimeout doesn't end request


pbrisbin:
I might be misunderstanding how this is supposed to work, but...

Given:

```hs
#!/usr/bin/env stack
-- stack script --resolver lts-11.17
-- vim: ft=haskell
{-# LANGUAGE OverloadedStrings #-}

module Main (main) where

import Network.Wai
import Network.HTTP.Types (status200)
import Network.Wai.Handler.Warp
import Control.Concurrent (threadDelay)

application :: Application
application _ respond = do
    threadDelay $ 10 * 1000000
    respond $ responseLBS status200 [("Content-Type", "text/plain")] "Hello World\n"

main :: IO ()
main = runSettings settings application
  where
    settings = setTimeout 5 $ setPort 3000 defaultSettings
```

I'd expect, this:

```console
% time curl localhost:3000
```

To complete in close to 5 seconds with an error.

Instead, it happily runs for 10 seconds.

```console
% time curl localhost:3000
Hello World
curl localhost:3000  0.01s user 0.00s system 0% cpu 10.226 total
```

In the context of our larger application, we also see this behavior with a more realistic slow request such as `runDB $ rawSql "SELECT pg_sleep(10);" []`.

pbrisbin:
Hmm, I'm a little confused. Are you saying that `setTimeout` is specifically a slowloris-related setting, to give a lower bound on how quickly that mechanism *may* kick in? Or that the two timeout mechanisms just interact in a surprising way?

dredozubov:
After I modified a script by @pbrisbin to have a timeout of 1 second and 2 second delay I got this:
```
% for i in `seq 1 30`; do time curl localhost:3000; done
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.018 total
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.014 total
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.017 total
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.020 total
Hello World curl localhost:3000  0.01s user 0.01s system 0% cpu 2.058 total
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.018 total
Hello World curl localhost:3000  0.00s user 0.01s system 0% cpu 2.016 total
Hello World curl localhost:3000  0.00s user 0.00s system 0% cpu 2.018 total
[...]

PR #702, opened 2018-07-30
Add Middleware.Timeout

pbrisbin:
See #701.

Wraps the inner `respond` in `System.Timeout.timeout` with the given seconds
and, by default, responds `status503`. The main `timeout` is implemented in
terms of `timeoutStatus` for customizing only the response status, which is
itself implemented in terms of `timeoutAs`, for customizing the `Response`
entirely.

Some things I'm not sure of:

1. `timeout` is a bit of a bold name, but it seems in line with how all
   `Middleware`s name themselves: assuming a very narrow context
1. Is `timeoutStatus` useful, or is just `timeout`/`timeoutAs` enough?
1. If we keep `timeoutStatus`, should `timeoutAs` be named `timeoutResponse`,
   which feels more consistent?
1. Should we accept microseconds directly? Is there ever a use-case for a
   timeout smaller than one second? (it would be nice to not have to wait an
   entire second for each of the spec examples)

Any other tweaks or style changes, don't hesitate to let me know.

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,3 +1,7 @@
+## 3.0.24.0
+
+* Add timeout middleware [#702](https://github.com/yesodweb/wai/pull/702).
+
 ## 3.0.23.0
 
 * Add rewriteRoot middleware [#697](https://github.com/yesodweb/wai/pull/697).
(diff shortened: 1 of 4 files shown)


pbrisbin:
Will do!

pbrisbin:
@snoyberg this should be all set now :)

PR #704, opened 2018-08-10
Use return rather than pure for GHC 7.8 compatibility

tmcgilchrist:
Introduced in wai-extra-3.0.24.0 this breaks building on 7.8.4.
I noticed that the Travis CI doesn't include 7.8.4 anymore so if it's not supported let me know.


bitemyapp:
GHC 7.10.3 is `base-4.8.2.0`, `wai`'s lower bound for `base` is `4.8`, so I think GHC 7.8 is not supported. [cf.](https://www.snoyman.com/base)

Accordingly, I'm closing. Apologies if this wasn't explicit enough.

Issue #705, opened 2018-08-11
wai-extra: ResourceCleanupException in tempFileBackEndOpts


dgaw:
After upgrading from stack lts-9.13 to lts-11.20 I noticed that my Spock web app started throwing the following exception when handling post requests:

`ResourceCleanupException {rceOriginalException = Nothing, rceFirstCleanupException = /tmp/webenc29043-0.buf: removeLink: does not exist (No such file or directory), rceOtherCleanupExceptions = []} `

I've traced the problem to the tempFileBackEndOpts function in the Parse module of wai-extra, specifically this line:
https://github.com/yesodweb/wai/blob/master/wai-extra/Network/Wai/Parse.hs#L131

Shouldn't removeFile check if the file exists first? I've added this check and it worked for me but I'm not sure if this isn't an indication of a bug somewhere else.

Can somebody else comment on this?

snoyberg:
You're mostly correct, though it's better to catch the not found exception
to avoid a race condition. PR welcome.

snoyberg:
Fixed by #706

PR #706, opened 2018-08-14
wai-extra: Catch the "file not found" exception in tempFileBackEndOpt…

dgaw:
A fix for #705


dgaw:
Thanks for the quick review. I've bumped the minor version and added a change log entry.

snoyberg:
Thanks!

PR #710, opened 2018-08-20
Helpers for modifying request headers as middleware

bitemyapp:

Yep, minor bump because it's a backwards compatible addition.

After submitting your PR:

### DIFF
--- a/wai/Network/Wai.hs
+++ b/wai/Network/Wai.hs
@@ -352,6 +352,6 @@ lazyRequestBody req =
                 bss <- loop
                 return $ LI.Chunk bs bss
 
--- | Apply the provided function to the response header list of the Response.
+-- | Apply the provided function to the request header list of the Request.
 mapRequestHeaders :: (H.RequestHeaders -> H.RequestHeaders) -> Request -> Request
 mapRequestHeaders f request = request { requestHeaders = f (requestHeaders request) }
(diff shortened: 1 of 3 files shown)


wai/Network/Wai.hs
```diff
@@ -303,6 +306,10 @@ defaultRequest = Request
 type Middleware = Application -> Application
 
 
+-- | apply a function that modifies a request as a 'Middleware'
+modifyRequest :: (Request -> Request) -> Middleware
+modifyRequest f app req respond = app (f req) $ respond
```
snoyberg: You can eta reduce and remove the `respond`.

wai/Network/Wai.hs
```diff
@@ -303,6 +306,10 @@ defaultRequest = Request
 type Middleware = Application -> Application
 
 
+-- | apply a function that modifies a request as a 'Middleware'
```
snoyberg: Can you add a `@since` comment here?

wai/Network/Wai.hs
```diff
@@ -344,3 +351,9 @@ lazyRequestBody req =
             else do
                 bss <- loop
                 return $ LI.Chunk bs bss
+
+-- | Apply the provided function to the request header list of the Request.
```
snoyberg: Perhaps add a comment about this being intended to work with `modifyRequest`?

Vlix:
Added to `wai-3.2.4`

Issue #708, opened 2018-08-17
graceful shutdown only fires for `IOException`s


mitchellwrosen:
Hi, I discovered that the graceful shutdown mechanism only works when the web server experiences an `IOException`. Is this the intended behavior? (It makes it so Ctrl-C hard-quits a warp server without gracefully shutting down, for example) Here is the relevant code from `Network.Wai.Handler.Warp.Run`:

```haskell
acceptConnection :: Settings
                 -> IO (IO (Connection, Transport), SockAddr)
                 -> Application
                 -> Counter
                 -> InternalInfo0
                 -> IO ()
acceptConnection set getConnMaker app counter ii0 = do
    void $ mask_ acceptLoop
    gracefulShutdown set counter
  where
    acceptLoop = do
        allowInterrupt

        mx <- acceptNewConnection
        case mx of
            Nothing             -> return ()
            Just (mkConn, addr) -> do
                fork set mkConn addr app counter ii0
                acceptLoop

    acceptNewConnection = do
        ex <- try getConnMaker
        case ex of
            Right x -> return $ Just x
            Left e -> do
                let eConnAborted = getErrno eCONNABORTED
                    getErrno (Errno cInt) = cInt
                if ioe_errno e == Just eConnAborted
                    then acceptNewConnection
                    else do
                        settingsOnException set Nothing $ toException e
                        return Nothing
```

The `try getConnMaker` is called at type `IO a -> IO (Either IOException a)`.

Thanks.

snoyberg:
This code isn't for graceful shutdown. It's to deal with a specific corner case of file descriptors becoming exhausted IIRC (CC @kazu-yamamoto). We should probably have a comment explaining that.

We could change this by using `finally` or similar to ensure that graceful shutdown kicks in, but I'm not sure it will behave properly in cases of things like async exceptions. The real question is what the expected semantics of this feature are in the different kinds of shutdown scenarios, which I'm not sure of.

mitchellwrosen:
Can you clarify what you mean by this?

> This code isn't for graceful shutdown.

Is there some other code that is for graceful shutdown?

The basic problem I'm trying to solve: Ctrl-C on a Warp server only shuts down after all outstanding requests are handled. Is that possible with the current Warp API? Thanks :)

snoyberg:
[...]

PR #716, opened 2018-10-10
Improve graceful shutdown docs a bit

mitchellwrosen:
#708


kazu-yamamoto:

PR #726, opened 2019-01-21
Deprecate requestBody

### LINKED ISSUE #669: Long-time renaming request: Make clear in the name that requestBody returns just one chunk
My biggest gripe with `wai` is `requestBody :: Request -> IO ByteString`.

> Get the next chunk of the body. Returns `empty` when the body is fully consumed.

Even though the docs say that it gets only the next chunk, not the full request body, this is just too easy to forget.

```haskell
bs <- requestBody req
```

is too easily written, it compiles, it looks correct, it works in most of the cases, but it is wrong: when you get over 16 KiB (16384 Bytes), stuff starts breaking.

It would be nice if over coming major releases we could fade its usage over to `requestBodyChunk` or `requestBodyGetNextChunk` or something like this.

MaxGabriel:
Closes #669

This implements the deprecation proposed in #669.

One thing to check: is a non-breaking version bump OK? I don't consider this a breaking change, but iirc the PVP does because people could be building with -Werror.

### DIFF
--- a/wai/ChangeLog.md
+++ b/wai/ChangeLog.md
@@ -1,3 +1,7 @@
+## 3.2.2
+
+* Deprecate `requestBody` in favor of the more clearly named `getRequestBodyChunk`. [#726](https://github.com/yesodweb/wai/pull/726)
+
 ## 3.2.1.2
 
 * Remove dependency on blaze-builder [#683](https://github.com/yesodweb/wai/pull/683)
--- a/wai/Network/Wai.hs
+++ b/wai/Network/Wai.hs
@@ -33,6 +33,8 @@ include:
 [wai-test] <http://hackage.haskell.org/package/wai-test>
 
 -}
+-- Ignore deprecations, because this module needs to use the deprecated requestBody to construct a response.
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
 module Network.Wai
     (
       -- * Types
@@ -325,7 +327,7 @@ strictRequestBody req =
     loop id
   where
     loop front = do
-        bs <- requestBody req
+        bs <- getRequestBodyChunk req
         if B.null bs
             then return $ front LI.Empty
             else loop (front . LI.Chunk bs)
@@ -339,7 +341,7 @@ lazyRequestBody req =
     loop
   where
     loop = unsafeInterleaveIO $ do
-        bs <- requestBody req
+        bs <- getRequestBodyChunk req
         if B.null bs
             then return LI.Empty
             else do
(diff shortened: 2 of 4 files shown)


StevenXL:
@MaxGabriel You beat me to it. 

I don't like the idea of adding the `DEPRECATED` pragma, because `requestBody` is not really deprecated. It must still be used as a getter.

Instead, I'd like to add to the docs of `requestBody` something along the lines of: "Use `requestBody` as setter. For new code, use `getRequestBodyChunk` as getter." 

Something similar was done to in [in this PR](https://github.com/yesodweb/wai/commit/54f4c711da09ca014456aafffc16c5646b2af255#diff-f75074d25c22148ba7161a2341173676R234), which I believe you are also involved in. 

That being said, if we do want to add the `DEPRECATED` pragma, I don't think there is any need to wait.

MaxGabriel:
It's true that you still need to use `requestBody` to set the body, but it's only exposed in the `Internal` module, which to me makes it more reasonable to make users put up with a deprecation warning.

This is the path I was imagining:

1. `requestBody` is deprecated; normal users of Wai get a warning and start transitioning code to `getRequestBodyChunk`. Users of Wai's internal module have a deprecation warning they can't get rid of (but they can ignore it with GHC pragmas)
2. When Wai goes through a major version bump at some point in the future,  the actual record field `requestBody` is renamed to `getRequestBodyChunk` and the standalone `getRequestBodyChunk` is deleted. To users of Wai who have already transitioned their code to `getRequestBodyChunk`, they don't even notice this breaking change. 

Basically this privileges users of Wai that are not using the internal module, which I think is reasonable because normal users vastly outnumber those using private APIs, and by using internal APIs users are already kind of agreeing to deal with annoyances like breaking changes.

Alternate path where `requestBody` isn't deprecated:

1. Normal users don't know to switch their code to `getRequestBodyChunk`. They may not realize the function returns a chunk and have subtle bugs in their code. If they read the docs they may change to `getRequestBodyChunk`, though. Users of the internal API are unaffected.
2. If the `requestBody` field is ever renamed to `getRequestBodyChunk`, this becomes a breaking change without a deprecation for all users of Wai.

nh2:
> Only concern from me: should we have a delay between adding the new function name and adding the deprecation? CC @nh2

@snoyberg I my opinion that's not necessary, I'd care about delays only for things that introduce real breaking changes. Existing users of the function would most likely not even notice they have to do anything until a deprecation is added, so they wouldn't benefit from extra time _before_ the deprecation.

---

I like @MaxGabriel's approach above, because it addresses the key reason I filed #669: The current name facilitates writing subtly bugged code.

PR #732, opened 2019-02-19
Added missing bang, which seems to resolve the memory leak in #728

### LINKED ISSUE #728: Session Manager consumes large amounts of memory
I've attached a simple project in which the warp-tls server consumes large amounts of memory. There is a single client which repeatedly sends requests to the server. The memory profile changes from extremely greedy to constant by varying the parameters of the `tlsSessionManagerConfig` in memleak\Crypto\TLS.hs.
[memleak.zip](https://github.com/yesodweb/wai/files/2784138/memleak.zip)

$ cd memleak
$ stack build
$ stack exec server &
$ stack exec client
Then watch the server's memory usage climb.

I expect that since there is a single client connecting to the server, the session manager settings shouldn't affect *anything*, since all messages should be sent in the same session. Thus there should be exactly one session open on the server at all times. Instead, with the [`defaultConfig`](http://hackage.haskell.org/package/tls-session-manager-0.0.0.2/docs/Network-TLS-SessionManager.html#t:Config), my memory increases over time, and not slowly. After starting at about 12MB of memory, usage had risen to 600MB within an hour (corresponding to the client sending 17,000 requests).

Here are some other settings I used, and their resulting memory usage:

`tlsSessionManagerConfig=Just $ defaultConfig{...}`

`{}`: Over 600 MB in an hour
`{ticketLifetime=10}`: over 400 MB when I killed the process in < 1 hr
`{dbMaxSize=3}`: over 300 MB when I killed the process in < 1 hr
`{pruningDelay=5, dbMaxSize=3}`: over 700 MB when I killed the process in < 1 hr
`{ticketLifetime=10,pruningDelay=5}`: stayed around 37 MB, didn't seem to be rising, or rising very slowly
`{ticketLifetime=10,pruningDelay=5,dbMaxSize=3}`: ran 16 hours and stayed at 38.2 MB

It's unclear why setting the session lifetime affects memory usage, especially since restricting the `dbMaxSize` doesn't help by itself. 

Here's the output for various profiling options. [Note: -hr doesn't work; see [GHC #16182](https://ghc.haskell.org/trac/ghc/ticket/16182)]
memoryprofile

My two questions are: is the session manager even reusing the session? In either case, why do most settings for the manager result in very high memory consumption?

This problem occurs with GHC 8.6.3 on Windows 10 and Ubuntu 18.04. I sent an email to Kazu Yamamoto, maintainer of the tls-session-manager package, but have not received a response.

crockeea:
Fixes #728.

After submitting your PR:

### DIFF
--- a/auto-update/Control/Reaper.hs
+++ b/auto-update/Control/Reaper.hs
@@ -113,7 +113,7 @@ data Reaper workload item = Reaper {
 
 -- | State of reaper.
 data State workload = NoReaper           -- ^ No reaper thread
-                    | Workload workload  -- ^ The current jobs
+                    | Workload !workload  -- ^ The current jobs
 
 -- | Create a reaper addition function. This function can be used to add
 -- new items to the workload. Spawning of reaper threads will be handled
@@ -154,9 +154,9 @@ add settings@ReaperSettings{..} stateRef tidRef item =
       next <- atomicModifyIORef' stateRef cons
       next
   where
-    cons NoReaper      = let !wl = reaperCons item reaperEmpty
+    cons NoReaper      = let wl = reaperCons item reaperEmpty
                          in (Workload wl, spawn settings stateRef tidRef)
-    cons (Workload wl) = let !wl' = reaperCons item wl
+    cons (Workload wl) = let wl' = reaperCons item wl
                          in (Workload wl', return ())
 
 spawn :: ReaperSettings workload item
--- a/auto-update/Control/Reaper.hs
+++ b/auto-update/Control/Reaper.hs
@@ -156,7 +156,7 @@ add settings@ReaperSettings{..} stateRef tidRef item =
   where
     cons NoReaper      = let !wl = reaperCons item reaperEmpty
                          in (Workload wl, spawn settings stateRef tidRef)
-    cons (Workload wl) = let wl' = reaperCons item wl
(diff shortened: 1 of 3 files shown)


auto-update/Control/Reaper.hs
```diff
@@ -156,7 +156,7 @@ add settings@ReaperSettings{..} stateRef tidRef item =
   where
     cons NoReaper      = let !wl = reaperCons item reaperEmpty
                          in (Workload wl, spawn settings stateRef tidRef)
-    cons (Workload wl) = let wl' = reaperCons item wl
+    cons (Workload wl) = let !wl' = reaperCons item wl
```
snoyberg: This is a good catch, thanks, and I'm a bit embarrassed for missing it. Question: to hopefully solve this in a more guaranteed way for the future, could you tell me if making the field of the `Workload` constructor a strict field also solves your problem?

crockeea: New commit with a strict data constructor instead of the bangs.

kazu-yamamoto:
This is not directly related to this issue but we should consider to add the following to `*.cabal` in the future:

```
  default-extensions:  Strict StrictData
```

snoyberg:
@kazu-yamamoto Funnily enough, I was discussing this bug with @bitemyapp and he also mentioned `StrictData`.

PR #736, opened 2019-03-28
Use lookupEnv instead of getEnvironment

andrewthad:


andrewthad:
Travis does not actually fail to build this. It just timed out when building with GHC 8.2.2 for some reason.

andrewthad:
Not really. I was just looking through the implementations of the various runner functions to see how they were related, and I noticed that the environment lookup was goofy.

Issue #647, opened 2017-09-30
Payload Too Large in WAI is 500 instead of 413


michalrus:
Would it be possible here:

https://github.com/yesodweb/wai/blob/a1cb17ff79887e0c33be7fb975b5a0f559e2cf45/wai-extra/Network/Wai/Parse.hs#L634

… to throw something that’d turn into a standard `413 Payload Too Large` response instead of a scary `500 Internal Server Error` with `Maximum size exceeded` logged only to app’s stdout/err (by default)?

This causes https://github.com/haskell-servant/servant-multipart/issues/9 also. :slightly_smiling_face: 

Thank you!

snoyberg:
I don't see a problem with such a change, what do you think @kazu-yamamoto?

@michalrus You interested in sending a PR?

kazu-yamamoto:
I think that we should change this to 413.

michalrus:
Great to hear that you both agree! I’ll take a look at how similar cases are handled over the weekend. If anyone wants to do it earlier, please, do. =)

Thank you!

vlatkoB:
Quite a complex and low-level library.
Where does the `error` get turned into HTTP error code? Tried:
```haskell
- Just max'' | cur > max'' -> E.throwIO $ HTTP2.ConnectionError HTTP2.FrameSizeError "Payload Too Large"
- Just max'' | cur > max'' -> E.throwIO $ HTTP2.ConnectionError (HTTP2.UnknownErrorCode 418) "Payload Too Large"
```
But it doesn't get turned into proper status. The only `catch` on `ConnectionError` I found is in `frameReceiver` (`sendGoaway`) in `Network/Wai/Handler/Warp/HTTP2/Receiver.hs`, but seems that one is not the right one.
 A little help, please.

alpmestan:
I think it's `warp`'s default exception handler that just turns any exception that's not caught by that point into a 500 response: https://github.com/yesodweb/wai/blob/b29e2705d0b4035a17227a8669c6430eb7c9d067/warp/Network/Wai/Handler/Warp/Settings.hs#L174

vlatkoB:
@alpmestan Thanks for the pointer. :-)
I added 413 (Payload too large) and 431 (Headers too large)

Interestingly, Linux and MacOS tests pass, but Windows fails in 1 test.

kazu-yamamoto:
#741 has been merged.

PR #741, opened 2019-04-19
Fix for #647 (Payload Too Large in WAI is 500 instead of 413)

vlatkoB:
- Instead of 'error "Header line length exceeds allowed maximum."'
  throw 'ConnectionError (UnknownErrorCode 431)
                          "Request Header Fields Too Large"'
- Instead of 'error "Maximum size exceeded"'
  throw 'ConnectionError (UnknownErrorCode 413) "Payload Too Large"'

I have :

- [X] Added Since declarations to the Haddock

### DIFF
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -4,6 +4,10 @@
   value of the `PORT` environment variable
   [#736](https://github.com/yesodweb/wai/pull/736)
 
+* Throw 413 for too large payload
+* Throw 431 for too large headers
+  [#741](https://github.com/yesodweb/wai/pull/741)
+
 ## 3.2.26
 
 * Support network package version 3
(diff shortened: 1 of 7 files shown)


kazu-yamamoto:
LGTM. @snoyberg What do you think?

warp/ChangeLog.md
```diff
@@ -1,3 +1,9 @@
+## 3.2.28
```
snoyberg: Looks like this can be included in 3.2.27, since that version hasn't been released to Hackage yet.

vlatkoB: I saw a discrepancy between changelog and cabal, so I raised it a bit more. 
Should I revert it or will you?

kazu-yamamoto:
I'm waiting for the result of CI.

kazu-yamamoto:
CI fails. But I believe that the failures are not due to this PR.
So, I have merged this.

PR #750, opened 2019-06-03
Time manager

kazu-yamamoto:
As an excise of HTTP/3 (or HTTP/2 over QUIC), I'm now extracting HTTP server library from Warp. This approach seems promising now. For this purpose, it is also necessary to extract the time out manager.

@snoyberg Would you check out the following points?
- Is license proper?
- Is the module name proper?


snoyberg:
Yes, I think it looks OK to me.

kazu-yamamoto:
Thank you for your review!
Merged and uploaded.

PR #752, opened 2019-06-12
using strict language extensions.

kazu-yamamoto:
This patch uses `Strict` and `StrictData` in wai, warp, warp-tls and auto-update.
I confirmed that this improves memory footprint.

@snoyberg Would you please explain why laziness is necessary for `requestBody`?

### DIFF
--- a/auto-update/auto-update.cabal
+++ b/auto-update/auto-update.cabal
@@ -21,6 +21,8 @@ library
   other-modules:       Control.AutoUpdate.Util
   build-depends:       base >= 4 && < 5
   default-language:    Haskell2010
+  if impl(ghc >= 8)
+      default-extensions:  Strict StrictData
 
 -- Test suite is currently not robust enough, gives too many false negatives.
 
--- a/warp-tls/warp-tls.cabal
+++ b/warp-tls/warp-tls.cabal
@@ -8,7 +8,7 @@ Maintainer:          michael@snoyman.com
 Homepage:            http://github.com/yesodweb/wai
 Category:            Web, Yesod
 Build-Type:          Simple
-Cabal-Version:       >=1.6
+Cabal-Version:       >= 1.10
 Stability:           Stable
 description:         SSLv1 and SSLv2 are obsoleted by IETF.
                      We should use TLS 1.2 (or TLS 1.1 or TLS 1.0 if necessary).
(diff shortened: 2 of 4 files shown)


snoyberg:
I have a (perhaps irrational) concern about these language extensions, since they non-locally modify the behavior of code, making it confusing to read a source file. I'm not 100% opposed though.

For `requestBody` (and other fields): the advantage of lazy fields is that someone can only fill in the ones they are using, e.g. for testing. I don't think this is a strong argument. However, changing the field to strict _is_ a breaking change, so I'd rather avoid it. Additionally: a field with an `IO` action in probably doesn't benefit much, if at all, from strictness.

kazu-yamamoto:
I understand your psychological barrier. :-) But honestly, I don't want to spend my time to seek the source of memory leak due to lazy evaluation anymore. :(

For experience of 10-years network programming in Haskell, I don't see any benefits of lazy evaluation except:

- `processing list data structures`
- `atomicModifyIORef'`

As SPJ described, lazyness is good test cases for purity. But it matters for Haskell compiler developers, not for users.

For months, I have been using `Strict` and `StrictData`. Bad things does not happen *implicitly*. They do happen *explicitly* (when running tests), so we can notice them before releasing. An interesting example is here:

https://gitlab.haskell.org/ghc/ghc/issues/16810

For soft-landing, it's OK for me to leave `wai` library as is and to apply these language extensions to libraries which we can convince, such as `warp`, `warp-tls` and `auto-update`.

What do you think?

snoyberg:
I'm definitely on board with the idea that laziness is generally a problem, I'm just worried about creating confusing code. But I'll defer to your judgement here: go for the release, but please leave `wai` alone for the moment.

bitemyapp:
I think @snoyberg already knows this, but I've started kicking around my code like @kazu-yamamoto where I use Strict/StrictData in projects.

kazu-yamamoto:
@snoyberg Thank you for understanding.

I'm now tackling the space leak of my HTTP server. `Strict` helps me a lot. After I find the source of the problem and I'm convinced more, I will ask you to merge `Strict` stuff and release some packages.

kazu-yamamoto:
Tips:

I misunderstood that the heap profile file is generated after a Haskell program (with `+RTS -h`) is finished. But it's growing while the program is running. So, we can observe how memory is leaking without stopping a server with `hp2ps`.

Another issue: `+RTS -hT` cannot be used with GHC 8.6, why? I'm using `+RTS -h` now.

kazu-yamamoto:
I have excluded `wai` and merged this PR.

kazu-yamamoto:
New versions of warp, warp-tls and auto-update have been released.

PR #754, opened 2019-06-21
Using Network.HTTP2.Server.

kazu-yamamoto:
I have extracted HTTP/2 implementation in `warp` and moved it to `Network.HTTP2.Server` in `http2`. This results in so beautiful APIs and I feel it's promising. It is independent from WAI hoping that other HTTP engines can use it. `http2` version 2.0.0 has been already released.

This PR deletes many modules relating to HTTP/2 and uses `Network.HTTP2.Server`.  This implementation passes all tests on my local machine including `h2spec`. Field testing has been done at least two weeks in my web site.

@snoyberg This PR is to telling Warp guys that a big change is coming. Please give a look and tell me if you find something wrong.


kazu-yamamoto:
hackage created its doc: http://hackage.haskell.org/package/http2-2.0.0/docs/Network-HTTP2-Server.html

kazu-yamamoto:
Background: https://kazu-yamamoto.hatenablog.jp/entry/2019/06/21/124224

snoyberg:
I didn't look _that_ thoroughly, but a cursory review looks fine to me. I'm happy with this kind of direction too, thanks for pursuing it. If you're comfortable with it, I'd say go ahead with such a release.

kazu-yamamoto:
Thank you for reviewing!
Rebased and merged.

kazu-yamamoto:
@snoyberg I'm trying to fix tests on Azure CI.
Since doctest fails even on my local machine at this moment, I disabled it.
Now only `Linux cabal-7.10.3` fails.
Would you give a look?
Because I think that this is not due to the HTTP/2 server library, I will release a new version of warp.

snoyberg:
Looks like it's a problem of old Cabal versions not liking the `Strict` and `StrictData` extensions for upload to Hackage. I'll try to put in a workaround.

kazu-yamamoto:
I saw green! Thank you very much!

snoyberg:
My pleasure, thanks for all the other fixes!

Issue #745, opened 2019-05-10
auto-update feature request: be able to control leading vs. trailing edge in Control.Debounce


thomasjm:
Most debounce packages I've seen allow you to specify whether the debounced action should occur on the leading edge or trailing edge of the wait period. For example, [Lodash's version](https://lodash.com/docs/4.17.11#debounce) has a good API.

For some debouncing use-cases, you want the update to happen only after one or more events have happened and things have settled down. But the current package always fires on the leading edge, so I end up with two updates when I would prefer only one.

snoyberg:
Would you be interested in submitting a PR to add this functionality?

thomasjm:
I won't have time in the foreseeable future unfortunately, but maybe eventually if nobody else gets around to it first.

kazu-yamamoto:
Fixed in #756. Let's close.

PR #756, opened 2019-06-24
Leading vs. trailing edges for Control.Debounce

thomasjm:
Okay @snoyberg , took a stab at #745 . What do you think?

After submitting your PR:

### DIFF
--- a/auto-update/Control/Debounce.hs
+++ b/auto-update/Control/Debounce.hs
@@ -11,6 +11,7 @@
 -- printString <- 'mkDebounce' 'defaultDebounceSettings'
 --                  { 'debounceAction' = putStrLn "Running action"
 --                  , 'debounceFreq' = 5000000 -- 5 seconds
+--                  , 'debounceEdge' = 'Trailing' -- Trigger on the trailing edge
 --                  }
 -- @
 --
@@ -25,70 +26,34 @@
 -- @since 0.1.2
 module Control.Debounce
     ( -- * Type
-      DebounceSettings
+      DI.DebounceSettings
     , defaultDebounceSettings
       -- * Accessors
-    , debounceFreq
-    , debounceAction
+    , DI.debounceFreq
+    , DI.debounceAction
+    , DI.debounceEdge
+    , DI.DebounceEdge(..)
       -- * Creation
     , mkDebounce
     ) where
 
-import           Control.Concurrent      (forkIO, threadDelay)
-import           Control.Concurrent.MVar (newEmptyMVar, takeMVar, tryPutMVar)
-import           Control.Exception       (SomeException, handle, mask_)
-import           Control.Monad           (forever, void)
-
--- | Settings to control how debouncing should work.
---
--- This should be constructed using 'defaultDebounceSettings' and record
--- update syntax, e.g.:
---
--- @
--- let settings = 'defaultDebounceSettings' { 'debounceAction' = flushLog }
--- @
---
--- @since 0.1.2
-data DebounceSettings = DebounceSettings
-    { debounceFreq   :: Int
-    -- ^ Microseconds lag required between subsequence calls to the debounced
-    -- action.
-    --
-    -- Default: 1 second (1000000)
-    --
-    -- @since 0.1.2
-    , debounceAction :: IO ()
-    -- ^ Action to be performed.
-    --
-    -- Note: all exceptions thrown by this action will be silently discarded.
-    --
-    -- Default: does nothing.
-    --
-    -- @since 0.1.2
-    }
+import           Control.Concurrent      (newEmptyMVar, threadDelay)
+import qualified Control.Debounce.Internal as DI
 
 -- | Default value for creating a 'DebounceSettings'.
 --
 -- @since 0.1.2
-defaultDebounceSettings :: DebounceSettings
-defaultDebounceSettings = DebounceSettings
-    { debounceFreq = 1000000
-    , debounceAction = return ()
+defaultDebounceSettings :: DI.DebounceSettings
+defaultDebounceSettings = DI.DebounceSettings
+    { DI.debounceFreq = 1000000
+    , DI.debounceAction = return ()
+    , DI.debounceEdge = DI.Leading
     }
 
--- | Generate an action which will trigger the debounced action to be
--- performed. The action will either be performed immediately, or after the
--- current cooldown period has expired.
+-- | Generate an action which will trigger the debounced action to be performed.
 --
 -- @since 0.1.2
-mkDebounce :: DebounceSettings -> IO (IO ())
-mkDebounce (DebounceSettings freq action) = do
-    baton <- newEmptyMVar
-    mask_ $ void $ forkIO $ forever $ do
-        takeMVar baton
-        ignoreExc action
-        threadDelay freq
-    return $ void $ tryPutMVar baton ()
-
-ignoreExc :: IO () -> IO ()
-ignoreExc = handle $ \(_ :: SomeException) -> return ()
+mkDebounce :: DI.DebounceSettings -> IO (IO ())
+mkDebounce settings = do
+  baton <- newEmptyMVar
+  DI.mkDebounce baton threadDelay settings
--- a/auto-update/Control/Debounce.hs
+++ b/auto-update/Control/Debounce.hs
@@ -56,4 +56,4 @@ defaultDebounceSettings = DI.DebounceSettings
 mkDebounce :: DI.DebounceSettings -> IO (IO ())
 mkDebounce settings = do
   baton <- newEmptyMVar
-  DI.mkDebounce baton threadDelay settings
+  DI.mkDebounceInternal baton threadDelay settings
--- a/auto-update/Control/Debounce.hs
+++ b/auto-update/Control/Debounce.hs
@@ -11,7 +11,7 @@
 -- printString <- 'mkDebounce' 'defaultDebounceSettings'
 --                  { 'debounceAction' = putStrLn "Running action"
 --                  , 'debounceFreq' = 5000000 -- 5 seconds
---                  , 'debounceEdge' = 'trailingEdge' -- Trigger on the trailing edge
+--                  , 'debounceEdge' = 'DI.trailingEdge' -- Trigger on the trailing edge
 --                  }
 -- @
 --
@@ -32,8 +32,8 @@ module Control.Debounce
     , DI.debounceFreq
     , DI.debounceAction
     , DI.debounceEdge
-    , leadingEdge
-    , trailingEdge
+    , DI.leadingEdge
+    , DI.trailingEdge
       -- * Creation
     , mkDebounce
     ) where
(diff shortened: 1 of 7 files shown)


thomasjm:
Oh, I realized that this doesn't actually break the public API, so maybe the version should just bump to `0.1.6` instead of `0.2.0`?

snoyberg:
Yes, I believe that's the right version number. Making that change should allow CI to pass.

snoyberg:
Looks like there's a build failure with older versions of GHC:

```
auto-update         > /home/vsts/work/1/s/auto-update/test/Control/DebounceSpec.hs:9:8:
auto-update         >     Could not find module ‘Control.Monad.IO.Class’
auto-update         >     It is a member of the hidden package ‘transformers-0.4.2.0@trans_GZTjP9K5WFq01xC9BAGQpF’.
auto-update         >     Perhaps you need to add ‘transformers’ to the build-depends in your .cabal file.
auto-update         >     Use -v to see a list of the files searched for.
```

thomasjm:
The remaining failures seem to be related to `warp`

kazu-yamamoto:
Very strange. tls 1.4.1 surely exports `CompressionID`.

kazu-yamamoto:
@thomasjm Would you do `git rebase master` and `git push -f`?

thomasjm:
Slightly different failures now, but still `warp` spec and `warp-tls` build respectively

kazu-yamamoto:
- We can ignore `Linux cabal-8.2.2`. This happens sometime. I don't know why.
- For `Linux cabal-7.10.3`, it is strange that `tls-1.4.1` is used instead of `tls-1.5.0`.

thomasjm:
I notice that the successful test run on `master` right now also contains `tls-1.4.1` in the logs for the `Linux cabal-7.10.3` build...

kazu-yamamoto:
@snoyberg Though CI fails, we don't have to worry about it too much in this case. So, if you like, would you please merge this PR?

snoyberg:
Just an FYI, this is a larger PR than I anticipated, and I'm a bit swamped right now. I probably won't have a chance to review until next week. Sorry for the delay.

kazu-yamamoto:
CI passed on `master`. I'm sure that CI also passed on this PR if re-run.
Probably, `git rebase master` and `git push -f` is a good thing.

thomasjm:
Just rebased on `master` again.

Here's a few notes on what this PR does in case it helps:
* Implement the desired leading/trailing edge logic in `mkDebounce` and add a corresponding setting to `DebounceSettings`.
* For testability, move the core `mkDebounce` to a `.Internal` module and make it accept the `baton` and the delay function as arguments (so that tests can manipulate them).
* Comment out the existing tests (since there's a note saying they're unstable) and re-enable the test suite.
* Add some detailed tests of every `mkDebounce` configuration.

auto-update/Control/Debounce.hs
```diff
@@ -25,70 +26,34 @@
...
-      DebounceSettings
+      DI.DebounceSettings
     , defaultDebounceSettings
       -- * Accessors
-    , debounceFreq
-    , debounceAction
+    , DI.debounceFreq
+    , DI.debounceAction
+    , DI.debounceEdge
+    , DI.DebounceEdge(..)
```
snoyberg: Is it conceivable that we'd want to add more data constructors here in the future? If so, I'd rather export value-level identifiers for each constructor to avoid a breaking change if we add more options. However, I can't think of any reasonable additions myself.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+    , debounceEdge :: DebounceEdge
+    -- ^ Whether to perform the action on the leading edge or trailing edge of
+    -- the timeout.
+    --
+    -- Default: Leading.
+    --
+    -- @since 0.2.0
+    }
+
+data DebounceEdge =
```
snoyberg: Please add a Haddock comment on the data type itself including a `@since` comment.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+    -- Default: does nothing.
+    --
+    -- @since 0.1.2
+    , debounceEdge :: DebounceEdge
+    -- ^ Whether to perform the action on the leading edge or trailing edge of
+    -- the timeout.
+    --
+    -- Default: Leading.
+    --
+    -- @since 0.2.0
```
snoyberg: This version number needs to be updated.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+data DebounceEdge =
+  Leading
+  -- ^ The action will either be performed immediately, or after the current cooldown period has expired.
+  | Trailing
+  -- ^ The action will be performed at the end of the current cooldown period.
+  | LeadingAndTrailing
+  -- ^ The action is performed on the leading edge. It's also performed on the trailing edge if
```
snoyberg: While reviewing, I haven't looked at the implementation yet, and I'm left with a question. Let's say that the cooldown period is 2 seconds. We trigger an action, and 1 second later trigger it again. I can see two possible outcomes:

1. Wait for another 1 second, perform the action, and immediately enter the cooldown period again
2. Wait for another 1 second, perform the action, and then do not enter the cooldown period

The former seems more logical, but the latter seems to be implied. Either way, a bit more clarity in the docs would be helpful.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+  -- the debounced function was invoked again during the cooldown.
+  deriving (Show, Eq)
+
+mkDebounce :: MVar () -> (Int -> IO ()) -> DebounceSettings -> IO (IO ())
```
snoyberg: Not a big deal, but would you mind renaming this to something like `mkDebounceInternal`?

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Control.Debounce.Internal (
```
snoyberg: Please add a comment that this module is an unstable API.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+  | LeadingAndTrailing
+  -- ^ The action is performed on the leading edge. It's also performed on the trailing edge if
+  -- the debounced function was invoked again during the cooldown.
+  deriving (Show, Eq)
+
+mkDebounce :: MVar () -> (Int -> IO ()) -> DebounceSettings -> IO (IO ())
+mkDebounce baton delayFn (DebounceSettings freq action edge) = do
+    mask_ $ void $ forkIO $ forever $ do
+        takeMVar baton
+        if edge `elem` [Leading, LeadingAndTrailing] then ignoreExc action >> runDelay False baton
```
snoyberg: Personally, I'd prefer to use a case expression here. For one, it's slightly more efficient than usage of `elem`. More importantly, it will lead to a compiler warning if more constructors are added in the future.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+        else runDelay True baton
+
+    return $ void $ tryPutMVar baton ()
+
+  where runDelay hasDelayedInitialActivation baton = do
```
snoyberg: It looks like `baton` doesn't need to be passed in here, it can be inherited from the top level function.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,84 @@
...
+            delayFn freq
+
+            case edge of
+                Trailing -> do
+                    firedDuringInterval <- isJust <$> tryTakeMVar baton
+                    when (firedDuringInterval || hasDelayedInitialActivation) $ ignoreExc action
```
snoyberg: This condition doesn't seem to make sense. `Trailing` here means that `hasDelayedInitialActivation` is always `True`, if I'm reading the code above correctly. It seems like what you actually want here is:

```haskell
void $ tryTakeMVar baton -- make sure it's empty
ignoreExc action
```

Or arguably, you want to empty the baton _after_ performing the action by switching those two statements.

auto-update/Control/Debounce.hs
thomasjm: Well, I was taking inspiration from Lodash's API. Looking at Lodash's manual, this seems to have been how they do it for the last 7 years so my thinking was that this API is tried and true. But happy to change to the identifiers if you prefer.

snoyberg: That seems fine to me then, we can leave as-is.

auto-update/Control/Debounce/Internal.hs
thomasjm: Oh you're right, that does simplify things.

I think if there's a new event concurrent with the processing of the action, it's better to err on the side of doing the action again after the next cooldown. That's also the approach taken with `LeadingAndTrailing`

auto-update/Control/Debounce/Internal.hs
thomasjm: The latter is how it works at the moment. So if another trigger happens at 3 seconds, then the action will happen immediately. 

The former is actually identical to the current `Leading` semantics, no?

Actually, I just played with Lodash in the console and found that for their `Leading` version they do *not* re-enter the cooldown. So if a new trigger happens at 3 seconds they perform the action immediately. This is making me wonder if there's even a difference between Lodash `Leading` and `LeadingAndTrailing`. I guess there's a subtle difference when it comes to which Javascript tick the action happens on, but that doesn't apply to us.

So really, the difference between `Leading` and `LeadingAndTrailing` for us is that `Leading` does re-enter the cooldown phase after a trigger during the previous cooldown, while `LeadingAndTrailing` does not. I guess this is a useful distinction to make--I'll try to clarify it in the docs.

thomasjm:
Thanks for the review! I resolved everything except for one comment about the details of the semantics. 

For that one I updated the docs to be as clear as possible but it would be good if you could take another look. After thinking about it I'm wondering if the distinction between `Leading` and `LeadingAndTrailing` is useful or if perhaps `LeadingAndTrailing` should be removed.

thomasjm:
(If `LeadingAndTrailing` were removed then you would be guaranteed that the action will never be invoked more frequently than once every `debounceFreq` microseconds, which seems nice.)

kazu-yamamoto:
I've run CI again resulting in green!

thomasjm:
After consideration I decided to remove `LeadingAndTrailing`. The distinction with `Leading` is quite small and the utility is questionable.

I converted the API to export value-level identifiers instead, so we can re-add it later if we decide we want it.

auto-update/Control/Debounce/Internal.hs
```diff
@@ -0,0 +1,104 @@
...
+-- @since 0.1.6
+trailingEdge :: DebounceEdge
+trailingEdge = Trailing
+
+mkDebounceInternal :: MVar () -> (Int -> IO ()) -> DebounceSettings -> IO (IO ())
+mkDebounceInternal baton delayFn (DebounceSettings freq action edge) = do
+    mask_ $ void $ forkIO $ forever $ do
+        takeMVar baton
+        case edge of
+          Leading -> ignoreExc action >> runDelay
```
snoyberg: The code looks correct. But I think it might end up being more clear by inlining the `runDelay` call.

thomasjm: Yep, done

thomasjm:
Okay, all feedback is addressed now. Thanks @snoyberg !

snoyberg:
Thanks!

PR #757, opened 2019-06-24
Supply response size to custom loggers when available.

ygale:

(N/A) Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
(N/A) Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,6 +1,6 @@
 # Changelog for wai-extra
 
-## 3.0.26
+## 3.0.26.1
 
 * When available, supply the response size to custom loggers
   [#757](https://github.com/yesodweb/wai/pull/757)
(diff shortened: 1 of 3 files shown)


ygale:
We were using the standard Apache logger, and we received a request from DevOps to supply an additional field. So we changed to a custom logger with details. But then the response size field vanished, with no way to restore it other than re-implementing the logger middleware.

ygale:
@snoyberg From Yuval, the dev who is working on this: "Wow that was fast !"

It means we can already use it in the current sprint. It came through even before we finished implementing the work-around hack. This is really, really appreciated. Thanks!

snoyberg:
No problem. And please pass on thanks for a good PR to Yuval ;)

PR #759, opened 2019-06-28
shutdown instead of close.

### LINKED ISSUE #673: Investigate how warp handles HTTP/2 GOAWAY
Hi @kazu-yamamoto,

yesterday I found a problem in nginx's implementation of the HTTP/2 `GOAWAY` feature. It is very easy to get that feature wrong. I have written up and linked the details on https://trac.nginx.org/nginx/ticket/1250#comment:4.

Would you be able to tell me if warp handles this correctly currently, or if that's not clear yet, how it implements sending `GOAWAY` or where the bit that handles this is in the code?

Thanks!

kazu-yamamoto:
This patch lets Warp to use `shutdown` instead of `close`. 

Background: when Warp sends GOAWAY in HTTP/2, `close` is called immediately. The kernel of the browser sends TCP ACK regarding to GOAWAY, but the socket of the server is already closed. So, TCP RST is sent back from the server to the client.

"UNIX Network Programming" suggests to use `shutdown` with `SHUT_WR`. This makes sure that the TCP ACK is processed in the server properly and the socket is closed in the proper timing. Note that the combination of `Linger` and `close` does not help.

- I confirmed that this fixes two bugs which `h2spec` finds.
- This should fix #673.
- I confirmed that sockets are properly closed with `lsof -p` in the field.

Cc: @nh2

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -23,6 +23,9 @@ import Foreign.C.Error (Errno(..), eCONNABORTED)
 import GHC.IO.Exception (IOException(..))
 import qualified Network.HTTP2 as H2
 import Network.Socket (Socket, close, accept, withSocketsDo, SockAddr(SockAddrInet, SockAddrInet6), setSocketOption, SocketOption(..))
+#if MIN_VERSION_network(3,1,1)
+import Network.Socket (gracefulClose)
+#endif
 import qualified Network.Socket.ByteString as Sock
 import Network.Wai
 import Network.Wai.Internal (ResponseReceived (ResponseReceived))
@@ -64,7 +67,11 @@ socketConnection s = do
         connSendMany = Sock.sendMany s
       , connSendAll = sendall
       , connSendFile = sendFile s writeBuf bufferSize sendall
+#if MIN_VERSION_network(3,1,1)
+      , connClose = gracefulClose s 5000
+#else
       , connClose = close s
+#endif
       , connFree = freeBuffer writeBuf
       , connRecv = receive s bufferPool
       , connRecvBuf = receiveBuf s
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -53,6 +53,9 @@ import qualified Data.IORef as I
 import Data.Streaming.Network (bindPortTCP, safeRecv)
 import Data.Typeable (Typeable)
 import Network.Socket (Socket, close, withSocketsDo, SockAddr, accept)
+#if MIN_VERSION_network(3,1,1)
+import Network.Socket (gracefulClose)
+#endif
 import Network.Socket.ByteString (sendAll)
 import qualified Network.TLS as TLS
 import qualified Crypto.PubKey.DH as DH
@@ -355,7 +358,11 @@ httpOverTls TLSSettings{..} s bs0 params = do
   where
     backend recvN = TLS.Backend {
         TLS.backendFlush = return ()
+#if MIN_VERSION_network(3,1,1)
+      , TLS.backendClose = gracefulClose s 5000
+#else
       , TLS.backendClose = close s
+#endif
       , TLS.backendSend  = sendAll' s
       , TLS.backendRecv  = recvN
       }


nh2:
> I confirmed that this fixes two bugs which `h2spec` finds.

@kazu-yamamoto Can you post how `h2spec` is invoked and the details of the problems it finds?

It'd be very interested to learn how that works.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -64,7 +64,7 @@ socketConnection s = do
         connSendMany = Sock.sendMany s
       , connSendAll = sendall
       , connSendFile = sendFile s writeBuf bufferSize sendall
-      , connClose = close s
+      , connClose = shutdown s ShutdownSend `E.catch` \(SomeException _) -> return ()
```
nh2: 1. Why is it sufficient to call `shutdown` but never `close` afterwards?
2. What if `shutdown()` returns `ENOTCONN`, do we leak the socket?
3. Is the catch-all exception handler good, or should it be more specific, so that we can observe bugs in the code? For example, should it not be at max some `IOError`? Should it perhaps be even specific to some `errno` we can expect, like `ENOTCONN`?
4. Of the 4 steps I list in https://github.com/nodejs/node/issues/12339#issuecomment-368124078, this does only 2 steps. Don't we have to
    * Keep `read()`ing from the socket until the empty string is read
    * Call `close()` on the socket
5. What about the case of misbehaving clients that never close their side? Will they be able to drain resources from the server by keeping ports open? Should there be some timeout/setting/mechanism to `close()` if the other side doesn't `close()` after some expected duration?

Even if these are fine, I think it deverses a code comment why they are fine.

kazu-yamamoto:
```
% h2spec http2/5.4 http2/7
  5. Streams and Multiplexing
    5.4. Error Handling
      5.4.1. Connection Error Handling
        × 1: Sends an invalid PING frame for connection close
          -> The endpoint MUST close the TCP connection
             Expected: Connection closed
               Actual: Error: read tcp 127.0.0.1:56529->127.0.0.1:80: read: connection reset by peer

  7. Error Codes
    × 1: Sends a GOAWAY frame with unknown error code
      -> The endpoint MUST NOT trigger any special behavior.
         Expected: Connection closed
                   PING Frame (length:8, flags:0x01, stream_id:0, opaque_data:h2spec)
           Actual: Error: read tcp 127.0.0.1:56578->127.0.0.1:80: read: connection reset by peer
```

kazu-yamamoto:
As you can see, `h2spec` expects connection close in both cases but receives TCP RST.

kazu-yamamoto:
It is already Saturday in Japan. So, I will answer other stuffs in Monday.

kazu-yamamoto:
@snoyberg I guess that it's better for WarpTLS to use `shutdown` instead of `close`. I will also take care of this in next Monday.

snoyberg:
I'd defer to @nh2 on my review, his points subsume what I noticed. The biggest question I was left with was why not do both shutdown _and_ close.

snoyberg:
Oh, also: making the same change to warp-tls would make sense to me.

kazu-yamamoto:
Quick reply: `shutdown` is a graceful `close`. There are alternatives. You don't have to call both. Rather, we must not call both for graceful closing.

nh2:
> You don't have to call both.

Is that true?

It contradicts my understanding of `man 2 shutdown` and `man 2 close` and [this StackOverflow answer](https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket):

>  However, close is the way to actually destroy a socket.
>
> With shutdown, you will still be able to receive pending data the peer already sent.

The second sentence suggests that you can continue to use the socket FD in some ways after calling `shutdown()`; this implies that the socket is still allocated in the kernel.

That is in line with what I wrote in the link and text posted in my detail comment above: https://github.com/yesodweb/wai/pull/759#discussion_r298548856

kazu-yamamoto:
@nh2 You are right. I considered only normal cases. I will consider error cases.

kazu-yamamoto:
@nh2 I rewrote the code and did `push -f`.

I'm not 100% convinced yet. What happens if TCP FIN is not returned from the client while receiving data? Does TCP timeout surely happen?

kazu-yamamoto:
- I guess that we cannot use `SO_RCVTIMEO` which results in `EWOULDBLOCK` on timeout. Sockets in Haskell are non-blocking where `EWOULDBLOCK` is also returned in many cases. We cannot distinguish them. What we can do is to set sockets back to blocking, set `SO_RCVTIMEO` and call `recv(2)` (not Haskell's `recv`). All parts are not provided from the `network` library.
- `SO_KEEPALIVE` would be another option. But the timeout is too long in some OSes and the timeout value cannot change for each socket.

kazu-yamamoto:
OK. As one of maintainers of the `network` library, I decided to implement `gracefulClose` in `network`.

https://github.com/haskell/network/pull/417

kazu-yamamoto:
I did `push -f` to use `gracefulClose` provided in the network lib.
We should wait for https://github.com/haskell/network/pull/417

kazu-yamamoto:
@snoyberg It appeared that this PR makes the tests very slow. This is probably because `http-client` does not sent TCP FIN *immediately* after `Manager` is not in use. `closeManager` is now obsoleted. Are there any ways to close all connections under `Manager` explicitly and immediately?

snoyberg:
There's nothing currently. If I added something on `master` to close out all open connections, would you be able to test that it speeds up the tests?

kazu-yamamoto:
Yes, of course!

snoyberg:
Actually, there might be a simpler solution: how about setting `managerConnCount` to 0? Or maybe `managerIdleConnectionCount`?

kazu-yamamoto:
@snoyberg It works. Thanks!

But testing is still slow for other reasons.  I try to understand where the slow down comes from.

kazu-yamamoto:
Since the PR of `network` has been merged, I merged this PR.
Also, new versions of `warp` and `warp-tls` have been released.

PR #766, opened 2019-07-29
Add openApplication and variants

mitchellwrosen:
This patch adds explicit `open`/`close` functions for testing applications. My use case is acquiring a `warp` server in a `tasty` test. It unfortunately does not seem to work at all with "with-style" resource acquiring functions:

```haskell
withResource :: IO a -> (a -> IO ()) -> (IO a -> TestTree) -> TestTree
```

I had to lift a considerable amount of guts into the returned type (socket, wait-for-server action, server async), so I stopped short of writing proper documentation and bumping the cabal version number. Please let me know how/if you'd like to proceed, thanks!

### DIFF
--- a/warp/Network/Wai/Handler/Warp/WithApplication.hs
+++ b/warp/Network/Wai/Handler/Warp/WithApplication.hs
@@ -2,8 +2,13 @@
 module Network.Wai.Handler.Warp.WithApplication (
   withApplication,
   withApplicationSettings,
+  openApplication,
+  openApplicationSettings,
+  closeApplication,
   testWithApplication,
   testWithApplicationSettings,
+  testOpenApplication,
+  testOpenApplicationSettings,
   openFreePort,
   withFreePort,
 ) where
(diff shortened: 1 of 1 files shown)


warp/Network/Wai/Handler/Warp/WithApplication.hs
```diff
@@ -27,26 +32,36 @@ import           Network.Wai.Handler.Warp.Types
...
-      Left () -> throwIO $ ErrorCall "Unexpected: runSettingsSocket exited"
-      Right x -> return x
+  (port, sock) <- openFreePort
+  started <- mkWaiter
+  let settings =
+        settings' {
+          settingsBeforeMainLoop
+            = notify started () >> settingsBeforeMainLoop settings'
+        }
+  serverAsync <- async (runSettingsSocket settings sock app)
```
snoyberg: This will result in a fairly significant change in behavior. The entire `openApplicationSettings` bit is now being run inside a masked exception state, meaning the `runSettingsSocket` call is _also_ in a masked exception state, which is almost certainly not what we want.

mitchellwrosen: Excellent catch. `asyncWithUnmask` is what I should have used.

mitchellwrosen:
I'm not sure I need this anymore, sorry for the spam!

Issue #765, opened 2019-07-26
wai-extra: new EventSource builder to enable recursive emitters?


xaviershay:
Let's say I'm writing an eventsource handler that counts upwards. To do this with current API, I _think_ the best way is to use an `IORef` with `eventSourceAppIO`. WDYT about a method like this as an alternative:

```
eventStreamIO :: ((Network.Wai.EventSource.ServerEvent -> IO()) -> IO ()) -> Application
eventStreamIO handler _ respond = respond $ responseStream status200 [(hContentType, "text/event-stream")] (streamer handler)
  where
    streamer handler sendChunk flush = do
      handler emit

      where
        emit :: Network.Wai.EventSource.ServerEvent -> IO ()
        emit event = do
           case eventToBuilder event of
             Nothing -> return ()
             Just b  -> sendChunk b >> flush
```

The type is a bit weird, but it lets me do:

```
myApp :: Application
myApp = do
  eventStreamIO (handle 0)

  where
    handle :: Int -> (Network.Wai.EventSource.ServerEvent -> IO ()) -> IO ()
    handle counter emit = do
      emit $ Network.Wai.EventSource.ServerEvent
               (Just "snapshot")
               Nothing
               [ Data.Binary.Builder.fromLazyByteString
                   (Data.Aeson.encode $ counter)
               ]

      Control.Concurrent.threadDelay 1000000

      handle (counter + 1) emit
```

... which I think is quite nice and feels more like how I would write this sort of loop normally?

I'm happy to contribute a patch (with better naming and docs) if you think this seems reasonable. Quite possible I'm doing it completely wrong though! Am pretty new to this library.

xaviershay:
There's some prior art in `wai-websockets`, which isn't exactly the same interface but allows the same approach: http://hackage.haskell.org/package/wai-websockets-3.0.1.2/docs/Network-Wai-Handler-WebSockets.html (recursively call `sendTextData`)

snoyberg:
I'm OK with something like this in principle, no strong opinions right now.

xaviershay:
Thanks, I'll try putting together a proper PR!

PR #767, opened 2019-08-03
EventSource builder to enable recursive emitters

xaviershay:
Follow up to #765.

I am not tied to the name of this method. "Raw" seemed most inline with the current naming scheme (`IO` and `Chan` named after the type of their parameter) - in theory you could build them on top of this method, though I haven't done that in this PR since don't have a good setup to benchmark against any potential perf. regression and wanted to keep low risk. Also considered `eventSourceAppEmit` or `Emitter` as names.

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/EventSource.hs
+++ b/wai-extra/Network/Wai/EventSource.hs
@@ -5,6 +5,8 @@
     headers:
 
     > [ ("X-Accel-Buffering", "no"), ("Cache-Control", "no-cache")]
+
+    There is a small example using these functions in the @example@ directory.
 -}
 module Network.Wai.EventSource (
     ServerEvent(..),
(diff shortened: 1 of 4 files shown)


xaviershay:
Not sure build failure is related:

```
basement-0.0.11 failed during the building phase. The exception was:
ExitFailure 1
```

wai-extra/Network/Wai/EventSource.hs
```diff
@@ -39,3 +41,16 @@ eventSourceAppIO src _ sendResponse =
...
+-- @since 3.0.28
+eventStreamAppRaw :: ((ServerEvent -> IO()) -> IO ()) -> Application
+eventStreamAppRaw handler _ sendResponse =
+  sendResponse $ responseStream
+      status200
+      [(hContentType, "text/event-stream")]
+      $ \sendChunk flush -> handler $ \event ->
+           case eventToBuilder event of
+               Nothing -> return ()
+               Just b  -> sendChunk b >> flush
```
snoyberg: It may be more general to provide two different callback action: one to `sendChunk`, and the other to `flush`. This would mirror other APIs we already have.

xaviershay: good point, will amend

xaviershay:
Updated

snoyberg:
Thanks!

Issue #746, opened 2019-05-19
setOnException not called in warp-tls


k-bx:
I see a difference in behavior between warp and warp-tls. The latter doesn't call my setOnException handler.

I've made a fully reproducible repo (run with `stack run`) where you can comment/uncomment to switch between http/https and see the difference: https://github.com/k-bx/servant-custommonad-logging/blob/master/waitls-onexception-bug/src/Main.hs

```
$ curl -i localhost:8000/api/error-out
```

HTTP Gives:

```
Allocating scarce resource
Cleaning up
Error: caught an exception: Erroring out
CallStack (from HasCallStack):
  error, called at src/Main.hs:33:12 in main:Main. Call stack: CallStack (from HasCallStack):
  onExceptionAct, called at src/Main.hs:37:29 in main:Main
```

Switchging to https:

```
$ curl --insecure -i https://localhost:8000/api/error-out
```

Gives no output from the app.

kazu-yamamoto:
I'm very sorry. I did not notice this and the code changed very much.
Would you give a look at the current code and think how to integrate?

k-bx:
Sure, I can check it out on latest wai/warp/http2. Just tried -- getting this build error:

```
http2                 > Configuring http2-2.0.0...
http2                 > build
http2                 > Preprocessing library for http2-2.0.0..
http2                 > Building library for http2-2.0.0..
http2                 > [ 1 of 45] Compiling Imports
http2                 > [ 2 of 45] Compiling Network.HPACK.Builder
http2                 > [ 3 of 45] Compiling Network.HPACK.HeaderBlock.Integer
http2                 >
http2                 > /tmp/stack31729/http2-2.0.0/Network/HPACK/HeaderBlock/Integer.hs:40:11: error:
http2                 >     Not in scope: type constructor or class ‘WriteBuffer’
http2                 >    |
http2                 > 40 | encode :: WriteBuffer -> (Word8 -> Word8) -> Int -> Int -> IO ()
http2                 >    |           ^^^^^^^^^^^
http2                 >
http2                 > /tmp/stack31729/http2-2.0.0/Network/HPACK/HeaderBlock/Integer.hs:85:27: error:
http2                 >     Not in scope: type constructor or class ‘ReadBuffer’
http2                 >    |
http2                 > 85 | decode :: Int -> Word8 -> ReadBuffer -> IO Int
http2                 >    |                           ^^^^^^^^^^
http2                 >
```

[...]

PR #770, opened 2019-09-28
746 setonexception not called

k-bx:
This fixes https://github.com/yesodweb/wai/issues/746

Checked with the repo https://github.com/k-bx/servant-custommonad-logging by doing:

```
$ cd waitls-onexception-bug
$ stack build && stack run
```

and running:

```
$ curl --insecure -i https://localhost:8000/api/error-out
HTTP/2 500                       
date: Sat, 28 Sep 2019 09:48:43 GMT
server: Warp/3.3.2                                            
content-type: text/plain; charset=utf-8          
                         
Something went wrong%                                                
```

Now seeing output:

```
Allocating scarce resource
Cleaning up
Error: caught an exception: Erroring out
  error, called at src/Main.hs:33:12 in main:Main. Call stack: CallStack (from HasCallStack):
  onExceptionAct, called at src/Main.hs:37:29 in main:Main
```

After submitting your PR:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2.hs
@@ -37,7 +37,7 @@ http2 conn ii addr settings readN app =
 
     http2server h2req aux response = do
         req <- toWAIRequest h2req aux
-        ResponseReceived <- app req $ \rsp -> do
+        eResponseReceived <- E.try $ app req $ \rsp -> do
             h2rsp <- fromResponse settings ii req rsp
             pps <- fromPushPromises ii req
             ex <- E.try $ response h2rsp pps
@@ -50,8 +50,23 @@ http2 conn ii addr settings readN app =
                 | Just E.ThreadKilled  <- E.fromException e -> return ()
                 -- killed by the local timeout manager
                 | Just T.TimeoutThread <- E.fromException e -> return ()
-                | otherwise -> S.settingsOnException settings (Just req) e
+                | otherwise -> do
+                    S.settingsOnException settings (Just req) e
             return ResponseReceived
+        case eResponseReceived of
+          Right ResponseReceived -> return ()
+          Left e@(E.SomeException _)
+            -- killed by the local worker manager
+            | Just E.ThreadKilled  <- E.fromException e -> return ()
+            -- killed by the local timeout manager
+            | Just T.TimeoutThread <- E.fromException e -> return ()
+            | otherwise -> do
+                S.settingsOnException settings (Just req) e
+                pps' <- fromPushPromises ii req
+                h2rsp' <- fromResponse settings ii req (S.settingsOnExceptionResponse settings e)
+                response h2rsp' pps'
+                logResponse h2rsp' req
+                mapM_ (logPushPromise req) pps'
         return ()
 
     toWAIRequest h2req aux = toRequest ii settings addr hdr bdylen bdy th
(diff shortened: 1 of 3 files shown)


k-bx:
Doesn't seem like the errors are mine, marking as "CI passes"

warp/Network/Wai/Handler/Warp/HTTP2.hs
```diff
@@ -50,8 +50,23 @@ http2 conn ii addr settings readN app =
                 | Just E.ThreadKilled  <- E.fromException e -> return ()
                 -- killed by the local timeout manager
                 | Just T.TimeoutThread <- E.fromException e -> return ()
-                | otherwise -> S.settingsOnException settings (Just req) e
+                | otherwise -> do
+                    S.settingsOnException settings (Just req) e
             return ResponseReceived
+        case eResponseReceived of
```
snoyberg: This essentially duplicates the `settingsOnException` logic: once for exceptions from the response, and once from the application itself. However, I'm not following the motivation for this: why do we need to `try` twice and handle these cases differently? At the very least, some comments explaining would be great.

warp/Network/Wai/Handler/Warp/HTTP2.hs
```diff
@@ -50,8 +50,23 @@ http2 conn ii addr settings readN app =
...
+          Right ResponseReceived -> return ()
+          Left e@(E.SomeException _)
+            -- killed by the local worker manager
+            | Just E.ThreadKilled  <- E.fromException e -> return ()
+            -- killed by the local timeout manager
+            | Just T.TimeoutThread <- E.fromException e -> return ()
+            | otherwise -> do
```
snoyberg: @kazu-yamamoto would you be able to review this?

warp/Network/Wai/Handler/Warp/HTTP2.hs
k-bx: > However, I'm not following the motivation for this: why do we need to try twice and handle these cases differently? At the very least, some comments explaining would be great.

@snoyberg the motivation comes not from analyzing the code but from a real-world servant app that doesn't trigger `settingsOnException` when actions are throwing an exception. I've made a minimal reproduction repository with full instructions present in this PR.

Now, to your question as to why do we need to catch anything inside the action itself – I don't know, but since I didn't have time to be able to write a piece of code that would be able to go through that branch of code, I decided to keep the behavior as it is currently. Definitely feels like it could be eliminated, but I am not 100% confident on that. Any comments from @kazu-yamamoto would be very much appreciated!

snoyberg: Separate from issues within `warp` or `warp-tls`, this sounds very reminiscent of https://github.com/haskell-servant/servant/issues/309.

k-bx: Yes, this was exactly the problem I started seeing when I've switched from default servant to a custom `AppM = ReaderT Env IO` thing to use it together with `UnliftIO` and rio stuff

> setOnException doesn't work for throwError in a Handler. It seems that if you throw an exception in IO (ie. liftIO $ throw someException), that gets handled by the default exception handler in warp. However, throwing a ServantErr doesn't get caught by warp. Servant responds with a 500 error, but nothing gets logged.

And this PR should fix it.

kazu-yamamoto:
@k-bx First of all, why do you bump the version of `wai`? I believe this fix is closed in `warp`.

kazu-yamamoto:
Closing.

PR #771, opened 2019-10-03
Calling setOnException in HTTP/2 (#746 and #770)

### LINKED ISSUE #746: setOnException not called in warp-tls

kazu-yamamoto:
This should fix #746 and be simpler than #770. 

Cc: @k-bx

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2.hs
@@ -6,6 +6,7 @@
 
 module Network.Wai.Handler.Warp.HTTP2 (http2) where
 
+import qualified Data.IORef as I
 import qualified Control.Exception as E
 import qualified Network.HTTP2.Server as H2
 import Network.Socket (SockAddr)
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
Rebased to delete unnecessary changes, recorded the credit for @k-bx and merged.

kazu-yamamoto:

k-bx:
Thank you! 🙏

PR #776, opened 2019-10-31
Add a way to hide passwords (fixes #774)

### LINKED ISSUE #774: Hiding passwords from RequestLogger
I'd like to use RequestLogger in my app because it seems like the simplest logging solution, but I need to be able to filter passwords from the logs. I see that I can create a `OutputFormatterWithDetails` but it's pretty annoying to have to import a bunch of low-level libraries to stitch together a custom logging string, when I really like the `Detailed` format except that it logs passwords. I have a fork where I added a param to `Detailed` that lets you add a function `Param -> Param` so you can remove whatever you want from the params, but I'm not sure if there's an easier way that I'm missing. If that solution would be acceptable to y'all, I'm happy to open a PR.

emhoracek:

- [ x ] Bumped the version number
- [ x ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [ x ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RequestLogger.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger.hs
@@ -33,7 +33,7 @@ import Network.Wai
   )
 import System.Log.FastLogger
 import Network.HTTP.Types as H
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, isJust)
 import Data.Monoid (mconcat, (<>))
 import Data.Time (getCurrentTime, diffUTCTime, NominalDiffTime)
 import Network.Wai.Parse (sinkRequestBody, lbsBackEnd, fileName, Param, File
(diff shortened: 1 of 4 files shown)


wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -50,9 +50,21 @@ import Network.Wai.Middleware.RequestLogger.Internal
...
+-- information (such as passwords) from the logs. If the second parameter is `Just`, then POST
+-- bodies are also hidden. For example:
+-- > myformat = Detailed True (Just hidePasswords)
+-- >   where hidePasswords p@(k,v) = if k = "password" then (k, "***REDACTED***") else p
+--
+-- Default is `Detailed True Nothing`
 data OutputFormat
   = Apache IPAddrSource
   | Detailed Bool -- ^ use colors?
+             (Maybe (Param -> Param)) -- ^ @since 3.1.0
```
snoyberg: If we instead had a `DetailedModParams` or similar data constructor, would we be able to keep backwards compat? We could even have an extensible `DetailedSettings` constructor so new additions in the future would be a trivial addition to that type.

Vlix:
Now with the detailed settings from #826 this issue can also be closed, right?

Vlix:
#826 seems to provide what this PR aims to provide in functionality, so I'm going to close it.

PR #783, opened 2019-12-16
Getting client certificate

### LINKED ISSUE #482: Support not only requesting client certificates but inspecting them once a connection established
There is a flag in `TLSSettings`, `tlsWantClientCert`, which pokes `tls` to ask for client certs after receiving the `ClientHello` however there are two and a half problems:
1. There's no way to get the `CertificateChain` returned by the client after the demand.
2. The TLS library by default (due to the default `onClientCertificate` hook) rejects the client certificates.
3. There's no way to hook into `onClientCertificate` if you want to be selective about which client certs you accept.

I've crafted a patch which addresses 1 and 2, though because if how it exposes the certificate chain via `Transport` it entails a new dependency on the `x509` package by the `warp` package for that type. I'm not sure if that's acceptable or not, so take the patch as an example if it's not acceptable for whatever reason.

My patch also exposes the `runServeTLS` and `runServeTLSSocket` functions which I'm guessing should have been exposed in the first place. This is in order to receive the `Transport`.

Branch: https://github.com/Dridus/wai/tree/client-certificates
Commit fixing 1 and 2: https://github.com/Dridus/wai/commit/c28f490eb3e2658a292077c28e67421a2671fad7
Commit exposing `runServeTLS` and `runServeTLSSocket`: https://github.com/Dridus/wai/commit/2dd2ffbbec593bb7f31662545f05da527d674038

Commit (on separate branch) with test app which exercises the new behavior: https://github.com/Dridus/wai/commit/36ef66f28a8072ed815ff1ff8fe8c0311cdcd788

kazu-yamamoto:
This should fix #482.

See also:
- https://github.com/vincenthz/hs-tls/pull/405
- https://github.com/vincenthz/hs-tls/pull/407


warp/Network/Wai/Handler/Warp.hs
```diff
@@ -491,3 +493,10 @@ setGracefulCloseTimeout2 x y = y { settingsGracefulCloseTimeout2 = x }
 -- Since 3.3.5
 getGracefulCloseTimeout2 :: Settings -> Int
 getGracefulCloseTimeout2 = settingsGracefulCloseTimeout2
+
+-- | Getting information of client certificate.
+--
+-- Since 3.3.5
+clientCertificate :: Request -> Maybe CertificateChain
+clientCertificate = fromMaybe (throw (userError "getClientCertificate")) . Vault.lookup getClientCertificateKey . vault
```
snoyberg: What's the motivation for using a bottom value here instead of propagating the `Maybe`? If we need to distinguish different cases, I'd rather a specific sum type to represent the different ways things can fail.

kazu-yamamoto: Oh. It's from copy-and-paste. Does the following code make sense?

```haskell
clientCertificate :: Request -> Maybe CertificateChain
clientCertificate = join . Vault.lookup getClientCertificateKey . vault
```

snoyberg: Yup, seems reasonable.

kazu-yamamoto: Thanks. I pushed this. Can we merge this PR now?

snoyberg:
I think the version bump has broken all of CI. Do you want me to take a crack at fixing that before we merge?

kazu-yamamoto:
No rush. Now I think we should wait for getting tls 1.5.3 released.

kazu-yamamoto:
New versions of `tls` and `tls-session-manager` have been released.
@snoyberg Would you try to fix CI?

snoyberg:
@kazu-yamamoto did the package that provides the `Data.X509` module change? I'm seeing this:

```
warp               > /home/vsts/work/1/s/warp/Network/Wai/Handler/Warp/Types.hs:11:1: error:
warp               >     Could not load module ‘Data.X509’
warp               >     It is a member of the hidden package ‘x509-1.7.5’.
warp               >     Perhaps you need to add ‘x509’ to the build-depends in your .cabal file.
warp               >     Use -v to see a list of the files searched for.
warp               >    |
```

I've pushed my attempted CI fixes to this branch.

kazu-yamamoto:
I compared v1.5.2 and v1.5.3 but there is *no* change for exporting. Very strange.

@ocheron Any thoughts?

kazu-yamamoto:
@snoyberg f8282948d891fbf6a9e6dc22ca1339b4a5275817 added `x509` as dependency. That's why?

ocheron:
Apparently it's needed in the benchmark module too but wasn't added.

kazu-yamamoto:
@ocheron Wow! Nice catch! Thanks. Hoping that CI will recover.

kazu-yamamoto:
@snoyberg Should we also add `tls-session-manager-0.0.4` to `extra-deps:`?

snoyberg:
Yes, good call. I've pushed a bunch of updates on the branch, including updates to the stack.yaml files and a lower bound in warp-tls. I've also improved the build matrix a bit by dropping 8.0.2 (no longer supported) and adding in 8.8.1 testing.

snoyberg:
@kazu-yamamoto woohoo, looks like CI passed!

kazu-yamamoto:
@snoyberg Thanks! I have merged this PR.

kazu-yamamoto:
I have released new versions of warp and warp-tls.

PR #787, opened 2020-01-08
The time manger does not close sockets to prevent "Bad file descriptor"

### LINKED ISSUE #739: Exception in 'threadWait' after a while
I'm getting `threadWait: invalid argument (Bad file descriptor)` after a while when an endpoint has been triggered.

### Example code
```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE TypeOperators #-}

module Main where

import           Data.Aeson (ToJSON)
import           Data.Text as T
import           GHC.Generics (Generic)
import           Network.Wai.Handler.Warp (run)
import           Servant (Application, Proxy(..), Server, serve)
import           Servant.API ((:>), Get, JSON)

type ModuleAPI = "modules" :> Get '[JSON] [Module]

data Module = Module { moduleName :: T.Text }
  deriving (Eq, Show, Generic)

instance ToJSON Module

mkModule :: T.Text -> Module
mkModule name = Module { moduleName = name }

modules :: [Module]
modules = [ mkModule "module#1" ]

server :: Server ModuleAPI
server = return modules

moduleAPI :: Proxy ModuleAPI
moduleAPI = Proxy

app :: Application
app = serve moduleAPI server

main :: IO ()
main = run 8080 app
```

### Steps to reproduce
1. Compile with profiling options
2. Start the application with `+RTS -xc`
3. Make a request against `http://localhost:8080/modules`
4. Wait for the exception stack trace to be printed to the console

### StackTrace
```
*** Exception (reporting due to +RTS -xc): (base:GHC.Exception.Type.SomeException), stack trace:
  Network.Wai.Handler.Warp.Run.serveConnection.\,
  called from Network.Wai.Handler.Warp.Recv.receiveloop,
  called from Network.Wai.Handler.Warp.Recv.receive.\,
  called from Network.Wai.Handler.Warp.Recv.receive,
  called from Network.Wai.Handler.Warp.Run.socketConnection,
  called from Network.Wai.Handler.Warp.Run.runSettingsSocket.getConn,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnection.getConnMaker,
  called from Network.Wai.Handler.Warp.Run.acceptConnection.acceptNewConnection,
  called from Network.Wai.Handler.Warp.Run.wrappedRecv,
  called from Network.Wai.Handler.Warp.Types.mkSource,
  called from Network.Wai.Handler.Warp.Types.readSource,
  called from Network.Wai.Handler.Warp.Request.headerLines,
  called from Network.Wai.Handler.Warp.Request.recvRequest,
  called from Network.Wai.Handler.Warp.Run.serveConnection.http1,
  called from Network.Wai.Handler.Warp.Run.serveConnection,
  called from Network.Wai.Handler.Warp.Run.fork.serve.\.\,
  called from Network.Wai.Handler.Warp.Run.fork.serve.\,
  called from Network.Wai.Handler.Warp.Run.fork.serve,
  called from Network.Wai.Handler.Warp.Run.fork.\.\,
  called from Network.Wai.Handler.Warp.Run.fork.withClosedRef,
  called from Network.Wai.Handler.Warp.Run.fork.\,
  called from Network.Wai.Handler.Warp.Settings.defaultSettings,
  called from Network.Wai.Handler.Warp.Settings.settingsFork,
  called from Network.Wai.Handler.Warp.Run.fork,
  called from Network.Wai.Handler.Warp.Run.acceptConnection.acceptLoop,
  called from Network.Wai.Handler.Warp.Run.acceptConnection,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withII0.\.\.\.\,
  called from Network.Wai.Handler.Warp.FileInfoCache.withFileInfoCache,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withII0.\.\.\,
  called from Network.Wai.Handler.Warp.FdCache.withFdCache,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withII0.\.\,
  called from Network.Wai.Handler.Warp.Date.withDateCache,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withII0.\,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withTimeoutManager,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure.withII0,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMakerSecure,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnectionMaker,
  called from Network.Wai.Handler.Warp.Run.runSettingsConnection,
  called from Network.Wai.Handler.Warp.Run.runSettingsSocket,
  called from Network.Wai.Handler.Warp.Run.runSettings.\,
  called from Network.Wai.Handler.Warp.Run.runSettings,
  called from Network.Wai.Handler.Warp.Run.run,
  called from Main.main
```

When adding a logger of some kind one can verify that the exception is `threadWait: invalid argument (Bad file descriptor)`. Is there anything I can do about this?

I created this over at servant but I was recommended to open an issue here too. https://github.com/haskell-servant/servant/issues/1165

kazu-yamamoto:
This should fix #739. To log exceptions from applications, I think `setOnException` should be used.

With this patch, 

```
threadWait: invalid argument (Bad file descriptor)
```

becomes

```
Thread killed by timeout manager
```

Please read #739 to know how to test.

@snoyberg I don't expect that this PR will be merged soon. Rather, I would like to discuss what behavior is reasonable.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Recv.hs
+++ b/warp/Network/Wai/Handler/Warp/Recv.hs
@@ -96,6 +96,9 @@ spell init0 siz0 recv recvBuf
                 siz' = siz - len
             loop bss' siz'
 
+-- The timeout manager may close the socket.
+-- In that case, an error of "Bad file descriptor" occurs.
+-- We ignores it because we expect TimeoutThread.
 receive :: Socket -> BufferPool -> Recv
 receive sock pool = E.handle handler $ withBufferPool pool $ \ (ptr, size) -> do
 #if MIN_VERSION_network(3,1,0)
--- a/warp/Network/Wai/Handler/Warp/Recv.hs
+++ b/warp/Network/Wai/Handler/Warp/Recv.hs
@@ -17,7 +17,9 @@ import Foreign.C.Types
 import Foreign.ForeignPtr (withForeignPtr)
 import Foreign.Ptr (Ptr, castPtr, plusPtr)
 import GHC.Conc (threadWaitRead)
+import qualified GHC.IO.Exception as E
 import Network.Socket (Socket)
+import qualified System.IO.Error as E
 #if MIN_VERSION_network(3,1,0)
 import Network.Socket (withFdSocket)
 #else
@@ -95,7 +97,7 @@ spell init0 siz0 recv recvBuf
             loop bss' siz'
 
 receive :: Socket -> BufferPool -> Recv
-receive sock pool = withBufferPool pool $ \ (ptr, size) -> do
+receive sock pool = E.handle handler $ withBufferPool pool $ \ (ptr, size) -> do
 #if MIN_VERSION_network(3,1,0)
   withFdSocket sock $ \fd -> do
 #elif MIN_VERSION_network(3,0,0)
@@ -105,6 +107,11 @@ receive sock pool = withBufferPool pool $ \ (ptr, size) -> do
 #endif
     let size' = fromIntegral size
     fromIntegral <$> receiveloop fd ptr size'
+  where
+    handler :: E.IOException -> IO ByteString
+    handler e
+      | E.ioeGetErrorType e == E.InvalidArgument = return ""
+      | otherwise                                = E.throwIO e
 
 receiveBuf :: Socket -> RecvBuf
 receiveBuf sock buf0 siz0 = do
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -17,7 +17,7 @@ import qualified Control.Concurrent as Conc (yield)
 import Control.Exception as E
 import qualified Data.ByteString as S
 import Data.Char (chr)
-import Data.IORef (IORef, newIORef, readIORef, writeIORef, atomicModifyIORef')
+import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Data.Streaming.Network (bindPortTCP)
 import Foreign.C.Error (Errno(..), eCONNABORTED)
 import GHC.IO.Exception (IOException(..))
@@ -282,10 +282,7 @@ fork :: Settings
 fork set mkConn addr app counter ii = settingsFork set $ \unmask ->
     -- Call the user-supplied on exception code if any
     -- exceptions are thrown.
-    handle (settingsOnException set Nothing) .
-    -- Allocate a new IORef indicating whether the connection has been
-    -- closed, to avoid double-freeing a connection
-    withClosedRef $ \ref ->
+    handle (settingsOnException set Nothing) $
         -- Run the connection maker to get a new connection, and ensure
         -- that the connection is closed. If the mkConn call throws an
         -- exception, we will leak the connection. If the mkConn call is
@@ -296,21 +293,13 @@ fork set mkConn addr app counter ii = settingsFork set $ \unmask ->
         -- We grab the connection before registering timeouts since the
         -- timeouts will be useless during connection creation, due to the
         -- fact that async exceptions are still masked.
-        bracket mkConn (cleanUp ref) (serve unmask ref)
+        bracket mkConn cleanUp (serve unmask)
   where
-    withClosedRef inner = newIORef False >>= inner
-
-    closeConn ref conn = do
-        isClosed <- atomicModifyIORef' ref $ \x -> (True, x)
-        unless isClosed $ connClose conn
-
-    cleanUp ref (conn, _) = closeConn ref conn `finally` connFree conn
+    cleanUp (conn, _) = connClose conn `finally` connFree conn
 
     -- We need to register a timeout handler for this thread, and
-    -- cancel that handler as soon as we exit. We additionally close
-    -- the connection immediately in case the child thread catches the
-    -- async exception or performs some long-running cleanup action.
-    serve unmask ref (conn, transport) = bracket register cancel $ \th -> do
+    -- cancel that handler as soon as we exit.
+    serve unmask (conn, transport) = bracket register cancel $ \th -> do
         -- We now have fully registered a connection close handler in
         -- the case of all exceptions, so it is safe to once again
         -- allow async exceptions.
@@ -322,8 +311,7 @@ fork set mkConn addr app counter ii = settingsFork set $ \unmask ->
            -- above ensures the connection is closed.
            when goingon $ serveConnection conn ii th addr transport set app
       where
-        register = T.registerKillThread (timeoutManager ii)
-                                        (closeConn ref conn)
+        register = T.registerKillThread (timeoutManager ii) (return ())
         cancel   = T.cancel
 
     onOpen adr    = increase counter >> settingsOnOpen  set adr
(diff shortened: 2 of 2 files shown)


snoyberg:
Just a guess, but: is it possible that the behavior of `network` changed? I think Warp used to rely on the idea that you could close a connection twice, and the second would be a no-op.

kazu-yamamoto:
This is not a double-close problem. The problem is that one thread closes a socket while the other thread is using it. Note that `threadWait` is used in `recv`.

snoyberg:
Oh, got it. But that's exactly what we want to happen: we want to cancel the current `recv` with this timeout. Maybe we should wrap the `recv` call with a `catch` that swallows that exception if some flag is set indicating that the timeout triggered.

kazu-yamamoto:
> Maybe we should wrap the recv call with a catch that swallows that exception if some flag is set indicating that the timeout triggered.

I have already tried this and the same exception was also thrown from a `write` related function. At that time, I felt that closing a socket by the timeout manager is not a good approach.

What I don't understand yet is why `setOnException` is not good enough for the user of #602.

kazu-yamamoto:
@snoyberg If you don't have strong objections, I would like to merge this PR.

snoyberg:
I'm concerned about it. I think it's breaking functionality we have right now. Specifically, I believe it opens us up to a slowloris attack, since an attacker would be able to delay sending data and the manager won't be able to kill the connection.

Maybe there's some other mitigation in place that I'm not remembering.

kazu-yamamoto:
I would like to check if we have the same view. If I understand correctly, the user of #602 would like to write the code like this:

```haskell

import qualified Control.Exception as E
import Network.Wai
import Network.Wai.Internal

main :: IO ()
main = runSettings settings application
  where
    settings = setPort 8080 $ defaultSettings
    application _ respond = respond (responseLBS status200 [("Content-Type", "text/plain")] "Hello World") `E.catch` \(E.SomeException e) -> do
        print e
        return ResponseReceived
```

And you concern that it is vulnerable against slowloris attacks if `print e` take a long time. But the standard way to write this code is:

```haskell

import Network.Wai

main :: IO ()
main = runSettings settings application
  where
    settings = setOnException (\_ e -> print e) $ setPort 8080 $ defaultSettings
    application _ respond = respond (responseLBS status200 [("Content-Type", "text/plain")] "Hello World")
```

When the timeout manager kills the thread, the socket is closed, then `settingsOnException` is called.

So, even if we recommend the standard way, it will not open up to a new slowloris attack.

Am I missing something?

snoyberg:
I'm not concerned about `print e` taking a long time. I'm concerned that while processing the request headers or body, either Warp or the application will implicitly call `recv` and be blocked while the client slowly sends 1 byte at a time.

kazu-yamamoto:
In that case, on time out, the thread with blocked `recv` is killed by the timeout manager and its socket is closed immediately.

snoyberg:
I suppose that's true. I'm not a huge fan of the change, but I'm OK if you want to merge it.

kazu-yamamoto:
Before merging, let me try to explain my observation. Let's call the current code "(A") and the code patched with this PR "(B)".

- `recv` optimistically calls `recv(2)` with a *non-blocking* socket. If it receives data, return them. If `EAGAIN` is returned, `threadWait` is called. This is just a `takeMVar` waiting for input (data or an error) from the IO manager.

Now suppose that timeout happens.

The case for (A):

- The timeout manager first closes the socket.
- The IO manager receives the event from the kernel and informs the thread by `putMVar`.
- `recv` throws this error and the thread dies due to this error. (This is an expected behavior.)
- It depends on scheduling, though, there is few chance for the timeout manager to kill the thread after closing the socket. (I guess that this is *not* an expected behavior.)

The case for (B):

- The timeout manager kills the thread.
- `recv` is canaled immediately and the thread is killed.
- The socket is closed thanks to `bracket`.

For (A):

- Pros: (I cannot find any advantages. I believe that the security level of (A) is the same as that of (B) agaist slowloris attacks.)
- Cons: As I experienced, it is hard to reason about this code since two threads are managing one resource (socket).

For (B):
- Pros: It is easy to reason about this code because only the standard `bracket` is used. I believe the security level is kept.
- Cons: (I cannot find any disadvantages.)

@snoyberg I hope that this explanation could ease your concern.

snoyberg:
The concern is that (B) relies exclusively on async exceptions behaving correctly, which isn't always the case. Application code, and even Warp itself, may mask exceptions (intentionally or unintentionally). There's also a delay between receiving the async exception and the socket closing, which makes the window of opportunity for a slowloris attack slightly larger.

kazu-yamamoto:
@snoyberg OK. Now I think I understand your concern correctly. Both approaches have their advantages and disadvantages. I would like to ask users which is prefer.

@t3hmrman @ebergq This PR would the window of opportunity for a slowloris attack slightly larger. But do you want this PR to be merged?

t3hmrman:
Hi @kazu-yamamoto thank you very much for the hard work.

I can't speak for the other contributor but my concern was a relatively trivial one --  I wanted to clean up my log messages for `-xc` output. That's not important enough to compromise security for the `wai` project. I can't quantify *how much* more vulnerable `wai` would be made to a slowloris DoS attack and since `wai` is so critical to everything that builds on it (`servant`, `scotty`, `yesod`, etc) **I don't think it's worth the possible breaking change at this moment.**

One idea -- rather than solving this in the timeout manager, is it possible to solve it in `wrappedRecv` or `wrappedRecvN` by adding the ability to check whether the connection is closed or not? A field like ` isClosed :: IORef Bool` on `Connection` could make it easier to detect this case.

If we know that the connection is closed in `wrappedRecv`/`wrappedRecvN`, maybe we can return an empty string instead of attempting to complete the read?

kazu-yamamoto:
> One idea -- rather than solving this in the timeout manager, is it possible to solve it in wrappedRecv or wrappedRecvN by adding the ability to check whether the connection is closed or not? A field like isClosed :: IORef Bool on Connection could make it easier to detect this case.

This is not feasible because `recv` is called when the ref is `False`.

kazu-yamamoto:
@snoyberg I modified this PR to close the socket again by the timeout manager and to ignore the error in `receive`. I think that `IORef` to manage the close status is unnecessary because `close` can be used twice.

t3hmrman:
@kazu-yamamoto thanks for the hard work on this issue!

kazu-yamamoto:
Rebased and merged.
Thank you all!

kazu-yamamoto:

ebergq:
@kazu-yamamoto, @t3hmrman, @snoyberg 

Thank you for all the hard work!

As with @t3hmrman I didn't want this to clutter our logs. I didn't know if the issue was serious or not. We never had any problem with our application despite this issue but it was irritating to see such error log messages when the application worked fine.

PR #792, opened 2020-02-02
Feature/remove auth headers from logger

mckayb:

After submitting your PR:

Was using JWT's in a personal project and realized it was logging all of them when they'd make a request. I have a workaround in the app, but figured this was potentially an easy change if you wanted something like that in here.


snoyberg:
1. This breaks all of the CI builds
2. I don't want to make a major version bump for this
3. If you believe this feature is in fact a breaking change (I tend to agree), please gate it behind some kind of configuration flag so this is a minor version bump

Vlix:
Is duplicated by #950, which is more recent and has more focus right now.

PR #793, opened 2020-02-04
Fixes warnings and bumped stack lts

flip111:


snoyberg:
Looks like this fails with older LTSes.

flip111:
What's the minimum LTS that needs to be supported ?

snoyberg:
You can see the test failures at the end of this PR page. We're currently going back to LTS 11 and Cabal with GHC 8.2.2.

flip111:
I don't know why these builds are timing out.

Issue #421, opened 2015-08-10
warp-tls with http2 emits "send: resource vanished (Broken pipe)" on google chrome requests


aemxdp:
Here's a simple application using wai-3.0.3.0, warp-3.1.2, warp-tls-3.1.0:
```haskell
main = do
    Warp.runTLS tlsSettings Warp.defaultSettings app
    -- Warp.runSettings Warp.defaultSettings app    
  where
    app request respond = do
        print request
        respond $ Wai.responseLBS status200 [] (BS.pack "")
    tlsSettings = Warp.defaultTlsSettings
        { Warp.certFile = "cert.pem"
        , Warp.keyFile  = "key.pem" }
```
When performing requests from Chrome 44, it emits the output in subj (though it's like warning: except the fact of the presence of such output, it works fine), while in Firefox everything is ok. Here's output:
```haskell
-- chrome with tls

<interactive>: send: resource vanished (Broken pipe)

Request { requestMethod = "GET", httpVersion = HTTP/2.0, rawPathInfo = "/breergere", rawQueryString = ""
        , requestHeaders =
            [ ("accept","text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8")
            , ("accept-encoding","gzip, deflate, sdch")
            , ("accept-language","en-US,en;q=0.8")
            , ("upgrade-insecure-requests","1")
            , ("user-agent","Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.130 Safari/537.36") ]
        , isSecure = True, remoteHost = 127.0.0.1:50111, pathInfo = ["breergere"], queryString = []
        , requestBody = <IO ByteString>, vault = <Vault>, requestBodyLength = ChunkedBody
        , requestHeaderHost = Just "localhost:3000", requestHeaderRange = Nothing }

-- firefox with tls

Request { requestMethod = "GET", httpVersion = HTTP/2.0, rawPathInfo = "/dfgsgse", rawQueryString = ""
        , requestHeaders =
            [ ("user-agent","Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:35.0) Gecko/20100101 Firefox/35.0")
            , ("accept","text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
            , ("accept-language","en-US,en;q=0.5")
            , ("accept-encoding","gzip, deflate") ]
[...]

Issue #196, opened 2013-11-07
Long-running application and closed connections


snoyberg:
I believe this is the cause of https://github.com/snoyberg/keter/issues/29. It's not necessarily a bug in Warp, but it's a design point that requires some consideration. Consider the following interaction between an HTTP client, Warp, and a WAI application:
1. Client connects to server.
2. Warp parses request and passes it to Application.
3. Application does something that takes a long period of time. Warp currently _disables timeout handling_ to specifically allow for these kinds of long-polling situations.
4. Client meanwhile gets bored and cancels the connection. Warp will not notice that the connection has been closed until it tries to send data to the client, which will only happen after the Application returns a Response.
5. Application completes its long-running action. It now goes to (for example) a message queue it has for messages to be sent to the client, takes a message off the queue, forms a Response from it, and returns that Response to Warp.
6. Warp tries to send that Response to the client, but the connection has already been closed, so the data is thrown away.

What we'd theoretically want is some kind of a notification in the Application when the connection is closed. Either the Application could do some polling to determine that, or an async exception could be thrown whenever the connection is closed. One idea would be for WAI to include a new field, `clientIsListener :: IO Bool`. That field wouldn't necessarily be accurate, so perhaps a `Maybe Bool` response would be more accurate.

Another possibility is to say that the current behavior is acceptable, and it's the responsibility of the Application to institute some kind of a heartbeat.

Pinging @chrisdone and @jwiegley.

exFalso:
Bump, this would be nice to have. I'm implementing a sockjs wai middleware and some tests in the specification require the server to close the connection if a poll request is aborted - however there is no way to detect this in warp/wai now.

If i understand it correctly one can detect the socket disconnect by checking whether recv returns a zero length bytestring. The only problem is that whatever thread is checking this needs to eagerly buffer incoming data just to see whether the connection has been dropped while the app is processing. Not fun.

However the user could provide a maximum buffer size - then the server could detect disconnects as long as unprocessed incoming data fits in this buffer. I think at least for polling protocols this would be a good enough "best effort" solution - polling requests are pretty much empty after all

reiddraper:
[...]

PR #795, opened 2020-02-27
Convert ResourceVanished error to ConnectionClosedByPeer exception an…

marinelli:
…d refine the catching

Hi everyone. At @typeable, we have a big web application, one of its components is a web API served by a warp+wai+servant server. Every request takes a lot of time to be resolved due to connections to multiple external interfaces. If I close the http client (for example killing curl or closing the browser), `warp` returns an error like `Network.Socket.sendBuf: resource vanished (Broken pipe)`.

The reason is that the function `socketConnection`, in the `Network.Wai.Handler.Warp.Run` module, calls the `sendAll` function (from the `Network.Socket.ByteString` module) but the remote http client has already closed the connection.

I tried to create a simple example to reproduce this error, without success.
It might be related to https://github.com/yesodweb/wai/issues/196 or some bad interactions between warp+wai and the servant application. I have to admit that I still don't know the reason.

Anyway, inspired by this old issue https://github.com/yesodweb/wai/issues/421, I think it would be good to catch the `IOException` with error type `ResourceVanished` and throw a `ConnectionClosedByPeer`, like has been done in the `Network.Wai.Handler.WarpTLS` module.

I also refined the exception catching for both the `warp` and `warp-tls` package like was suggested here https://github.com/yesodweb/wai/commit/91d054635b35bfc1a7c7022cda9255365f231ffd#r12752366.

What do you think?

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -43,7 +43,7 @@ module Network.Wai.Handler.WarpTLS (
     ) where
 
 import Control.Applicative ((<|>))
-import Control.Exception (Exception, throwIO, bracket, finally, handle, fromException, try, IOException, onException, handleJust)
+import Control.Exception (Exception, throwIO, bracket, finally, handle, fromException, try, IOException, onException, SomeException(..), handleJust)
 import qualified Control.Exception as E
 import Control.Monad (void, guard)
 import qualified Data.ByteString as S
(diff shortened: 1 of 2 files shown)


marinelli:
Hi @kazu-yamamoto,
If I may, what do you think about this solution?

kazu-yamamoto:
Sorry for the delay. I didn't notice this. I have merged this PR.

marinelli:
It would be helpful to have a new release.

kazu-yamamoto:
@snoyberg This changed the behavior. Would you suggest a new version number?

snoyberg:
I don't think it's necessary

kazu-yamamoto:
@snoyberg Did you mean that 3.3.10 is good enough?

snoyberg:
Yes I think it's fine

kazu-yamamoto:
@snoyberg Thank you!

kazu-yamamoto:
3.3.10 has been released.

marinelli:
Thank you everyone

PR #800, opened 2020-04-30
fixing timeout of builder (#797)

### LINKED ISSUE #797: Large builder-generated responses are unreasonably timed out while being written
If an application provides a builder for a response body that builds a sufficiently large value then it cannot be written to a client before the "slow-loris" timeout engages and closes the connection.  This seems to be because the `sendRsp` implementation for `RspBuilder` never tickles the timeout manager while sending the response data to the connection.

Imagine a response body of 1000 KiB and a client which can receive data at 10 KiB / sec.  It will take around 100 seconds to deliver this response to the client.  Meanwhile, the reasonable default timeout will elapse and the connection will be closed.  This is regardless of the real transfer progress being made.  The client may have received most of the data at the point the timeout kicks in but this won't save it.  Of course, if these numbers seem too small, they can all be multiplied by any factor and the scenario holds.

The issue can be worked around, to some extent, by raising the timeout.  However, a timeout which is large enough to handle cases like this is also too large for small responses.

It seems like the "slow-loris" timeout should be reserved for truly slow transfers.  Perhaps it could be tickled whenever enough progress is made writing the response.

It is also worth noting that if one manages to hand Warp a *stream* response, the `sendRsp` implementation for `RspStream` uses `sendFragment` which does tickle the timeout manager.  It's hard to see why, from an application perspective, there should be any difference in timeout behavior between providing a builder and providing a stream.  This seems doubly important in cases where it might be hard to provide one instead of the other (in my particular case, returning a JSON body by way of Servant, it's not clear how to cause the response to be built with a stream instead of a builder (from a lazy bytestring)).

kazu-yamamoto:
This should fix #797. But not tested by myself.

@snoyberg The essential change is `(\bs -> connSendAll conn bs >> T.tickle th) `. Other changes are just passing `Handle` as an argument.

@exarkun Could you test this PR?

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -116,14 +116,14 @@ sendResponse settings conn ii th req reqidxhdr src response = do
         -- and status, the response to HEAD is processed here.
         --
         -- See definition of rsp below for proper body stripping.
-        (ms, mlen) <- sendRsp conn ii ver s hs rspidxhdr rsp
+        (ms, mlen) <- sendRsp conn ii th ver s hs rspidxhdr rsp
         case ms of
             Nothing         -> return ()
             Just realStatus -> logger req realStatus mlen
         T.tickle th
         return ret
       else do
-        _ <- sendRsp conn ii ver s hs rspidxhdr RspNoBody
+        _ <- sendRsp conn ii th ver s hs rspidxhdr RspNoBody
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
@snoyberg May I merge this PR?

snoyberg:
Yes, sorry, I thought I'd submitted feedback!

kazu-yamamoto:
Thanks.

Rebased and merged.

PR #801, opened 2020-05-12
adding setAltSvc.

kazu-yamamoto:
I'm now implementing WarpQUIC. To access contents provided by WarpQUIC via Firefox nightly, the `AltSvc:` header is necessary:

```
Alt-Svc: h3-27=":4433"
```

This PR adds `setAltSvc` to specify such a value if necessary.

### DIFF
--- a/warp/Network/Wai/Handler/Warp.hs
+++ b/warp/Network/Wai/Handler/Warp.hs
@@ -75,6 +75,7 @@ module Network.Wai.Handler.Warp (
   , setGracefulCloseTimeout1
   , setGracefulCloseTimeout2
   , setMaxTotalHeaderLength
+  , setAltSvc
     -- ** Getters
   , getPort
   , getHost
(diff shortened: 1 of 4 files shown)


snoyberg:
Also: very exciting, thank you for doing this!

kazu-yamamoto:
Here is context: URL cannot tell that QUIC is available because QUIC is based on UDP. Chrome Canary has an option `--origin-to-force-quic-on=`. For example, if we specify `localhost:4433` to Canary and access to 4433 with Canary, Canary tries to use QUIC.

However, Firefox nightly does not have such a command line option. So, what Nighly can do is just access `https://localhost`(port 443) first. In this case, we need to tell Nightly that QUIC is also available for the resources via `Alt-Svc: h3=":4433"`. Then Nighly tries to access the resources with QUIC. Note that `https://localhost:4433` does not work.

kazu-yamamoto:
Note that `Alt-Svc` is designed so that it can also be used in plain HTTP. But, if I understand correctly, Firefox and Canary only use it in the *HTTPS* context.

kazu-yamamoto:
@snoyberg Thank you for your review. I have merged this PR.

kazu-yamamoto:
FYI: https://http3-explained.haxx.se/en/h3/h3-altsvc

kazu-yamamoto:
Warp v3.3.11 has been released.

snoyberg:
Thanks for the pointers, I'm looking forward to this :D

PR #804, opened 2020-05-25
flush SSE headers early

### LINKED ISSUE #803: is it possible to flush headers early when using responseStream ?
I'm using wai-extra's `eventSourceAppChan` and the headers are only sent after the first activity on the chan, which causes clients like curl to timeout with `curl: (52) Empty reply from server`.

I think the headers should be flushed right after the connection is initiated.
Any idea how I could achieve that?

docteurklein:
fixes #803

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for wai-extra
 
+## 3.0.29.2
+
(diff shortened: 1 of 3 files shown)


snoyberg:
Could you add a minor version bump and a changelog entry?

docteurklein:
done, thank you!

snoyberg:
Thanks!

PR #805, opened 2020-05-27
TLS parameters for sharing

kazu-yamamoto:
I would like to share credentials and a session manager between warp-tls and warp-quic. For this purpose, I would like to add new config fields. If these values are `Nothing` (default), no behavior changed. If these values are `Just`, the values are used. This code is already tested in the real world.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -24,6 +24,7 @@ module Network.Wai.Handler.WarpTLS (
     -- * Accessors
     , certFile
     , keyFile
+    , tlsCredentials
     , tlsLogging
     , tlsAllowedVersions
     , tlsCiphers
@@ -50,6 +51,7 @@ import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as L
 import Data.Default.Class (def)
 import qualified Data.IORef as I
+import Data.Maybe (isJust, fromJust)
 import Data.Streaming.Network (bindPortTCP, safeRecv)
 import Data.Typeable (Typeable)
 import GHC.IO.Exception (IOErrorType(..))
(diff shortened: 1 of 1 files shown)


warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -260,26 +275,36 @@ runTLS tset set app = withSocketsDo $
 
 ----------------------------------------------------------------
 
+loadCredentials :: TLSSettings -> IO TLS.Credentials
+loadCredentials TLSSettings{..}
+  | isJust tlsCredentials = return $ fromJust tlsCredentials
```
snoyberg, line 280: I'd rather see pattern matching than `fromJust`, how about:

```haskell
  | Just creds <- tlsCredentials = return creds
```

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -260,26 +275,36 @@ runTLS tset set app = withSocketsDo $
...
+        return $ TLS.Credentials [cred]
+    (mcert, mkey) -> do
+        cert <- maybe (S.readFile certFile) return mcert
+        key <- maybe (S.readFile keyFile) return mkey
+        cred <- either error return $ TLS.credentialLoadX509ChainFromMemory cert chainCertsMemory key
+        return $ TLS.Credentials [cred]
+
+getSessionManager :: TLSSettings -> IO TLS.SessionManager
+getSessionManager TLSSettings{..}
+  | isJust tlsSessionManager = return $ fromJust tlsSessionManager
```
snoyberg, line 293: Same comment about partial `fromJust` from above

warp-tls/Network/Wai/Handler/WarpTLS.hs
kazu-yamamoto: Done.

warp-tls/Network/Wai/Handler/WarpTLS.hs
kazu-yamamoto: Done.
Please review again.

kazu-yamamoto:
Rebased and merged. Thank you for your review.

Issue #463, opened 2015-11-13
Add support for the Automatic Certificate Management Environment to warp


werksjan:
Since [Let's Encrypt](https://letsencrypt.org/) will start it's public beta in early december, an [ACME](https://github.com/ietf-wg-acme/acme/blob/master/draft-ietf-acme-acme.md)-enabled CA would be available. Adding support for the protocoll to warp in time, as does the [Let's Encrypt Client](https://github.com/letsencrypt/letsencrypt) with Apache, would be a good statement pro https.

I could imagine warp default to launching warp-tls alongside http and self-requesting an appropriate certificate from any free, trusted DV-CA as it sees fit (no certificate avaiable, certificate about to expire).

This could enable https working out-of-the-box.
At least a configuration flag for warp-tls would be great that could toggle this behavior.
The only problem I see at the moment is to let warp-tls know the intended domain names it should ask for certificates for.

kazu-yamamoto:
I'm strongly interested in let's encrypt, too.

> self-requesting an appropriate certificate from any free, trusted DV-CA as it sees fit 

For this, how should we modify warp-tls?

werksjan:
I have been thinking about that for a while.

1. This might be a question of modifying the default settings to hold appropriate, sensible default values that allow to use a afaik yet-to-be-done ACME client to do so.

The downside would be that this cannot cover the domains the certificate should be valid for.

2. Another possible way could be to integrate a small WAI-App that allows user interaction as long as no certificates are found. (think: wordpress installation progress)

The downside here is that this can introduce problems if anyone deployed such a configuration reachable from the outside because anyone could get certificates then.

3. The third way I can think of would be to use the beforementioned ACME client whenever no certificates are found, prompting the developer starting warp-tls w/o certificates for the needed information.

This might cause problems w/ current use cases such as yesod devel.

Another problem might be certificate renewal and integration w/ HPKP. To be reliable and cause no additional problems, checking the status should be done regulary. For example on startup and then scheduling a thread to renew the certificate and modify the HPKP-Headers as indicated, which schedules another one for actual switching out of certificates.

werksjan:
As Let's encrypt started it's open beta in the meanwhile, I'd like to bump this again. Probably a good time, too, to keep an eye on new hackage package releases may helping us out by implementing at least part of the features.

kazu-yamamoto:
[...]

PR #806, opened 2020-06-22
Add the ability to load TLS credentials from IORefs

mbg:
This PR adds three new configuration fields to `TLSSettings` for retrieving the certificate bytes, chain certificate bytes, and key bytes from `IORef`s. This is potentially useful for scenarios where the certificate may need to change at runtime, e.g. if ACME is used. In theory, this would prevent an ACME implementation from having to terminate the server thread in order to call `runTLS` with new TLS credentials. However, there is currently no ACME implementation in Haskell (to my knowledge), so this is a somewhat speculative PR.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -91,7 +91,9 @@ data TLSSettings = TLSSettings {
     -- loaded from?
     --
     -- >>> certSettings defaultTlsSettings
+    -- tlsSettings "certificate.pem" "key.pem"
     -- 
+    -- @since 3.3.0
   , onInsecure :: OnInsecure
     -- ^ Do we allow insecure connections with this server as well?
     --
@@ -264,7 +266,7 @@ tlsSettingsChainMemory cert chainCerts key = defaultTlsSettings {
 -- | A smart constructor for 'TLSSettings', but uses references to in-memory
 -- representations of the certificate and key based on 'defaultTlsSettings'.
 --
--- @since 3.2.13
+-- @since 3.3.0
 tlsSettingsRef 
     :: I.IORef S.ByteString -- ^ Reference to certificate bytes
     -> I.IORef (S.ByteString) -- ^ Reference to key bytes 
@@ -276,7 +278,7 @@ tlsSettingsRef cert key = defaultTlsSettings {
 -- | A smart constructor for 'TLSSettings', but uses references to in-memory
 -- representations of the certificate and key based on 'defaultTlsSettings'.
 --
--- @since 3.2.13
+-- @since 3.3.0
 tlsSettingsChainRef 
     :: I.IORef S.ByteString -- ^ Reference to certificate bytes
     -> [I.IORef S.ByteString] -- ^ Reference to chain certificate bytes
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -21,6 +21,8 @@ module Network.Wai.Handler.WarpTLS (
     , tlsSettingsMemory
     , tlsSettingsChain
     , tlsSettingsChainMemory
+    , tlsSettingsRef
+    , tlsSettingsChainRef
     -- * Accessors
     , certFile
     , keyFile
@@ -82,6 +84,12 @@ data TLSSettings = TLSSettings {
   , certMemory :: Maybe S.ByteString
   , chainCertsMemory :: [S.ByteString]
   , keyMemory :: Maybe S.ByteString
+  , certRef :: Maybe (I.IORef S.ByteString)
+    -- ^ An optional reference to the certifiate in memory.
+  , chainCertsRef :: Maybe (I.IORef [S.ByteString])
+    -- ^ An optional reference to a list of chain certificates in memory.
+  , keyRef :: Maybe (I.IORef S.ByteString)
+    -- ^ An optional reference to the key in memory.
   , onInsecure :: OnInsecure
     -- ^ Do we allow insecure connections with this server as well?
     --
@@ -181,6 +189,9 @@ defaultTlsSettings = TLSSettings {
   , certMemory = Nothing
   , chainCertsMemory = []
   , keyMemory = Nothing
+  , certRef = Nothing
+  , chainCertsRef = Nothing 
+  , keyRef = Nothing
   , onInsecure = DenyInsecure "This server only accepts secure HTTPS connections."
   , tlsLogging = def
 #if MIN_VERSION_tls(1,5,0)
@@ -262,6 +273,34 @@ tlsSettingsChainMemory cert chainCerts key = defaultTlsSettings
     , keyMemory = Just key
     }
 
+-- | A smart constructor for 'TLSSettings', but uses references to in-memory
+-- representations of the certificate and key based on 'defaultTlsSettings'.
+--
+-- Since ???
+tlsSettingsRef 
+    :: I.IORef S.ByteString -- ^ Reference to certificate bytes
+    -> I.IORef (S.ByteString) -- ^ Reference to key bytes 
+    -> TLSSettings 
+tlsSettingsRef cert key = defaultTlsSettings
+    { certRef = Just cert 
+    , keyRef = Just key
+    }
+
+-- | A smart constructor for 'TLSSettings', but uses references to in-memory
+-- representations of the certificate and key based on 'defaultTlsSettings'.
+--
+-- Since ???
+tlsSettingsChainRef 
+    :: I.IORef S.ByteString -- ^ Reference to certificate bytes
+    -> I.IORef [S.ByteString] -- ^ Reference to chain certificate bytes
+    -> I.IORef (S.ByteString) -- ^ Reference to key bytes 
+    -> TLSSettings 
+tlsSettingsChainRef cert chainCerts key = defaultTlsSettings
+    { certRef = Just cert 
+    , chainCertsRef = Just chainCerts
+    , keyRef = Just key
+    }
+
 ----------------------------------------------------------------
 
 -- | Running 'Application' with 'TLSSettings' and 'Settings'.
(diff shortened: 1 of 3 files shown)


warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -82,6 +84,12 @@ data TLSSettings = TLSSettings {
   , certMemory :: Maybe S.ByteString
   , chainCertsMemory :: [S.ByteString]
   , keyMemory :: Maybe S.ByteString
+  , certRef :: Maybe (I.IORef S.ByteString)
```
snoyberg, line 87: It seems to me like the logic here is starting to get difficult to follow, comparing `certMemory` vs `certRef` vs `certFile`. What about introducing sum types for the different possibilities?

mbg: That makes a lot of sense - done in d9c8861. 

I don't think this is a breaking change, although the behaviour in `loadCredentials` is now different. For the `ByteString` options it previously used `maybe` to fall back to filenames, but I don't think it was actually possible to trigger this behaviour since the fields of `TLSSettings` are not exported and thus `tlsSettingsMemory` and `tlsSettingsChainMemory` were the only ways to set `certMemory` and `keyMemory`.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -71,17 +73,52 @@ import System.IO.Error (isEOFError, ioeGetErrorType)
 
 ----------------------------------------------------------------
 
--- | Settings for WarpTLS.
-data TLSSettings = TLSSettings {
+-- | Determines where to load the certificate, chain 
+-- certificates, and key from.
+data CertSettings 
+  = CertFromFile {
     certFile :: FilePath
```
snoyberg, line 80: I prefer not putting record names on sum types, since it creates partial functions and accessors.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -71,17 +73,52 @@ import System.IO.Error (isEOFError, ioeGetErrorType)
 
 ----------------------------------------------------------------
 
--- | Settings for WarpTLS.
-data TLSSettings = TLSSettings {
+-- | Determines where to load the certificate, chain 
+-- certificates, and key from.
+data CertSettings 
+  = CertFromFile {
```
snoyberg, line 79: I'd been thinking of something slightly different, more like:

```haskell
data BytesSource = FromFile !FilePath | FromMemory !ByteString | FromRef !(IORef ByteString)

data TLSSettings = TLSSettings
  { ...
  , certSource :: !BytesSource
  , chainCertsSource :: ![BytesSource]
  , keySource :: !BytesSource
  }
```

Not convinced my way is better, but it does allow a bit more flexibility in mixing and matching data sources.

warp-tls/Network/Wai/Handler/WarpTLS.hs
mbg: No problem, I can remove them.

warp-tls/Network/Wai/Handler/WarpTLS.hs
mbg: That is true. I could change this, but we would no longer be able to use `credentialLoadX509Chain`. We would then have to use `readSignedObject` directly for each file, but that doesn't return a `ByteString` and neither does `readSignedObjectFromMemory`, so we would need something like:

```haskell
certFromBytesSource :: (ASN1Object a, Eq a, Show a) => BytesSource -> IO [SignedExact a]
certFromBytesSource (FromFile fp) = readSignedObject fp
certFromBytesSource (FromMemory bs) = readSignedObjectFromMemory fp
certFromBytesSource (FromRef ref) = readIORef ref >>= readSignedObjectFromMemory

-- something similar for keys with readKeyFile and readKeyFileFromMemory

loadCredentials TLSSettings{..} = do 
  cert <- certFromBytesSource certSource
  chain <- mapM certFromBytesSource chainCertsSource
  key <- keyFromBytesSource keySource
  case keys of
    []    -> error "no keys found"
    (key:_) -> return (CertificateChain . concat $ cert : chain, key)
```

snoyberg: In that case, I don't have a strong opinion either way. If you think the current approach is better, let's stick with it. I'd like to get sign-off from Kazu, but otherwise no more comments from me. Thanks!

warp-tls/Network/Wai/Handler/WarpTLS.hs
mbg: One slight problem with removing the fields is that `certFile` and `keyFile` are currently exported by the module, so removing them or changing their types would not be backwards compatible.

snoyberg: Leaving it as-is is a problem anyway, since (1) it changes their types already from `TLSSettings -> val` to `CertSettings -> val` and (2) it makes them partial. IMO, it's worth a breaking change to hide these away, since they don't behave the way an outside user would really expect.

mbg:
I can bump the version to `3.3.0` in light of the export changes if everything else is OK.

kazu-yamamoto:
LGTM.

@snoyberg Should we bump the version to 3.3.0?

kazu-yamamoto:
This should close #463

snoyberg:
Yes, let's go for the 3.3.0 version bump.

mbg:
Done.

kazu-yamamoto:
I would like to test this code for a while. After that I will take care of merging and releasing.

kazu-yamamoto:
v3.3.0 has been released. Thank you for your contribution!

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -71,17 +71,29 @@ import System.IO.Error (isEOFError, ioeGetErrorType)
 
 ----------------------------------------------------------------
 
+-- | Determines where to load the certificate, chain 
+-- certificates, and key from.
+data CertSettings 
+  = CertFromFile !FilePath ![FilePath] !FilePath
+  | CertFromMemory !S.ByteString ![S.ByteString] !S.ByteString
+  | CertFromRef !(I.IORef S.ByteString) ![I.IORef S.ByteString] !(I.IORef S.ByteString)
```
vdukhovni, line 79: I'm rather late to party, sorry about that, just noticed this PR today, and sadly it has been merged for some time now.  And yet perhaps I should make my comments anyway.  My first issue is with the choice of separate IORefs for each of the cert, chain and keys.  I think that's likely a mistake.  Rather I think that the IORef representation should have been:
```haskell
CertFromRef !(I.IORef (S.ByteString,  [S.ByteString], S.ByteString))
```
That is, a single `IORef` holding a triple with the certificate, issuer chain, and key.  This would allow atomic retrieval of the triple, in which one can be confident that all three objects correspond to each other.  The current design is subject race conditions between reader and writer, because the writer updates multiple `IORef`s separately, and the reader may see an inconsistent view of the data. :-(

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -256,11 +259,34 @@ tlsSettingsChainMemory
...
+  }
+
+-- | A smart constructor for 'TLSSettings', but uses references to in-memory
+-- representations of the certificate and key based on 'defaultTlsSettings'.
+--
+-- @since 3.3.0
+tlsSettingsRef 
+    :: I.IORef S.ByteString -- ^ Reference to certificate bytes
+    -> I.IORef (S.ByteString) -- ^ Reference to key bytes 
+    -> TLSSettings
```
vdukhovni, line 273: I would have argued strongly against providing this function. The most widely used CA on the planet is now "Let's Encrypt" and it along with almost all the other CAs does not sign the leaf certificate directly with a trusted root CA. Instead pretty much all certificates that chain up to one of the usual WebPKI CAs are signed by an _intermediate_ CA.

When you have a certificate signed by an intermediate CA, while various browsers are willing to pay the cost of dynamically locating intermediate CA certs from various certificate extensions, and may have an intermediate cert cache, other software tools (e.g. the ubiquitous `curl`, etc.) are not, and the TLS specifications DO NOT make intermediate certificates optional. They are a required part of the server's certificate chain.

Here the caller can always provide an empty chain if that's what they really want to do, but there's no need to encourage such negligence by providing an API entry point that can only do the wrong thing.  The more general entry point below is the only one that should have been offered.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -256,11 +259,34 @@ tlsSettingsChainMemory
...
+
+-- | A smart constructor for 'TLSSettings', but uses references to in-memory
+-- representations of the certificate and key based on 'defaultTlsSettings'.
+--
+-- @since 3.3.0
+tlsSettingsChainRef 
+    :: I.IORef S.ByteString -- ^ Reference to certificate bytes
+    -> [I.IORef S.ByteString] -- ^ Reference to chain certificate bytes
+    -> I.IORef (S.ByteString) -- ^ Reference to key bytes 
+    -> TLSSettings
```
vdukhovni, line 286: And of course in this function, there should have been just a single `IORef` holding a triple of `(Cert, [Chain], Key)` bytestrings/lists of bytestrings.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -276,15 +302,19 @@ runTLS tset set app = withSocketsDo $
...
+loadCredentials TLSSettings{..} = case certSettings of 
+  CertFromFile cert chainFiles key -> do
+    cred <- either error id <$> TLS.credentialLoadX509Chain cert chainFiles key
+    return $ TLS.Credentials [cred]
+  CertFromRef certRef chainCertsRef keyRef -> do 
+    cert <- I.readIORef certRef
+    chainCerts <- mapM I.readIORef chainCertsRef
+    key <- I.readIORef keyRef
+    cred <- either error return $ TLS.credentialLoadX509ChainFromMemory cert chainCerts key
+    return $ TLS.Credentials [cred]
```
vdukhovni, line 314: Which brings us to perhaps the most important issue. What is the purpose of the `IORef` alternative, if ultimately `loadCredentials` converts them (if I'm not mistaken just once) immediately to a set of pure objects used in:
```haskell
runTLSSocket :: TLSSettings -> Settings -> Socket -> Application -> IO ()
runTLSSocket tlsset set sock app = do
    credentials <- loadCredentials tlsset
    mgr <- getSessionManager tlsset
    runTLSSocket' tlsset set credentials mgr sock app
```
This still means that the application loop needs to be restarted when new certificates are available, and so I fail to see the difference between this `CertFromMemory`.  What does the indirection via `IORef`s achieve?

It seems like there's much missing code here to make a practically useful deployment model.  Perhaps the idea is to also provide a restart loop (I may start work on that shortly) and have the restart loop be configured once with these (or ideally just one) `IORefs`, allowing another thread to just worry about updating the data as needed?

If so I think that code should probably have appeared concurrently with or even before this PR, which would have made it easier to see whether this API is the right one for the job.

[ My tentative loop restart implementation idea will only work on POSIX systems that support dup(2) on socket file descriptors. Which precludes Windows.  I hope that having reasonably clean certificate reload without dropping any connections even briefly, but just on Unix-like systems is an acceptable design choice.  More ambitious designs would likely require invasive changes in the TLS library, which are perhaps a good idea, but I'm not planning to go there at this time. ]

vdukhovni (commented):
If the results of this merged PR are not getting used much yet, perhaps it might not be too late to make post-release changes?

PR #813, opened 2020-09-17
Fix GHC 7.10 build failure in wai-extra

RyanGlScott:
`wai-extra-3.0.31` fails to build with GHC 7.10.3, as demonstrated in [this failing Travis build](https://travis-ci.org/github/ku-fpg/blank-canvas/jobs/728110964#L1377):

<details>

```
[ 4 of 32] Compiling Network.Wai.Test ( Network/Wai/Test.hs, interpreted )

Network/Wai/Test.hs:210:15:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertBool’:
      assertBool :: HasCallStack => String -> Bool -> Session ()

Network/Wai/Test.hs:213:17:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertString’:
      assertString :: HasCallStack => String -> Session ()

Network/Wai/Test.hs:216:18:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertFailure’:
      assertFailure :: HasCallStack => String -> Session ()

Network/Wai/Test.hs:223:22:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertContentType’:
      assertContentType :: HasCallStack =>
                           ByteString -> SResponse -> Session ()

Network/Wai/Test.hs:240:17:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertStatus’:
      assertStatus :: HasCallStack => Int -> SResponse -> Session ()

Network/Wai/Test.hs:250:15:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertBody’:
      assertBody :: HasCallStack =>
                    L8.ByteString -> SResponse -> Session ()

Network/Wai/Test.hs:258:23:
    Non type-variable argument
      in the constraint: ?callStack::GHC.Stack.CallStack
    (Use FlexibleContexts to permit this)
    In the type signature for ‘assertBodyContains’:
[...]

PR #814, opened 2020-09-18
Network.Wai.Test: Re-enable support for source locations + turn WaiTestFailure into a pattern synonym

sol:
This PR includes https://github.com/yesodweb/wai/pull/813 (thanks @RyanGlScott).

---

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/Test.hs
+++ b/wai-extra/Network/Wai/Test.hs
@@ -2,6 +2,8 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ViewPatterns #-}
 module Network.Wai.Test
     ( -- * Session
       Session
@@ -30,7 +32,8 @@ module Network.Wai.Test
     , assertClientCookieExists
     , assertNoClientCookieExists
     , assertClientCookieValue
-    , WaiTestFailure (..)
+    , WaiTestFailure
+    , pattern WaiTestFailure
     ) where
 
 #if __GLASGOW_HASKELL__ < 710
@@ -46,9 +49,6 @@ import Control.Monad.Trans.Class (lift)
 import qualified Control.Monad.Trans.State as ST
 import Control.Monad.Trans.Reader (runReaderT, ask)
 import Control.Monad (unless)
-import Control.DeepSeq (deepseq)
-import Control.Exception (throwIO, Exception)
-import Data.Typeable (Typeable)
 import qualified Data.Map as Map
 import qualified Web.Cookie as Cookie
 import Data.ByteString (ByteString)
@@ -63,6 +63,7 @@ import qualified Data.Text as T
 import qualified Data.Text.Encoding as TE
 import Data.IORef
 import Data.Time.Clock (getCurrentTime)
+import qualified Test.HUnit.Lang as HUnit
 import Data.CallStack (HasCallStack)
 
 -- |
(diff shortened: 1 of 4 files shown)


wai-extra/Network/Wai/Test.hs
```diff
@@ -214,8 +216,9 @@ assertString :: HasCallStack => String -> Session ()
 assertString s = unless (null s) $ assertFailure s
 
 assertFailure :: HasCallStack => String -> Session ()
-assertFailure msg = msg `deepseq` liftIO (throwIO (WaiTestFailure msg))
+assertFailure = liftIO . HUnit.assertFailure
```
snoyberg, line 219: This may just be deferring the breakage until runtime. If anyone is checking for `WaiTestFailure` in their code, this will now silently break. Is there no way to keep the call stack _and_ keep `WaiTestFailure`?

sol: Yes, that is true.  A user will still get a deprecation warning, so it's not 100% silent, but still not ideal.

We can try to turn `WaiTestFailure` into a pattern synonym; I'll have to check for compiler support, don't know from the top of my head when they were introduced.

Friday night here already, I'll look into this tomorrow.

wai-extra/test/Network/Wai/TestSpec.hs
```diff
@@ -223,5 +227,15 @@ spec = do
...
+      it "matches HUnitFailure" $ do
+        Left (WaiTestFailure err) <- try $ 23 `shouldBe` (42 :: Int)
+        err `shouldBe` "expected: 42\n but got: 23"
 
+    context "when used as an expression" $ do
+      it "it types as HUnitFailure" $ do
+        throwIO (WaiTestFailure "foo") `shouldThrow` instanceOf (undefined :: HUnitFailure)
 
+      it "it types as WaiTestFailure" $ do
+        throwIO (WaiTestFailure "foo") `shouldThrow` instanceOf (undefined :: WaiTestFailure)
```
sol, line 241: @snoyberg this is what we get by turning `WaiTestFailure` into a pattern synonym.  It's close to, but not 100% behavior preserving, as catching `WaiTestFailure` will now catch all instances of `HUnitFailure`.

I still hope that this is good enough for all practical purposes and that it provides a smooth upgrade path.

wai-extra/Network/Wai/Test.hs
```diff
@@ -214,11 +216,14 @@ assertString :: HasCallStack => String -> Session ()
...
 
-data WaiTestFailure = WaiTestFailure String
-    deriving (Show, Eq, Typeable)
-instance Exception WaiTestFailure
+{-# DEPRECATED WaiTestFailure "Use `HUnit.HUnitFailure` instead." #-}
```
sol, line 221: @snoyberg deprecation is optional.  I'm in favor of it for two reasons:

- The pattern synonym is not 100% behavior preserving (as mentioned before).
- Catching and then re-throwing `WaiTestFailure` will not preserve the source location.  Working with `HUnitFailure` directly does not have this limitation.

But not very opinionated on it, we can drop the deprecation if you prefer.

wai-extra/Network/Wai/Test.hs
sol: > I'll have to check for compiler support

Pattern synonyms were introduced with GHC `7.8.1`.  We require `base >= 4.8`, which implies GHC `7.10.1` or later.  So this should work.

sol: > Is there no way to keep the call stack and keep `WaiTestFailure`?

I'm not aware of a backwards compatible and straight forward way, no.

If we would want to keep `WaiTestFailure` as a data type and add source locations to it, we would need to change it to something like:
```haskell
data WaiTestFailure = WaiTestFailure SrcLoc String
```

snoyberg:
After seeing the hoops I'm asking you to jump through here, I'm thinking my original stance on the major version bump. How's this for a plan:

* Re-release the last working version (for older GHCs) to fix Ryan's builds
* Introduce a breaking change that:
    * Drops support for older GHCs (it's entirely accidental that this hasn't happened yet)
    * Drops the WaiTestFailure type

If you're OK with this plan, I'll move ahead with it. Sorry for the extra legwork.

sol:
@snoyberg hey, works for me! :+1:

sol:
https://github.com/yesodweb/wai/pull/817

PR #816, opened 2020-09-21
Drop older GHCs

snoyberg:
@kazu-yamamoto we recently had a PR that broke older GHCs. This is a safety mechanism to prevent that from happening.


kazu-yamamoto (approved):
LGTM.

Please merge this by yourself.

Issue #811, opened 2020-09-14
Behavior of strictRequestBody and lazyRequestBody is undocumented


MaxGabriel:
`strictRequestBody` and `lazyRequestBody` consume the request body such that calling them twice will yield empty string the second time. This is fairly error prone and I think unexpected.

Example cases where you might access the body twice:

1. Middleware, which might read the body to log data about it up front, or could catch an exception in processing the request and log the request's body.
2. Authentication checks, which are separate from application code in e.g. Yesod. For example, [Github's webhook API](https://developer.github.com/webhooks/securing/) is authenticated by applying HMAC to the request body 

In these cases, it may be difficult or impossible to get the `ByteString` body and thread it through to later calls.

I would propose:

1. The functions should definitely be documented to say that a second call with return empty string. 
2. The functions be renamed to something that suggests they are not idempotent, if people can come up with anything better. Possibly `consumeRequestBodyStrict`? `nonIdempotentStrictRequestBody`? `nonRepeatableReadStrictRequestBody`? Not sure.
3. The docs should explain the rationale for this (I assume to keep total memory usage down if possible)
4. The docs should suggest a workaround, if there is a decent way to avoid this behavior. I'm not sure if there is one—the best method I know in Yesod would be to use the per-request caching to store the ByteString there so it can be reused multiple times, but that's Yesod specific. Possibly WAI could be modified to optionally store the results to be retrieved later?

Thoughts?

snoyberg:
1. Agreed
2. I'm not in favor of breaking the API
3. That's reasonable, but I think the answer is straightforward: the entire point of the API is to be streaming, so consuming is a requirement
4. No, that breaks the point of the API and encourages bad behavior. There shouldn't be a "workaround," there should be "if you really think you have this need, do X."

MaxGabriel:
I guess, how do you think about the two example use cases I provided? Is the answer that middleware or middleware-esque functions like Yesod's [isAuthorized](https://www.stackage.org/haddock/lts-16.14/yesod-core-1.6.18/Yesod-Core.html#v:isAuthorized) should never touch the request body?

To me those use-cases are things I run into all the time and are legitimate (especially authentication checks like HMAC on a body). To me it's OK that WAI doesn't handle all use-cases perfectly in order to support the streaming one, but I don't think it's fair to call them "bad behavior".

[...]

PR #818, opened 2020-09-28
Add `consuming` synonyms functions + document streaming

MaxGabriel:
Closes https://github.com/yesodweb/wai/issues/811

After submitting your PR:


wai/Network/Wai.hs
```diff
@@ -315,11 +318,61 @@ ifRequest :: (Request -> Bool) -> Middleware -> Middleware
...
+-- * Consider streaming directly to a file on disk. For an example of this, see the
+-- "Data.Conduit.Binary" module in @conduit-extra@.
+-- * If you need to direct the request body to multiple destinations, you can stream to both those
+-- destinations at the same time.
+-- For example, if you wanted to run an HMAC on the request body as well as parse it into JSON,
+-- you could use Conduit's @zipSinks@ to send the data to @cryptonite-conduit@'s 'sinkHMAC' and
+-- @aeson@'s Attoparsec parser.
+-- * If possible, avoid processing large data on your server at all.
+-- For example, instead of uploading a file to your server and then to AWS S3,
+-- you can have the browser upload directly to S3.
```
MaxGabriel, line 349: I like this section a lot. For several of these cases, I wouldn't have considered streaming before I researched this. For example, I didn't know every attoparsec parser supported streaming, that there were streaming HMAC options, and hadn't considered splitting a conduit stream in two.

snoyberg: I like this a lot too, very nice!

wai/Network/Wai.hs
```diff
@@ -315,11 +318,61 @@ ifRequest :: (Request -> Bool) -> Middleware -> Middleware
...
+--
+-- That said, sometimes it is convenient, or even necessary to read the whole request body into memory.
+-- For these purposes, functions like 'strictRequestBody' or 'lazyRequestBody' can be used.
+-- When this is the case, consider these strategies to mitigating potential DOS attacks:
+--
+-- * Set a limit on the request body size you allow.
```
snoyberg, line 355: This makes me think we should consider taking an optional parameter `Maybe Word64` of the maximum allowed body size for the newly added functions.

MaxGabriel: So, coming from a Yesod perspective I feel like this is already well handled by `maximumContentLength`, but I don't have a good perspective on what WAI attempts to accomplish vs Yesod/Servant/etc

snoyberg: Well, that's the thing: it doesn't really try to address these things. But with this added documentation, it's begging the question: why are these functions encouraging a dangerous practice? I realize this isn't really added by your PR, since the original poorly-named functions really created the problem in the first place. But it's certainly highlighting it.

MaxGabriel: Hm, yeah, it does highlight it. I thought about it a little more, and I don't really think having a parameter on this function to limit the body size is great from a security perspective. Thoughts:

* It would scatter the size checks across the codebase. This makes it harder to audit that it's being used properly
* It requires every developer to make security decisions they might not understand.
* In practice I would expect a lot of `Nothing` values passed (especially since frameworks like Yesod handle this in a better way, so it would be improper to try to do it here as well).

I think what Yesod doing is really great. If I added a middleware to wai-extra to limit request size (should be trivial), and pointed people to either use that or Yesod's functionality (or whatever is in their own web framework), do you think that would be sufficient? I'd rather direct people to use what I think is a great solution

snoyberg: That's a great idea!

wai-extra/Network/Wai/Middleware/RequestSizeLimit.hs
```diff
@@ -0,0 +1,69 @@
...
+                    remaining <- readIORef ref
+                    let len = fromIntegral $ S8.length bs
+                        remaining' = remaining - len
+                    if remaining < len
+                        then do
+                            putStrLn "Too large; sending too large response here"
+                            -- This won't work, because execution continues after sendResponse
+                            sendResponse $ tooLargeResponse maxLen len
+                            putStrLn "Too large; just sent too large response here"
+                            pure ""
```
MaxGabriel, line 33: Ok, I've been pulling in the code from the Yesod implementation, but this part is a little tricky. In Yesod, it does `throwIO $ HCWai $ tooLargeResponse maxLen len`, which Yesod catches and interprets as "send this WAI response".

I initially tried just using `sendResponse` here, but that doesn't work because execution continues past this point. Should I try a similar pattern as Yesod, where I throw an exception here, and attempt to catch it and give the 413? I'm not sure if WAI has this kind of pattern

Just worked on this a little bit tonight, haven't investigated too deeply so need to do more research, but if you have an idea off-hand I'd take that

MaxGabriel:
```
newtype TooLargeRequestException = TooLargeRequestException Response

instance Show TooLargeRequestException where
    show (TooLargeRequestException r) = "TooLargeRequestException: The request exceeded the total available limit. This exception is automatically caught by WAI to provide a proper response"
instance Exception TooLargeRequestException

requestSizeLimit :: Word64 -> Middleware
requestSizeLimit maxLen app req sendResponse = do
  putStrLn "In request size limit"
  case requestBodyLength req of
    KnownLength actualLen -> if actualLen > maxLen then
      sendResponse (tooLargeResponse maxLen actualLen)
      else app req sendResponse
    ChunkedBody -> do
        ref <- newIORef maxLen
        let newReq = req
                { requestBody = do
                    bs <- getRequestBodyChunk req
                    remaining <- readIORef ref
                    let len = fromIntegral $ S8.length bs
                        remaining' = remaining - len
                    if remaining < len
                        then do
                            putStrLn "Too large; sending too large response here"
                            throwIO (TooLargeRequestException $ tooLargeResponse maxLen len)
                            -- This won't work, because execution continues after sendResponse
                            sendResponse $ tooLargeResponse maxLen len
                            putStrLn "Too large; just sent too large response here"
                            pure ""
                        else do
                            writeIORef ref remaining'
                            return bs
                }
        app newReq sendResponse `catch` \(TooLargeRequestException resp) -> sendResponse resp
```

Ok, this seems to work. Is this pattern acceptable in WAI?

snoyberg: That's great! One request though: please take out the `putStrLn`, middlewares shouldn't generate output like that in general. The exception being thrown should be sufficient. If someone wants to generate a message, they can `onException` it. Alternatively: you could take a callback function to perform an action before throwing the exception.

Also, what's the point of the `sendResponse` and other two lines after the `throwIO` call?

MaxGabriel: Oh yeah, this isn’t the final code, this is the middle of me debugging stuff. I just wanted to check the exception approach was correct.

wai-extra/Network/Wai/Middleware/RequestSizeLimit.hs
```diff
@@ -0,0 +1,40 @@
...
+import qualified Data.ByteString.Lazy.Char8 as LS8
+import Control.Exception (try, catch)
+
+-- | Middleware to limit request bodies to a certain size. 
+-- If the body is too large, this middleware returns HTTP 413, with a plain text body describing the error.
+--
+-- This uses 'requestSizeCheck' under the hood; see that function for details.
+--
+-- @since 3.1.1
+requestSizeLimitMiddleware :: Word64 -> Middleware
```
MaxGabriel, line 17: Thoughts on if this should take some optional function to allow a custom error? The main use case I can think of is if someone wanted to e.g. do a JSON or XML response, or a different status code for some reason.

wai-extra/Network/Wai/Middleware/RequestSizeLimit.hs
```diff
@@ -0,0 +1,40 @@
...
+    let sendTooLargeResponse = sendResponse (tooLargeResponse maxLen (requestBodyLength req))
+
+    eitherSizeExceptionOrNewReq <- try (requestSizeCheck maxLen req)
+    case eitherSizeExceptionOrNewReq of
+        -- In the case of a known-length request, RequestSizeException will be thrown immediately
+        Left (RequestSizeException _maxLen) -> sendTooLargeResponse
+        -- In the case of a chunked request (unknown length), RequestSizeException will be thrown during the processing of a body
+        Right newReq -> app newReq sendResponse `catch` \(RequestSizeException _maxLen) -> sendTooLargeResponse
+
+tooLargeResponse :: Word64 -> RequestBodyLength -> Response
```
MaxGabriel, line 28: This error message is largely (but not exactly) copied from Yesod's message

wai/Network/Wai.hs
```diff
@@ -315,11 +318,62 @@ ifRequest :: (Request -> Bool) -> Middleware -> Middleware
...
+-- If certain endpoints need larger bodies, whitelist just those endpoints for the large size.
+-- Be especially cautious about endpoints that don't require authentication, since these are easier to DOS.
+-- You can accomplish this with @wai-extra@'s @requestSizeLimitMiddleware@ or Yesod's @maximumContentLength@.
```
MaxGabriel, line 358: I couldn't find a similar thing for Servant, but I'm not very familiar with it

snoyberg:
Looks like there's a build error on lts-11 due to missing `<>`.

wai-extra/Network/Wai/Middleware/RequestSizeLimit.hs
snoyberg: Yeah, it probably should. The most future-proofed way of doing this would be with the builder pattern: have a `RequestSizeLimitSettings` type, a `mkRequestSizeLimitSettings :: Word64 -> RequestSizeLimitSettings`, `requestSizeLimitMiddleware :: RequestSizeLimitSettings -> Middleware`, and then functions like `setErrorResponse :: Response -> RequestSizeLimitSettings -> RequestSizeLimitSettings`.

snoyberg (commented):
LGTM, only issues is the build failure for lts-11. Your call on whether my builder comment is worth pursuing. Thanks!

wai-extra/wai-extra.cabal
```diff
@@ -206,6 +208,7 @@ test-suite spec
                    , time
                    , case-insensitive
                    , http2
+                   , aeson
```
MaxGabriel, line 211: This is already a dependency of the library, so this isn't adding a new dependency

wai-extra/Network/Wai/Middleware/RequestSizeLimit/Internal.hs
```diff
@@ -0,0 +1,42 @@
...
+-- ===== JSON response
+-- > {-# LANGUAGE OverloadedStrings #-}
+-- > import Network.Wai
+-- > import Network.Wai.Middleware.RequestSizeLimit
+-- > import Network.HTTP.Types.Status (requestEntityTooLarge413)
+-- > import Data.Aeson
+-- > import Data.Text (Text)
+-- >
+-- > let jsonResponse = \_maxLen _app _req sendResponse -> sendResponse $ responseLBS requestEntityTooLarge413 [("Content-Type", "application/json")] (encode $ object ["error" .= ("request size too large" :: Text)])
+-- > let finalSettings = defaultRequestSizeLimitSettings { onLengthExceeded = jsonResponse }
```
MaxGabriel, line 36: I tested these examples in GHCI

wai-extra/Network/Wai/Middleware/RequestSizeLimit/Internal.hs
```diff
@@ -0,0 +1,42 @@
...
+--
+-- @since 3.1.1
+data RequestSizeLimitSettings = RequestSizeLimitSettings
+    { maxLengthForRequest :: Request -> IO (Maybe Word64) -- ^ Function to determine the maximum request size in bytes for the request. Return 'Nothing' for no limit. Since 3.1.1
```
MaxGabriel, line 40: I realized that before I was only taking a static Word64 parameter, but this would be limiting because it'd be only a single value. Now this is updated to mirror Yesod's interface, which allows constructing the limit dynamically based on the request.

wai-extra/Network/Wai/Middleware/RequestSizeLimit.hs
```diff
@@ -0,0 +1,71 @@
+-- | The functions in this module allow you to limit the total size of incoming request bodies.
```
MaxGabriel, line 1: Rendered haddocks:

image

wai-extra/Network/Wai/Middleware/RequestSizeLimit/Internal.hs
```diff
@@ -0,0 +1,42 @@
+-- | Internal constructors and helper functions. Note that no guarantees are given for stability of these interfaces.
```
MaxGabriel, line 1: Rendered haddocks:

image

MaxGabriel:
Ok, I think this is ready for re-review!

wai-extra/Network/Wai/Middleware/RequestSizeLimit/Internal.hs
snoyberg: This is a good improvement. But I would recommend against exposing these accessors directly. It ends up tying our hands with API changes in the future. I'd recommend only exposing explicit setter functions.

snoyberg (changes requested):
Overall looks great, only one minor comment. Thank you for all the time spent on this PR!

wai-extra/Network/Wai/Middleware/RequestSizeLimit/Internal.hs
MaxGabriel: Ah ok. I've exposed just explicit setter functions now :+1:

MaxGabriel:
Ok, think this is ready for re-review!

snoyberg (approved):
This is great, thanks!

PR #826, opened 2020-11-25
Request logger detailed settings

BrechtSerckx:

After submitting your PR:

This PR solves #774 in combination with a way to hide health-check logs, adapting the solution of #776.
The request filter can be generalized to the other output formats, but that was out of scope for our use case and would be more invasive.


snoyberg (approved):
LGTM, thanks!

PR #827, opened 2020-12-03
Add requestToJSON to API, for external use

ivanbakel:
This adds `requestToJSON` to the API, since I want to access WAI's representation of requests as JSON for [`yesod-katip`](https://github.com/ivanbakel/yesod-katip), a logging library.

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RequestLogger/JSON.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger/JSON.hs
@@ -4,12 +4,15 @@
 module Network.Wai.Middleware.RequestLogger.JSON
   ( formatAsJSON
   , formatAsJSONWithHeaders
+
+  , requestToJSON
   ) where
 
 import qualified Data.ByteString.Builder as BB (toLazyByteString)
 import Data.ByteString.Lazy (toStrict)
 import Data.Aeson
 import Data.CaseInsensitive (original)
+import Data.Maybe (maybeToList)
 import Data.Monoid ((<>))
 import qualified Data.ByteString.Char8 as S8
 import Data.IP
@@ -30,7 +33,7 @@ formatAsJSON :: OutputFormatterWithDetails
 formatAsJSON date req status responseSize duration reqBody response =
   toLogStr (encode $
     object
-      [ "request"  .= requestToJSON duration req reqBody
+      [ "request"  .= requestToJSON req reqBody (Just duration)
       , "response" .=
       object
         [ "status" .= statusCode status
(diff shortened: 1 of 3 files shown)


wai-extra/Network/Wai/Middleware/RequestLogger/JSON.hs
```diff
@@ -73,19 +76,26 @@ word32ToHostAddress = T.intercalate "." . map (T.pack . show) . fromIPv4 . fromH
...
-requestToJSON :: NominalDiffTime -> Request -> [S8.ByteString] -> Value
-requestToJSON duration req reqBody =
-  object
+-- | Get the JSON output for a request
+--
+-- This gives a sensible JSON representation for the given request. You can
+-- optionally include a request body, as well as a duration.
+--
+-- @since 3.1.4
+requestToJSON ::  Request -> [S8.ByteString] -> Maybe NominalDiffTime -> Value
```
snoyberg, line 85: Some Haddocks on these parameters to explain what these parameters are for would be helpful

wai-extra/Network/Wai/Middleware/RequestLogger/JSON.hs
```diff
@@ -73,19 +76,26 @@ word32ToHostAddress = T.intercalate "." . map (T.pack . show) . fromIPv4 . fromH
 readAsDouble :: String -> Double
 readAsDouble = read
 
-requestToJSON :: NominalDiffTime -> Request -> [S8.ByteString] -> Value
-requestToJSON duration req reqBody =
-  object
+-- | Get the JSON output for a request
```
snoyberg, line 79: This description isn't clear enough from my reading. I'm not sure what a "JSON output for a request" means. I think "JSON representation" is closer to the mark, but a little more clarity around this would definitely be helpful.

snoyberg (changes requested):

wai-extra/Network/Wai/Middleware/RequestLogger/JSON.hs
ivanbakel: I was struggling to describe this myself. I'll opt for "JSON representation" - but how would you describe *how* the representation is defined? I didn't want the doc comment to imply some kind of API for serialisation that users might then depend on.

snoyberg: I've then spell that out. Clarify what _kind_ of information it includes (like method etc) but that the exact representation may change over time.

ivanbakel: Let me know what you think. I tried to be clear about the structure without being too verbose - there didn't seem to be a need to explain what most of the object fields did, for example.

snoyberg (approved):
Looks great!

PR #661, opened 2017-12-07
Add Vary: Accept-Encoding in Gzip middleware

mithrandi:
Before submitting your PR, check that you've:

- [x] Bumped the version number

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

Closes #627.

Better late than never? I'm not sure if this is the best way to manipulate the response header fields, let me know if there's a better way.

Issue #627, opened 2017-06-14
Gzip middleware does not set Vary: Accept-Encoding


mithrandi:
I think this should be set for all responses passing through the middleware, because they might be gzipped if the client supplies `Accept-Encoding: gzip`.

snoyberg:
Seems reasonable. Want to send a PR?

On Wed, Jun 14, 2017 at 2:02 PM, Tristan Seligmann <notifications@github.com

snoyberg:
Implemented elsewhere, closing

PR #829, opened 2020-12-12
Add Vary header in gzip middleware

pdedmon:
Resurrecting #661 to address #627.

After submitting your PR:

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for wai-extra
 
+## 3.1.4.1
+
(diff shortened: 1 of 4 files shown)


snoyberg:
I don’t think this code handles the case when the underlying application itself returns a vary response header

pdedmon:
I did think about that case, and this code will currently result in multiple vary headers, which should be equivalent to one header with comma separated values, e.g.
```
Vary: Accept-Encoding
Vary: User-Agent
```
vs.
```
Vary: Accept-Encoding, User-Agent
```
I can instead append/prepend to the existing header if that's preferable.

snoyberg:
Good point about the multiple headers, this looks good to me. Thank you!

PR #834, opened 2020-12-31
add real ip middleware

pdedmon:
Add middleware for finding the "real" client IP address from headers (cf. [nginx's realip module](https://nginx.org/en/docs/http/ngx_http_realip_module.html) and [rails' remote IP middleware](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html), among others).

One potential issue with this middleware's approach is that the original remote host address is overwritten, and is therefore unavailable to middleware further down the stack, as well as the application itself. I don't expect this to be an issue for users of this middleware, but we could put the inferred IP address in a header instead of overwriting the remoteHost on the Request.

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RealIp.hs
+++ b/wai-extra/Network/Wai/Middleware/RealIp.hs
@@ -1,0 +1,79 @@
+-- | Infer the remote IP address using headers
+module Network.Wai.Middleware.RealIp
+    ( realIp
+    , realIpHeader
+    , realIpTrusted
+    , defaultTrusted
+    ) where
+
+import qualified Data.ByteString.Char8 as B8 (unpack, split)
+import qualified Data.IP as IP
+import Data.Maybe (fromMaybe, mapMaybe, listToMaybe)
+import Network.HTTP.Types (HeaderName, RequestHeaders)
+import Network.Wai
+import Text.Read (readMaybe)
+
+-- | Infer the remote IP address from the @X-Forwarded-For@ header,
+-- trusting requests from any private IP address. See 'realIpHeader' and
+-- 'realIpTrusted' for more information and options.
+--
+-- @since 3.1.5
+realIp :: Middleware
+realIp = realIpHeader "X-Forwarded-For"
+
+-- | Infer the remote IP address using the given header, trusting
+-- requests from any private IP address. See 'realIpTrusted' for more
+-- information and options.
+--
+-- @since 3.1.5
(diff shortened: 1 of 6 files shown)


wai-extra/Network/Wai/Middleware/RealIp.hs
```diff
@@ -0,0 +1,79 @@
...
+-- | Infer the remote IP address using the given header, but only if the
+-- request came from an IP in one of the trusted ranges. The last
+-- non-trusted address is used to replace the 'remoteHost' in the
+-- 'Request', unless all present IP addresses are trusted, in which case
+-- the first address is used. Invalid IP addresses are ignored, and the
+-- remoteHost value remains unaltered if no valid IP addresses are
+-- found.
+--
+-- @since 3.1.5
+realIpTrusted :: HeaderName -> [IP.IPRange] -> Middleware
```
snoyberg, line 41: Just a thought: would it be more flexible to provide a predicate argument here to check the IP address, with the current `ipInRange` as a helper function in this module?

snoyberg (commented):
Looks great! I've added one comment. To your point about overwriting the IP address: one possibility would be to add the original address into the `vault`. I don't feel strongly about this.

wai-extra/Network/Wai/Middleware/RealIp.hs
pdedmon: That's a good idea. So we would have
```
realIpTrusted :: HeaderName -> (IP.IP -> Bool) -> Middleware
```
with an example implementation being
```
realIpTrusted "X-Forwarded-For" (\ip -> any (ipInRange ip) defaultTrusted)
```

(I'll see if I can rework things to avoid the lambda.)

pdedmon:
Using the vault to store the original remote host could work, but I think I'll leave that for now and see if it's ever actually needed.

I noticed that the lts-11 and lts-12 builds failed because the iproute package is not new enough for these changes - I'm not sure what to do about that.

snoyberg:
> I noticed that the lts-11 and lts-12 builds failed because the iproute package is not new enough for these changes - I'm not sure what to do about that.

Please add an appropriate lower bound in the cabal file and then add the correct extra-dep to stack-lts-11.yaml and stack-lts-12.yaml.

pdedmon:
I think this is ready to merge now.

snoyberg (approved):
Thanks!

PR #836, opened 2021-01-12
UTF-8 encoding in exceptionResponseForDebug

vshabanov:
`exceptionResponseForDebug` truncates any non-ASCII characters in exception error message due to packing error `String` as 8-bit `ByteString`. This PR fixes it.

---

After submitting your PR:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Settings.hs
+++ b/warp/Network/Wai/Handler/Warp/Settings.hs
@@ -11,6 +11,7 @@ import qualified Data.ByteString.Char8 as C8
 import Data.ByteString.Lazy (fromStrict)
 import Data.Streaming.Network (HostPreference)
 import qualified Data.Text as T
+import qualified Data.Text.Encoding as TE
 import qualified Data.Text.IO as TIO
 import Data.Version (showVersion)
 import GHC.IO.Exception (IOErrorType(..))
@@ -235,4 +236,4 @@ exceptionResponseForDebug :: SomeException -> Response
 exceptionResponseForDebug e =
     responseBuilder H.internalServerError500
                     [(H.hContentType, "text/plain; charset=utf-8")]
-                    $ byteString . C8.pack $ "Exception: " ++ show e
+                    $ byteString . TE.encodeUtf8 . T.pack $ "Exception: " ++ show e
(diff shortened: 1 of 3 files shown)


warp/Network/Wai/Handler/Warp/Settings.hs
```diff
@@ -235,4 +236,4 @@ exceptionResponseForDebug :: SomeException -> Response
 exceptionResponseForDebug e =
     responseBuilder H.internalServerError500
                     [(H.hContentType, "text/plain; charset=utf-8")]
-                    $ byteString . C8.pack $ "Exception: " ++ show e
+                    $ byteString . TE.encodeUtf8 . T.pack $ "Exception: " ++ show e
```
snoyberg, line 239: Good catch, nice improvement! How about using [`encodeUtf8Builder`](https://www.stackage.org/haddock/lts-16.28/text-1.2.4.0/Data-Text-Encoding.html#v:encodeUtf8Builder) instead?

vshabanov: Sure, changed to `encodeUtf8Builder`.

snoyberg (commented):
Thanks!

snoyberg:
Actually, looking more closely, I realized we can bypass the `Text` entirely using [`stringUtf8`](https://www.stackage.org/haddock/nightly-2021-01-13/bytestring-0.10.12.0/Data-ByteString-Builder.html#v:stringUtf8). I don't want to make you go through another iteration, so I'll make that change while merging. Thanks again.

vshabanov:
Unlike `T.pack` `stringUtf8` does not do validation and [can return invalid UTF-8](https://hackage.haskell.org/package/bytestring-0.11.0.0/docs/Data-ByteString-Builder.html#v:stringUtf8):

```
> TE.decodeUtf8 $ toStrict $ toLazyByteString $ stringUtf8 "\xD800"
*** Exception: Cannot decode byte '\xed': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```

```
> TE.decodeUtf8 $ toStrict $ toLazyByteString $ TE.encodeUtf8Builder $ T.pack "\xD800"
"\65533"
```

So maybe it's better to keep `TE.encodeUtf8Builder . T.pack`

snoyberg:
Seems like a reasonable garbage in, garbage out situation to be

PR #837, opened 2021-01-17
remove unused dependencies in wai-extra

pdedmon:

After submitting your PR:

I noticed a few dependencies that were specified but never used. I also noticed that the directory package can be used to get the size of a file, but only if the lower bound is increased; I think this should be alright, but let me know if not. Does this kind of change require a version bump?

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for wai-extra
 
+## 3.1.6
+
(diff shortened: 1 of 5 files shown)


snoyberg:
This looks great, thanks! Yes, a minor version bump and ChangeLog entry would be great too.

snoyberg (approved):

snoyberg:
Thanks!

PR #844, opened 2021-05-17
Cleanly close connection when client closes connection prematurely

larskuhtz:
The function `Network.Socket.ByteString.recv` returns a `ByteString` of length zero when the connection has been closed.
The `mkConn` function does not handle that special case and instead calls `plainHTTP`, which raises a spurious `InsecureConnectionDemied` Exception. On the client side http-client-tls (and probably most other clients) reacts to this with an ungraceful connection reset.

This PR adds a new `ClientClosedConnectionPrematurely` constructor to `WarpTLSException`. When the client closes a connection immediately after the opening handshake and before the TLS "hello" message, the connection is gracefully closed and a `ClientClosedConnectionPrematurely` exception is raised.

After submitting your PR:

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -393,14 +393,12 @@ getter tlsset set sock params = do
     return (mkConn tlsset set s params, sa)
 
 mkConn :: TLS.TLSParams params => TLSSettings -> Settings -> Socket -> params -> IO (Connection, Transport)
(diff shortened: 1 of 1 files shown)


snoyberg:
@kazu-yamamoto this looks reasonable to me, but I think you'd be in a better position to review.

kazu-yamamoto (approved):
Looks excellent to me!

kazu-yamamoto:

kazu-yamamoto:
A new version of warp-tls has been released.

PR #847, opened 2021-05-21
845 async exceptions

snoyberg:
Before submitting your PR, check that you've:

- [X] Bumped the version number
- [X] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [X] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [ ] Update the Changelog.md file with a link to your PR
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #848, opened 2021-05-24
using unliftio more.

kazu-yamamoto:
@snoyberg This PR replaces `Control.Exception` and `Control.Concurrent.Async` which were not replaced in #847. Please review.

They are still left. Ideally we should replace all.


snoyberg (approved):
Awesome!

Issue #850, opened 2021-06-18
Thread killed by timeout manager


chrisdone:
With this file
```
stack runhaskell wrp.hs --resolver lts-18.0
```
```haskell
{-# LANGUAGE TypeFamilies, QuasiQuotes, MultiParamTypeClasses,
             TemplateHaskell, OverloadedStrings #-}
import           Control.Concurrent
import qualified Text.Blaze.Html5 as H
import qualified Data.Text as T
import           Data.Text (Text)
import           Control.Monad
import           Yesod
import           Yesod.WebSockets
data Piggies = Piggies
instance Yesod Piggies
mkYesod "Piggies" [parseRoutes|
  / HomeR GET
|]
getHomeR :: Handler Html
getHomeR = do webSockets (forever (do liftIO (threadDelay (1000 * 1000))
                                      sendTextData ("Bump!" :: Text)))
              pure (H.script "// Create WebSocket connection.\n\
                             \const socket = new WebSocket('ws://localhost:3035');\n\
                             \\n\
                             \// Connection opened\n\
                             \socket.addEventListener('open', function (event) {\n\
                             \    socket.send('Hello Server!');\n\
                             \});\n\
                             \\n\
                             \// Listen for messages\n\
                             \socket.addEventListener('message', function (event) {\n\
                             \    console.log('Message from server ', event.data); socket.close()\n\
                             \});")
main = warpDebug 3035 Piggies
```

I see this output

```
$ stack runhaskell wrp.hs --resolver lts-18.0

wrp.hs:30:8: warning: [-Wdeprecations]
    In the use of ‘warpDebug’
    (imported from Yesod, but defined in Yesod.Core.Dispatch):
    Deprecated: "Please use warp instead"
   |
30 | main = warpDebug 3035 Piggies
   |        ^^^^^^^^^
18/Jun/2021:17:38:23 +0100 [Info#yesod-core] Application launched @(yesod-core-1.6.20.1-Jp7zmLxTY8gEM32OuxAOUR:Yesod.Core.Dispatch src/Yesod/Core/Dispatch.hs:175:11)
  C-c C-c
```

```
 $ stack runhaskell wrp.hs --resolver lts-18.0

wrp.hs:30:8: warning: [-Wdeprecations]
    In the use of ‘warpDebug’
    (imported from Yesod, but defined in Yesod.Core.Dispatch):
    Deprecated: "Please use warp instead"
   |
30 | main = warpDebug 3035 Piggies
[...]

PR #851, opened 2021-06-20
Swallow async exceptions in forked thread #850

snoyberg:

After submitting your PR:

* * *

CC @chrisdone @kazu-yamamoto @alaendle I think this PR fixes the issues described. There are two commits: the first one implements the actual fix (catching all exceptions in the forked thread). The second implements a minor optimization revealed by these reports, turning off the default exception handling in the forked thread, since Warp's exception handler must catch all exceptions.

Confirmation that this fixes the reported issues would be great, I can no longer repro Chris's problem with this change.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         args:
-        - "--resolver nightly"
+        # Missing some packages now
+        #- "--resolver nightly"
         - "--resolver lts-17"
         - "--resolver lts-16"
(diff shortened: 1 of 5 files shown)


chrisdone:
Looks good. I don't reproduce the issue anymore with the codebase I was using. :+1:  

Thanks!

alaendle:
Agree - with this fork I couldn't reproduce my problem anymore. Many thanks!

snoyberg:
@kazu-yamamoto you OK with me merging/releasing?

kazu-yamamoto (approved):
Excellent. I will release a new version.

kazu-yamamoto:
I have released a new version. Thank you all for your contribution!

PR #857, opened 2021-10-19
Option to log requests before application is called in wai-extra RequestLogging Middleware

sordina:

After submitting your PR:

### DIFF
--- a/wai-extra/Network/Wai/Middleware/RequestLogger.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger.hs
@@ -36,6 +36,7 @@ import Network.Wai
 import System.Log.FastLogger
 import Network.HTTP.Types as H
 import Data.Maybe (fromMaybe, isJust, mapMaybe)
+import Data.Monoid (mconcat, (<>))
 import Data.Time (getCurrentTime, diffUTCTime, NominalDiffTime, UTCTime)
 import Network.Wai.Parse (sinkRequestBody, lbsBackEnd, fileName, Param, File
                          , getRequestBodyType)
--- a/wai-extra/Network/Wai/Middleware/RequestLogger.hs
+++ b/wai-extra/Network/Wai/Middleware/RequestLogger.hs
@@ -1,10 +1,12 @@
 {-# LANGUAGE RecordWildCards #-}
+
 -- NOTE: Due to https://github.com/yesodweb/wai/issues/192, this module should
 -- not use CPP.
 module Network.Wai.Middleware.RequestLogger
     ( -- * Basic stdout logging
       logStdout
     , logStdoutDev
+    , logStdoutDevWithRequestPrelogging
       -- * Create more versions
     , mkRequestLogger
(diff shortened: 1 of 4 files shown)


wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -360,6 +372,15 @@ detailedMiddleware' cb DetailedSettings{..} ansiColor ansiMethod ansiStatusCode
               in if null par then [""] else ansiColor White "  Params: " <> par <> ["\n"]
 
   t0 <- getCurrentTime
+
+  -- Optionally prelog the request
+  when mPrelogRequests $ do
+    cb $ mconcat $ map toLogStr $
```
snoyberg, line 378: Instead of using `mconcat` like this, I'd rather see something that leverages the `Semigroup` instance, which should be more efficient. Something like: `cb $ "PRELOGGING REQUEST: " <> toLogStr (ansiMethod ...) <> " " <> toLogStr (rawPathInfo req) <> ...`

wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -234,6 +238,14 @@ logStdout = unsafePerformIO $ mkRequestLogger def { outputFormat = Apache FromSo
 logStdoutDev :: Middleware
 logStdoutDev = unsafePerformIO $ mkRequestLogger def
 
+-- | Development request logger middleware.
+--
+-- This uses the 'Detailed' 'True' logging format and logs to 'stdout'.
+-- @since 3.1.7
+{-# NOINLINE logStdoutDevWithRequestPrelogging #-}
+logStdoutDevWithRequestPrelogging :: Middleware
```
snoyberg, line 246: I'm not sure it's worth adding in a new value here, versus letting people use `mkRequestLogger` directly for this case.

sordina: Fair enough! I'll remove it. I guess I'd like it to be very convenient to handle this scenario, but on second thought this is mainly due to the situation where logging is cancelled due to an exception.

snoyberg: I should point out: the exception scenario is almost certainly a mistake. I know at least Servant mistakenly does not handle exceptions and propagates them up to Warp for handling, which is wrong. You may be encountering that issue here. If so, I'd recommend wrapping the application in an exception handler. Warp is ill equipped to handle exceptions, and middleware ideally shouldn't have to handle those cases at all.

sordina: Perhaps the best solution would be a new exception handling middleware?

snoyberg: I'm overall -1 on this, since it really should be handled in the `Application` itself. The code in question here doesn't handle some painful corner cases such as "what if the exception handler throws an exception?"

PR #858, opened 2021-10-20
Document the Middleware type alias

Profpatsch:
I tried to come at this from the angle of

1) A user of Wai trying to understand what `Middleware` is
   and how to use it, especially what happens if you chain
   multiple middlewares and in what order their are called.

2) A developer trying to implement new middleware, how to
   modify requests and responses, and how to pass on data.

by giving examples of how the type signatures and
simple usage/implementations look.

I think `Wai` is a super fundamental module for the Haskell
ecosystem, so we should aim for the documentation to be top
notch (think: rust stdlib).

I think the rest of the module needs a simple overhaul,
but I was especially annoyed by having to reconstruct Middleware
from first principles every time I built a server (“which order
does it chain again?”).

After submitting your PR:

### DIFF
--- a/wai/Network/Wai.hs
+++ b/wai/Network/Wai.hs
@@ -291,20 +291,149 @@ defaultRequest = Request
     }
 
 
--- | Middleware is a component that sits between the server and application. It
--- can do such tasks as GZIP encoding or response caching. What follows is the
--- general definition of middleware, though a middleware author should feel
--- free to modify this.
+-- | A @Middleware@ is a component that sits between the server and application.
 --
--- As an example of an alternate type for middleware, suppose you write a
--- function to load up session information. The session information is simply a
--- string map \[(String, String)\]. A logical type signature for this middleware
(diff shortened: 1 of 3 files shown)


wai/Network/Wai.hs
```diff
@@ -291,20 +291,149 @@ defaultRequest = Request
...
+-- @
+--
+-- Notice how the inner app is called when the validation was successful.
+-- If it was not, we can respond
+-- e.g. with <https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401 HTTP 401 Unauthorized>,
+-- by constructing a 'Response' with 'responseLBS' and passing it to @respond@.
+--
+-- == Passing arguments to and from your 'Middleware'
+--
+-- Usually, the middleware should be configurable somehow.
```
snoyberg, line 380:
```suggestion
-- Middleware must often be configurable.
```

PR #860, opened 2021-11-02
Export Network.Wai.Middleware.Gzip.GzipSettings.

DataKinds:
Export `Network.Wai.Middleware.Gzip.GzipSettings`. 

- This is such a small change that I did not bump the version number, labeling it "Unreleased" in the changelog with accordance to the policy. 
- Documentation updated.

TO-DO: 

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for wai-extra
 
+# Unreleased 
+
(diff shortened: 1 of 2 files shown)


snoyberg:
We normally avoid exposing data constructors for strings types so the settings can be extended in the future without breaking with public API.

Vlix:
I feel that now with better documentation about how to adjust the `GzipSettings` (using `def` and then altering the fields), this issue can be closed, right?

PR #866, opened 2022-01-01
wai-extra/RequestLogger: Add ApacheWithSettings OutputFormat to allow request filtering and logging the current user

wolfgangwalther:

---

wai-logger 2.3.7 has `initLoggerUser` now (https://github.com/kazu-yamamoto/logger/pull/198), which is exposed in this PR via a new `ApacheWithSettings`. This also copies the `mFilterRequests` from `DetailedWithSettings`.


wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE DuplicateRecordFields #-}
```
wolfgangwalther, line 1: I added this because `mFilterRequests` is used in both `DetailedWithSettings` and `ApacheWithSettings`. I noticed that I needed to add the same pragma to my own code using `RequestLogger` - wasn't able to figure out how to import only `ApacheWithSettings`, yet.

So I guess this is a breaking change for everyone importing the whole module and not using this pragma so far.

Any better idea?

wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -9,11 +10,13 @@ module Network.Wai.Middleware.RequestLogger
       -- * Create more versions
     , mkRequestLogger
     , RequestLoggerSettings
+    , def
```
wolfgangwalther, line 13: I noticed this is exported in the `Gzip` module so far. Exporting it here could break some imports that import both `Gzip` and `RequestLogger` at the same time. Importing from `Gzip` and using it for `RequestLogger` seems... wrong, though :D

Not sure, however, to deal best with this?

wai-extra/Network/Wai/Middleware/RequestLogger.hs
snoyberg: Don't use the same field names, use something else instead. This is not a change I'd merge.

wai-extra/Network/Wai/Middleware/RequestLogger.hs
snoyberg: Don't use the `Default` typeclass. Instead, expose a new identifier such as `defaultApacheSettings`.

wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -9,11 +10,13 @@ module Network.Wai.Middleware.RequestLogger
...
     , outputFormat
     , autoFlush
     , destination
     , OutputFormat (..)
-    , DetailedSettings(..)
+    , ApacheSettings (..)
```
snoyberg, line 18: For new identifiers, I would recommend _not_ exposing the fields and data constructor directly, but instead exposing the type, a `default` value, and setter functions, to allow easier extension in the future.

snoyberg (changes requested):

wai-extra/Network/Wai/Middleware/RequestLogger.hs
wolfgangwalther: Ok, I took inspiration from `Network.Wai.Handler.Warp` and its `Settings`. Much better indeed. WDYT?

wolfgangwalther:
To be able to use `ApacheWithSettings` with `mkRequestLogger` I still needed `Data.Default` for `RequestLoggerSettings`. I added another export `defaultRequestLoggerSettings` to be able to use the same style consistently in that case.

wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -51,16 +57,65 @@ import Network.Wai.Logger
...
+data ApacheSettings = ApacheSettings
+    { apacheIPAddrSource :: IPAddrSource
+    , apacheUserGetter :: Request -> Maybe BS.ByteString
+    , apacheRequestFilter :: Request -> Response -> Bool
+    }
+
+defaultApacheSettings :: ApacheSettings
+defaultApacheSettings = ApacheSettings
+    { apacheIPAddrSource = FromSocket
+    , apacheRequestFilter = \_ -> \_ -> True
```
snoyberg, line 89: Would normally be `\_ _ -> True` IMO, but not really an issue.

wai-extra/Network/Wai/Middleware/RequestLogger.hs
```diff
@@ -141,12 +196,15 @@ data RequestLoggerSettings = RequestLoggerSettings
...
+    , destination = Handle stdout
+    }
+
 instance Default RequestLoggerSettings where
-    def = RequestLoggerSettings
-        { outputFormat = Detailed True
-        , autoFlush = True
-        , destination = Handle stdout
-        }
+    def = defaultRequestLoggerSettings
```
snoyberg, line 207: Yup, I like this approach.

snoyberg (approved):
LGTM!

snoyberg:
Everything looks good, but the CI job is failing due to the tighter lower bounds on wai-logger.

wolfgangwalther:
> Everything looks good, but the CI job is failing due to the tighter lower bounds on wai-logger.

Not sure about the process - do I need to do / can I do anything there?

wai-extra/Network/Wai/Middleware/RequestLogger.hs
wolfgangwalther: I'll fix that.

snoyberg:
You'd need to modify the stack.yaml files to include extra-deps with the relevant deps.

wolfgangwalther:
> You'd need to modify the stack.yaml files to include extra-deps with the relevant deps.

Ok, I did so. Unfortunately, I can't even make `stack build` or `stack build --resolver=lts-14` run locally on the master branch, without the changes in this PR. Only `stack build --resolver=nightly` works for me.

Let's see what CI says?

snoyberg:
Thanks!

PR #831, opened 2020-12-17
allowing eMFILE in acceptNewConnection.

kazu-yamamoto:
Should fix #830.

PR #870, opened 2022-02-01
allowing eMFILE in acceptNewConnection

ProofOfKeags:
This is a duplicate of #831. I have fast forwarded it to be based off of the current latest warp code. I have also tested that it solves the problem it sets out to. While I cannot attest to the fact that this is the *right* way to solve this problem, I figured that since @kazu-yamamoto is the original author of the patch, that it is sound in context. I am now attesting that it fixes things.


kazu-yamamoto:
Sorry but my credit is lost in this commit. So, I will merge #831.

ProofOfKeags:
Perfectly OK. Did not intend to remove credit, and I in no way improved on your approach, so no hard feelings whatsoever. I'm just glad it's fixed now ^_^

PR #871, opened 2022-02-24
Add flags to avoid Cryptonite/X509

ysangkok:

Many warp users use a reverse proxy for TLS termination. But before this PR,
the warp library had an unconditional dependency on the x509 library. The x509
library has a big dependency footprint, which would be irrelevant for these
users. Another problem with the x509 dependency is that it depends on
Cryptonite, which is updated slowly, and is still not available for GHC 9.2.

The wai-app-static library also depends on Cryptonite, but only for computing
MD5 hashes. It seems excessive to depend on this very large library for just a
single hash function. So this PR adds a flag to enable compilation with a
different, also very well-trusted library.

Because both of these newly added flags are enabled by default, it means that a
user that isn't carefully reading the warp/wai-app-static release notes will
not notice any disruption. People that try to upgrade to GHC 9.2 after this PR
is merged may notice that the client certificates are no longer available. But
that is not so bad, since it only happens while upgrading their compiler, which
commonly introduces all kinds of breaking changes. So I don't consider it a big
detraction.

Tested using the following `cabal.project`: `packages: wai-app-static warp`

* `cabal test all -f -cryptonite -f -x509 -w ghc-9.2.1`
* `cabal test all -f cryptonite -f x509 -w ghc-9.0.2`

### DIFF
--- a/wai-app-static/WaiAppStatic/Storage/Embedded/Runtime.hs
+++ b/wai-app-static/WaiAppStatic/Storage/Embedded/Runtime.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 -- | Lookup files stored in memory instead of from the filesystem.
 module WaiAppStatic.Storage.Embedded.Runtime
     ( -- * Settings
@@ -15,8 +16,13 @@ import Data.Function (on)
 import qualified Data.Text as T
 import Data.Ord
 import qualified Data.ByteString as S
+#ifdef MIN_VERSION_cryptonite
 import Crypto.Hash (hash, MD5, Digest)
 import Data.ByteArray.Encoding
+#else
+import Crypto.Hash.MD5 (hash)
+import Data.ByteString.Base64 (encode)
+#endif
 import WaiAppStatic.Storage.Filesystem (defaultFileServerSettings)
 import System.FilePath (isPathSeparator)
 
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
I understand your pain. I have no objection. To my quick check, the code itself is high quality.
@snoyberg What do you think?

wai-app-static/WaiAppStatic/Storage/Embedded/Runtime.hs
```diff
@@ -15,8 +16,13 @@ import Data.Function (on)
 import qualified Data.Text as T
 import Data.Ord
 import qualified Data.ByteString as S
+#ifdef USE_CRYPTONITE
```
snoyberg, line 19: I think you could use the build in Cabal macros for this instead. There's definitely `MIN_VERSION_cryptonite`, but I think there may be others too.

snoyberg (commented):
I added a minor comment about the CPP variable, but otherwise have no issue, the motivation here makes perfect sense.

wai-app-static/WaiAppStatic/Storage/Embedded/Runtime.hs
ysangkok: `MIN_VERSION_cryptonite` seems to work perfectly here, so I have force-pushed revised commits that use this instead. Thank you for the review.

snoyberg (approved):
LGTM. I'll defer to @kazu-yamamoto on merging. Thanks!

kazu-yamamoto:
Merged. Thank you for your contribution.

ysangkok:
@kazu-yamamoto Thank you for the merge. Yes, a release would be very nice, since that would allow me to run CI for 9.2 for servant-multipart and servant-openapi3 without using `source-repository-package`.

PR #872, opened 2022-03-04
Expose `TLS.supportedHashSignatures` via TLSSettings

miguelclean:
### Proposal 

I noticed that with **TLS1.2** the **Client Certifiicate Types** provided in the **Client Certificate Request** generated during handskake, are derived from **TLS.supportedHashSingatures** (see code snipplet below)

I do not see any way to control this value, when using **runTLS** so propose to expose it via **TLSSettings**.

_excerpt from Network.TLS.Handshake.Server:_
```
(certTypes, hashSigs)
                        | usedVersion < TLS12 = (defaultCertTypes, Nothing)
                        | otherwise =
                            let as = supportedHashSignatures $ ctxSupported ctx
                             in (nub $ mapMaybe hashSigToCertType as, Just as)
```

### Motivation

Some of the **Client Certificate Types** returned, using the default **TLS.supportedHashSignatures** seem not to be recognized by calling clients:

```
$ openssl s_client -tls1_2 -port 1234 -host localhost | grep "Client Certificate Types"
Client Certificate Types: UNKNOWN (0),, UNKNOWN (0),, ECDSA sign, RSA sign, DSA sign
```

We encountered instances of clients that seem unable to cope with that at all.

### DIFF
--- a/warp-tls/ChangeLog.md
+++ b/warp-tls/ChangeLog.md
@@ -1,3 +1,8 @@
+## 3.3.3
+
+* Expose TLS.supportedHashSignatures via TLSSettings
+  [#872](https://github.com/yesodweb/wai/pull/872)
+
 ## 3.3.2
 
 * Providing the Internal module.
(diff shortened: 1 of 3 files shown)


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:

kazu-yamamoto:

miguelclean:
Thanks. No need for new release.

PR #875, opened 2022-04-05
Linting and refactoring `wai-extra`

Vlix:
After submitting your PR:

----

## Review per commit

Refactored the gzip middleware to by DRY-er, added rethrowing of `Exception`s that can't come from this `Middleware`, tweaked adding of `Vary` header, made sure to always skip if response status is 206, and mainly:

### Added option to take the ETag header into account when caching zipped files

A few points why this is a WIP:
- I also removed the `Data.Default` dependency, probably as a relfex since it's useless (in my honest opinion), but I can revert that if necessary
- I'm not sure what kind of version bump this would be
- I still want to add some tests
- I want to add more documentation for the whole `Gzip` module
 
# EDIT: Changed this PR into just linting and refactoring

### DIFF
--- a/wai-extra/Network/Wai/EventSource.hs
+++ b/wai-extra/Network/Wai/EventSource.hs
@@ -9,17 +9,17 @@
     There is a small example using these functions in the @example@ directory.
 -}
 module Network.Wai.EventSource (
-    ServerEvent(..),
+    ServerEvent (..),
     eventSourceAppChan,
     eventSourceAppIO,
-    eventStreamAppRaw
-    ) where
+    eventStreamAppRaw,
+) where
 
-import           Data.Function (fix)
-import           Control.Concurrent.Chan (Chan, dupChan, readChan)
-import           Control.Monad.IO.Class (liftIO)
-import           Network.HTTP.Types (status200, hContentType)
-import           Network.Wai (Application, responseStream)
+import Control.Concurrent.Chan (Chan, dupChan, readChan)
+import Control.Monad.IO.Class (liftIO)
+import Data.Function (fix)
+import Network.HTTP.Types (hContentType, status200)
+import Network.Wai (Application, responseStream)
 
 import Network.Wai.EventSource.EventStream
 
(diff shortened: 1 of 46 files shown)


snoyberg:
Would it be possible to separate out general code cleanup/reformatting changes from the meat of the change here?

Vlix:
Just make two PRs you mean? Yeah, that should be possible.
Might close this one and make two new ones, we'll see.

snoyberg:
Yes, two PRs would be better, thanks!

Vlix:
Would you prefer me not to rip out the `Default` instance(s) for backwards compatibility? Or would it be ok te just get rid of `Data.Default` in its entirety? (This would probably be a somewhat breaking change 🤔 )

snoyberg:
I'd rather avoid any breaking changes.

Vlix:
Ok, removed all the functional changes, now this PR is just import linting, readability cleanup and refactors.

* [10377c3](https://github.com/yesodweb/wai/pull/875/commits/10377c3dc2dc6eb3b0b6cfa4faa4462827e58b59): mainly readability cleanup (redundant brackets, unnecessary `$`, avoid punning, etc)
* [c968a03](https://github.com/yesodweb/wai/pull/875/commits/c968a036248c281f09b4a6b4a5c2fec741e55b4e): **Most interesting commit; refactor of `gzip` middleware to keep it DRY, skip compression earlier, etc.**
* [25e5de5](https://github.com/yesodweb/wai/pull/875/commits/25e5de5b241ae6e63cf8292085d46847f4166679): linting imports and pragmas, more CPP around `import Applicative/Monoid` and some more readability cleanup

I also hanged two `data` into `newtype`, since they only had one constructor with one argument. I don't see how this could negatively impact anything, but if you do, I can revert those two.

Vlix:
The one failing test is on Windows, and it fails on 1 test in `warp` because of an `IOException` where `sendbuf()` gets a wrong argument (I'm guessing the implementation is too Linux specific?)

Vlix:
I don't feel this warrants an addition to the Changelog, right? So I feel this PR is done.
Especially as it's just the setup for the actual addition I'll put on top of this change after it's accepted.

snoyberg (approved):

PR #878, opened 2022-04-18
Add Select Middleware

blackheaven:

After submitting your PR:

Sorry for the two bumps in two days.

See you tomorrow.

PS: Just kidding

### DIFF
--- a/wai-extra/Network/Wai/Middleware/Select.hs
+++ b/wai-extra/Network/Wai/Middleware/Select.hs
@@ -77,8 +77,8 @@ passthroughMiddleware = id
 
 -- | Use the 'Middleware' when the predicate holds.
 selectMiddlewareOn :: (Request -> Bool) -> Middleware -> MiddlewareSelection
-selectMiddlewareOn doesApplypath mw = MiddlewareSelection $ \request ->
-  if doesApplypath request
+selectMiddlewareOn doesApply mw = MiddlewareSelection $ \request ->
+  if doesApply request
     then Just mw
     else Nothing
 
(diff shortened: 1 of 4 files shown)


wai-extra/Network/Wai/Middleware/Select.hs
```diff
@@ -0,0 +1,91 @@
...
+newtype MiddlewareSelection = MiddlewareSelection
+  { applySelectedMiddleware :: Request -> Maybe Middleware
+  }
+
+instance Semigroup MiddlewareSelection where
+  MiddlewareSelection f <> MiddlewareSelection g =
+    MiddlewareSelection $ \req -> f req <|> g req
+
+instance Monoid MiddlewareSelection where
+  mempty = MiddlewareSelection $ const $ Just passthroughMiddleware
```
snoyberg, line 61: This doesn't look law-abiding to me. It should probably be `const Nothing`. I don't think it's just an issue of laws, I think the current implementation could lead to surprising behavior.

blackheaven: Good catch, thanks

wai-extra/Network/Wai/Middleware/Select.hs
```diff
@@ -0,0 +1,91 @@
...
+-- * Helpers
+--------------------------------------------------
+
+passthroughMiddleware :: Middleware
+passthroughMiddleware = id
+
+-- | Use the 'Middleware' when the predicate holds.
+selectMiddlewareOn :: (Request -> Bool) -> Middleware -> MiddlewareSelection
+selectMiddlewareOn doesApplypath mw = MiddlewareSelection $ \request ->
+  if doesApplypath request
```
snoyberg, line 81: I don't think `path` is needed in `doesApplypath`, this predicate can work for any kind of test.

snoyberg (changes requested):

wai-extra/Network/Wai/Middleware/Select.hs
blackheaven: Actually, it was a remaining of a refactoring that obviously went wrong :/

PR #880, opened 2022-04-19
"Overhaul" of `gzip` middleware

Vlix:

After submitting your PR:

---

# Gzip Makeover

Few things done to the `gzip` middleware:
* accepting `q` values in the `Accept-Encoding` header
* adding a few more checks (e.g. don't compress partial content)
* don't just catch `SomeException`
* don't toss away any previously set `Vary` header
* made the size threshold customizable
* added tests to be sure the gzipping actually does what it's supposed to
* added documentation to explain to a new user how this middleware works


Vlix:
Ah, `dropWhileEnd` is a new function in `bytestring`... will change it to `spanEnd` at some point. If I have time, tomorrow.

Vlix:
Merged with `master` and bumped version to 3.1.11 (and adjusted ChangeLog to reflect the correct state of events)

Vlix:
Only failure is:
```
warp                 > Failures:
warp                 > 
warp                 >   test/RunSpec.hs:257:17: 
warp                 >   1) Run, special input, multiline headers
warp                 >        expected: [("foo","bar baz\tbin")]
warp                 >         but got: []
warp                 > 
warp                 >   To rerun use: --match "/Run/special input/multiline headers/"
warp                 > 
warp                 > Randomized with seed 1543620305
```
NOOOOO idea why

wai-extra/Network/Wai/Header.hs
```diff
@@ -14,5 +21,64 @@ contentLength hdrs = lookup H.hContentLength hdrs >>= readInt
...
+    go = checkQ . S.break (== _semicolon)
+    checkQ :: (S.ByteString, S.ByteString) -> (S.ByteString, Maybe Int)
+    checkQ (val, "") = (val, Just 1000)
+    checkQ (val, bs) =
+        -- RFC 7231 says optional whitespace can be around the semicolon.
+        -- So drop any before it       ,           . and any behind it       $ and drop the semicolon
+        (dropWhileEnd (== _space) val, parseQval . S.dropWhile (== _space) $ S.drop 1 bs)
+      where
+        parseQval qVal = do
+            q <- S8.stripPrefix "q=" qVal
```
snoyberg, line 70: `S8` doesn't seem necessary here, `stripPrefix` still works on `ByteString`s.

wai-extra/Network/Wai/Header.hs
```diff
@@ -14,5 +21,64 @@ contentLength hdrs = lookup H.hContentLength hdrs >>= readInt
...
+            (i, rest) <- S8.uncons q
```
snoyberg, line 71: It would be preferable to not use `S8.uncons` since this involves a conversion cost to a `Char`. Instead, using the codes for `'0`' and `'1'`, or https://www.stackage.org/package/word8, is ideal.

wai-extra/Network/Wai/Header.hs
```diff
@@ -14,5 +21,64 @@ contentLength hdrs = lookup H.hContentLength hdrs >>= readInt
...
+            guard $
+                i `elem` ['0', '1']
+                    && S.length rest <= 4
+            case S.uncons rest of
+                -- q = "0" or "1"
+                Nothing -> (1000 *) <$> readMaybe (S8.unpack q)
```
snoyberg, line 77: `readMaybe` has a similar cost issue with data type conversion

wai-extra/Network/Wai/Middleware/AcceptOverride.hs
```diff
@@ -1,10 +1,28 @@
...
     ) where
 
 import Control.Monad (join)
-import Data.ByteString (ByteString)
-import Network.Wai
+import Network.Wai (Middleware, Request (..))
+
+import Network.Wai.Header (replaceHeader)
+
+-- $howto
```
snoyberg, line 11: Again, it's preferable to include unrelated changes in separate PRs

wai-extra/Network/Wai/Middleware/Gzip.hs
```diff
@@ -126,32 +231,49 @@ gzip set app req sendResponse'
...
-    vary = (hVary, "Accept-Encoding")
+    sendResponse = sendResponse' . mapResponseHeaders mAddVary
+    acceptEncoding = "Accept-Encoding"
+    -- Instead of just adding a header willy-nilly, we check if
+    -- "Vary" is already present, and add to it if not already included.
+    mAddVary [] = [(hVary, acceptEncoding)]
+    mAddVary (h@(nm, val) : hs)
+        | nm == hVary =
+            let vals = splitCommas val
+                hasAccEnc = isJust $ L.find (== acceptEncoding) vals
```
snoyberg, line 242: Is this the same as `any`?

wai-extra/Network/Wai/Middleware/Gzip.hs
```diff
@@ -189,11 +311,23 @@ compressFile s hs file cache sendResponse = do
                             Z.feedDeflate deflate bs >>= goPopper
                             loop
                     goPopper $ Z.finishDeflate deflate
-            either onErr (const onSucc) (x :: Either SomeException ()) -- FIXME bad! don't catch all exceptions like that!
```
snoyberg, line 192: Is the FIXME not relevant?

Vlix: The change fixes the FIXME, no?
It only catches what could be thrown in the `try` block and rethrows everything else.

Or do you mean the comment should remain?
What if I added a comment like:
`We generally don't want to catch all exceptions, but we catch `SomeException` here to check on two types of exceptions and then rethrow any other exception. Effectively only catching the ones we want to catch.`

Vlix:
I don't want to seem pushy, but I don't know if the new commits and reactions to the review comments are muted or not, so just wanted to make post a comment to keep this PR alive. Please don't feel rushed.

For convenience: [here's the diff from after @snoyberg 's review](https://github.com/yesodweb/wai/pull/880/files/f7392c043836d529199be7813e908988a9849fcb..a981f1188217aeb333b3999c62de47c905875f61)

wai-extra/Network/Wai/Middleware/Gzip.hs
snoyberg: No it's fine as is. I didn't grok on first read.

snoyberg (approved):
Thanks!

Issue #243, opened 2014-04-03
Gzip middleware cache should index by etag as well


meteficha:
Hi! =)

If I'm reading the code correctly, right now the gzip middleware may serve stale data from the gzip cache since it only takes the filename into account.  Its etag, if available on the response header, should be used as well, right?

kanishka-azimi:
Current progress, next steps?

Vlix:
I feel like taking a jab at the GZip middleware. Any objections? (since this issue is a bit old)

snoyberg:
No objection from me.

Vlix:
First draft made in #875

Vlix:
I found a bunch of subtle edge cases and I'm making this Middleware a bit more robust before I add the ETag functionality.
I think I'm done with the extra checks and tweaks in handling headers, but I still want to add some tests for what is testable with this Middleware. Hope make a PR before end of next week.

Vlix:
Second step in #880 , new and improved `gzip` middleware. After this, I'll actually add the ETag functionality 😅

Vlix:
Implemented with #885

PR #885, opened 2022-05-08
Use ETag gzip middleware (finally)

Vlix:

After submitting your PR:

---

And finally, the PR we've all been waiting for!  (Ok, maybe just me, but hey, happy to have gotten this far)

This change adds using the ETag in the caching of files, so that when used, different versions of the same compressed file can be cached.

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -2,7 +2,7 @@
 
 ## 3.1.12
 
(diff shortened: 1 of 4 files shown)


Vlix:
If this is accepted and merged, this will implement #243

snoyberg (approved):
Awesome!

PR #889, opened 2022-05-24
GHC 9.4 Support

parsonsmatt:

After submitting your PR:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Settings.hs
+++ b/warp/Network/Wai/Handler/Warp/Settings.hs
@@ -249,7 +249,16 @@ exceptionResponseForDebug e =
 -- @since 3.3.17
 defaultFork :: ((forall a. IO a -> IO a) -> IO ()) -> IO ()
 defaultFork io =
+#if __GLASGOW_HASKELL__ >= 904
+  IO $ \s0 ->
+    case io unsafeUnmask of
+      IO io' ->
+        case (fork# io' s0) of
+          (# s1, _tid #) ->
+            (# s1, () #)
+#else
   IO $ \s0 ->
     case (fork# (io unsafeUnmask) s0) of
       (# s1, _tid #) ->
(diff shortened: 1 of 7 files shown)


warp/Network/Wai/Handler/Warp/Settings.hs
```diff
@@ -249,7 +249,16 @@ exceptionResponseForDebug e =
 -- @since 3.3.17
 defaultFork :: ((forall a. IO a -> IO a) -> IO ()) -> IO ()
 defaultFork io =
+#if __GLASGOW_HASKELL__ >= 904
```
kazu-yamamoto, line 252: Would you provide us a material to know the background of this change?

parsonsmatt: Would love to but could not find any info on the changelog for `base` or `ghc-prim` about this. When GHC 9.4 is closer to release, I hope that this info will be more obvious.

kazu-yamamoto: @snoyberg I trust Matt and have no objection for merging. What about you?

kazu-yamamoto (approved):
LGTM

warp/Network/Wai/Handler/Warp/Settings.hs
snoyberg: No objection from me either, go for it! As always thanks Matt 💪

parsonsmatt: Of course!

True 9.4 support relies on updated `foundation` and `basement`, but I think we can release this as-is and it'll work when those libraries are up-to-date.

I can leave the `cabal.project` in the repo and merge as-is, or we can leave pending until that's been taken care of.

kazu-yamamoto:
Merged and a new version has been released.

PR #890, opened 2022-07-22
Using InvalidRequest instead of HTTP2Error

kazu-yamamoto:
I'm preparing http2 lib v4 since RFC9113 was published. http2 v4 discovered a flaw of #741.

413 and 431 is in the semantics level while HTTP2Error is in the transport level. It is mis-use that a response status code is contained in ConnectionError.

My review of #741 was not good enough. I'm very sorry.

I would like to add RequestHeaderFieldsTooLarge and PayloadTooLarge to InvalidRequest and use them instead of ConnectionError. This reduces the dependency from wai-extra to http2.

This PR also contains preparation for http2 v4.

Cc: @michalrus and @vlatkoB

### DIFF
--- a/wai-extra/Network/Wai/Parse.hs
+++ b/wai-extra/Network/Wai/Parse.hs
@@ -70,14 +70,10 @@ import Data.Word (Word8)
 import Network.HTTP.Types (hContentType)
 import qualified Network.HTTP.Types as H
 import Network.Wai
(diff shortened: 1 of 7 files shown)


snoyberg (approved):
Nice!

kazu-yamamoto:
Thanks. I will merge this in the next week.

kazu-yamamoto:
Rebased and merged.

PR #891, opened 2022-07-23
added hie.yaml and adjusted stack.yaml

Vlix:
Wanted to get everything to work with HLS/VSCode.

Added hie.yaml, bumped the resolver version to 18.28 and removed 'extra-deps' from the stack.yaml

Also added `wai-frontend-monadcgi` and `wai-http2-extra` to the stack.yaml, because without those there, HLS will not work on those packages. **If these should not be included for specific reasons, I can comment them out**


snoyberg (approved):
LGTM, thank you!

PR #892, opened 2022-07-24
Refactors that should not impact anything

Vlix:
Just went through all modules to see if there were any redundant easy fixes.

Only one that might not work for all GHCs is a module where I removed `ScopedTypeVariables`, but I'm pretty sure the type inference should pick that up without the explicit type signature.

---


Vlix:
Seems to all still work.

Also "fixed" (hopefully) 3 flaky tests that I've seen fail randomly previously. (upping some `threadDelay`s to 5ms, instead of just 1ms)

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:

PR #895, opened 2022-07-27
Fix #894 by creating a bigger buffer when the current one is too small to fit the Builder

### LINKED ISSUE #894: toBufIOWith: BufferFull: minSize
* Your OS name and version: **NixOS-22.05, Linux kernel 5.15.53**
* The versions of tools you're using: **GHC-9.2.3**
* The versions of dependencies you're using: [cabal.project.freeze](https://gist.github.com/basvandijk/8b6d4e000d66655fc044888fcce5634a#file-cabal-project-freeze)

I have a warp-3.3.20 web-server that is serving a prometheus `/metrics` endpoint using the [prometheus-2.2.3](https://hackage.haskell.org/package/prometheus-2.2.3) package. Yesterday it started to occasionally throw the error:

```
Jul 25 19:20:33 XXX XXX[3866464]:
  Request {
    requestMethod = "GET",
    httpVersion = HTTP/1.1,
    rawPathInfo = "/metrics",
    rawQueryString = "",
    requestHeaders = [
      ("Connection","close"),
      ("Host","XXX"),
      ("X-Real-IP","XXX"),
      ("X-Forwarded-For","XXX"),
      ("X-Forwarded-Proto","https"),
      ("X-Forwarded-Host","XXX"),
      ("X-Forwarded-Server","XXX"),
      ("User-Agent","Prometheus/2.23.0"),
      ("Accept","application/openmetrics-text; version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1"),
      ("Accept-Encoding","gzip"),
      ("X-Prometheus-Scrape-Timeout-Seconds","10.000000")
    ],
    isSecure = False,
    remoteHost = 127.0.0.1:41744,
    pathInfo = ["metrics"],
    queryString = [],
    requestBody = <IO ByteString>,
    vault = <Vault>,
    requestBodyLength = KnownLength 0,
    requestHeaderHost = Just "XXX",
    requestHeaderRange = Nothing
  }
    threw exception: toBufIOWith: BufferFull: minSize
      error, called at ./Network/Wai/Handler/Warp/IO.hs:23:36 in warp-3.3.20-76HAz05JSpF9OEUtOoqO7e:Network.Wai.Handler.Warp.IO
```

(As you might have guessed from the headers, there's a nginx reverse proxy server in front of warp handling TLS termination).

That error is thrown at [`warp/Network/Wai/Handler/Warp/IO.hs#L23`](https://github.com/yesodweb/wai/blob/warp-3.3.20/warp/Network/Wai/Handler/Warp/IO.hs#L23).

The endpoint is implemented by calling [`prometheusApp`](https://hackage.haskell.org/package/prometheus-2.2.3/docs/System-Metrics-Prometheus-Http-Scrape.html#v:prometheusApp) which internally uses [`responseBuilder`](https://hackage.haskell.org/package/wai-3.2.3/docs/Network-Wai.html#v:responseBuilder) to construct a response from a Builder.

I counted 164 failures out of 5017 successes so far. The failures usually happen in bursts sometimes lasting over 10 minutes.

Note that this web-server is also serving lots of other endpoints. Serving `/metrics` only makes up about 3% of overall traffic but it's the only endpoint that is failing.

The most significant change is that I upgraded GHC from 9.0.2 to 9.2.3 yesterday at 11:17 UTC. The first error appeared 8 hours later at 19:20 that day.

Any idea why this is happening?

basvandijk:
This fixes the following error in `toBufIOWith` (#894):

```haskell
    | size < minSize -> error "toBufIOWith: BufferFull: minSize"
```

by, instead of erroring, dynamically creating a bigger buffer (of size `minSize`) that will fit the next batch of bytes from the Builder.

Note I didn't bump the version since this only changes internal APIs.


warp/Network/Wai/Handler/Warp/Types.hs
```diff
@@ -94,6 +94,17 @@ type BufferPool = IORef ByteString
 -- | Type for buffer
 type Buffer = Ptr Word8
 
+-- | A write buffer of a specified size
+-- containing bytes and a way to free the buffer.
+data WriteBuffer = WriteBuffer {
+      bufBytes :: Buffer
```
kazu-yamamoto, line 100: In my opinion, `bufBytes` is confusing. What about `bufBuffer`?

kazu-yamamoto:
The idea seems good. My concern is that there are no hard limit. To prevent allocating a really huge buffer, I would like to set a hard limit somehow.

warp/Network/Wai/Handler/Warp/Types.hs
basvandijk: I agree. Renamed.

basvandijk:
> To prevent allocating a really huge buffer, I would like to set a hard limit somehow.

That makes sense. What about ^ ?

I just picked some arbitrary default maximum of 1 MiB.

kazu-yamamoto (approved):
Now LGTM.

kazu-yamamoto:
Merged. Thanks!

kazu-yamamoto:

PR #899, opened 2022-08-19
Extracting receiving functions as the recv library

kazu-yamamoto:
This PR extracts receiving functions which manages buffer pools.
I hope that the "recv" library will be used from other projects.
The document is, I think, well-written.
So, I also hope that users can understand the idea of the efficient receiving functions.

I think CI will not pass since "recv" is not out there. But I tested this change on macOS and Windows.

After submitting your PR:

- [] Update the Changelog.md file with a link to your PR


snoyberg (approved):
I haven't done a thorough review, but have no problem with the general concept of extracting to a new helper library.

kazu-yamamoto:
Merged. Thank you for your review!

PR #900, opened 2022-08-19
using the same logic for "spec" as "library".

### LINKED ISSUE #898: warp tests don't build on non-linux/freebsd/darwin
Hello,

As can be read on

https://buildd.debian.org/status/fetch.php?pkg=haskell-warp&arch=hurd-i386&ver=3.3.21-2&stamp=1660597971&raw=0

```
Building test suite 'spec' for warp-3.3.21..
[ 1 of 48] Compiling HTTP             ( test/HTTP.hs, dist-ghc/build/spec/spec-tmp/HTTP.o )
[ 2 of 48] Compiling Network.Wai.Handler.Warp.Date ( Network/Wai/Handler/Warp/Date.hs, dist-ghc/build/spec/spec-tmp/Network/Wai/Handler/Warp/Date.o )

Network/Wai/Handler/Warp/Date.hs:17:1: error:
    Could not load module ‘System.Posix’
    It is a member of the hidden package ‘unix-2.7.2.2’.
    Perhaps you need to add ‘unix’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
17 | import System.Posix (epochTime)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Warp tests don't build on GNU/Hurd, because there's a missing dependency on unix. Looking at `warp.cabal` we can read

```
Test-Suite spec
[...]
  if (os(linux) || os(freebsd) || os(darwin)) && flag(allow-sendfilefd)
    Cpp-Options:   -DSENDFILEFD
    Build-Depends: unix
```

And thus on non-linux/freebsd/darwin, the unix dependency is not pulled, even though we have in  `Network/Wai/Handler/Warp/Date.hs` :

```
import System.Posix (epochTime)
```

I'd say that that part of `warp.cabal` should be modified to always build-depend on `unix` on non-windows?

kazu-yamamoto:
This should fix #898.

@sthibaul Would you review this patch?

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,6 +1,7 @@
 packages:
     auto-update
     mime-types
+    recv
     wai
     wai-extra
     warp
--- a/warp/warp.cabal
+++ b/warp/warp.cabal
@@ -226,11 +226,13 @@ Test-Suite spec
                    , transformers
 
   if (os(linux) || os(freebsd) || os(darwin)) && flag(allow-sendfilefd)
-    Cpp-Options:   -DSENDFILEFD
-    Build-Depends: unix
+      Cpp-Options:   -DSENDFILEFD
   if os(windows)
-    Cpp-Options:   -DWINDOWS
-    Build-Depends: time
+      Cpp-Options:   -DWINDOWS
+      Build-Depends: time
+  else
+      Build-Depends: unix
+      Other-modules: Network.Wai.Handler.Warp.MultiMap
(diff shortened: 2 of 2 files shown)


sthibaul:
That looks good to me indeed

kazu-yamamoto:
Rebased and merged. Thank you for your review.

PR #902, opened 2022-08-21
Cleanup and Data.List prep

Vlix:

After submitting your PR:

---

Just some cleanup to get less warnings when compiling and removal of unused packages.

Only one I'm not completely sure about is the `network` package that had a version restriction on it. If that was on purpose (so that the other packages using it don't use a version that's too low?) then I might revert that one if anyone can confirm that it was there for a reason, but if CI passes everywhere I guess it should be fine?


Vlix:
Hmm, I don't really get what happens in the MacOS tests 🤔

Vlix:
@kazu-yamamoto @snoyberg 
Do you maybe know what's happening here? It seems to be some kind of `doctest` failure? But I'm kind of stumped to figure out what it could be. (and why it only, but consistently, fails on MacOS)

Vlix:
I've added "zero-width spaces" to the problematic `*/*` in the `RequestLogger.hs` module, since it only seems to be a haddock comment that's on a function that isn't even exported by the module. And there's no doctest going over it or anything, so I feel this works just fine.

If anyone has a big problem with this, I can revert it and remove the `CPP` pragma again.

**With this, I think this PR is ready for review now.**

 (I also noticed `warp`'s version has been bumped because of the new `recv` package, but the ChangeLog wasn't updated and it hasn't been released, so I took the liberty of adding that as well)

snoyberg (approved):
Thanks!

PR #903, opened 2022-08-23
Recovering CI

kazu-yamamoto:
Test purpose only. Do not merge now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,6 @@ jobs:
         - "--resolver lts-18"
         - "--resolver lts-17 --stack-yaml stack-lts-17.yaml"
         - "--resolver lts-16 --stack-yaml stack-lts-17.yaml"
-        - "--resolver lts-14 --stack-yaml stack-lts-14.yaml"
 
     steps:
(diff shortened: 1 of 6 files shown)


kazu-yamamoto:
@snoyberg How should we modify stack.yaml files to pass CI when we extracted a new library?
I tried several things but no luck.

snoyberg:
I got this to build locally by removing the extra-deps version in the stack.yaml:

```
-extra-deps:
-  - ./recv
```

The basic idea is that each package should appear exactly once, and for local packages in the monorepo you should put them under `packages` so that their test suites run. I can push another commit to your branch if you want, just let me know.

kazu-yamamoto:
@snoyberg Please push your patch.

snoyberg:
Check out #904. It doesn't pass, but I think those are legitimate failures. (Though we may want to drop lts-14.)

kazu-yamamoto:
@snoyberg All tests except lts-14 passed. Since you want to drop lts-14, I dropped it. Are you OK to merge this PR to master?

kazu-yamamoto:
@snoyberg @Vlix Thank you for your help!

PR #904, opened 2022-08-24
CI fix for recv package

snoyberg:


Vlix:
It seems `recv` should be added to the build-dependecies of `warp.cabal`?

kazu-yamamoto:
`warp.cabal` has `recv`.

snoyberg:
It looks like the dep is missing for the benchmarks.

Vlix:
> It looks like the dep is missing for the benchmarks.

Ah, indeed.

For reference:
```
warp                         > Building benchmark 'parser' for warp-3.3.23..
[...]
warp                         > [7 of 8] Compiling Network.Wai.Handler.Warp.Types
warp                         > 
warp                         > /home/runner/work/wai/wai/warp/Network/Wai/Handler/Warp/Types.hs:14:1: error:
warp                         >     Could not load module ‘Network.Socket.BufferPool’
warp                         >     It is a member of the hidden package ‘recv-0.0.0’.
```

PR #906, opened 2022-09-05
Mime types updates

tarleb:

After submitting your PR:


Vlix:
Do you have sources one can use to review your changes? 🤔

ysangkok:
RAR doesn't seem to have a source in the commit, but there is a registration at IANA: https://www.iana.org/assignments/media-types/application/vnd.rar

tarleb:
Thanks, I'll amend the commit for reference.

tarleb:
The last commit (TeX-related types) is probably the most debatable. I followed what appears to be common practice for handling these files. For `.bib` files I followed https://blog.datacite.org/changes-to-doi-content-negotiation/. The preprint repository arXiv uses `application/x-eprint` for `.tex` files, but that seems to be the exception.

tarleb:
Point that I forgot to mention: the `application/x-tex` type is already used for `tex` files. The respective commit just adds more file extensions.

Some background on why I'm doing this PR: pandoc maintains a similar list in the `Text.Pandoc.MIME` module. I have [updated](https://github.com/jgm/pandoc/commit/b9a7de32f6bd729266861ca0942cd2471a9ad2c4) the module to use mime-types as its base. My goal is that that there is only one such list in the Haskell ecosystem. Hence the effort to include important types in this list here; the list in pandoc could then be trimmed down further. I assume that's a win-win.

mime-types/Network/Mime.hs
```diff
@@ -494,9 +499,9 @@ defaultMimeMap = Map.fromAscList [
     , ("m2a", "audio/mpeg")
     , ("m2v", "video/mpeg")
     , ("m3a", "audio/mpeg")
-    , ("m3u", "audio/x-mpegurl")
+    , ("m3u", "audio/mpegurl")
     , ("m3u8", "application/vnd.apple.mpegurl")
```
Vlix, line 503: Maybe turn this into `audio/mpegurl` as well? (since `.m3u` and `.m3u8` are the same thing)

Vlix (commented):
I've gone through all the provided sources and reasonings and all changes seem legitimate to me.

I've added one comment in a spot where it might be nice to keep it consistent (`m3u` and `m3u8`), but overall this seems like an ok change/addition.

mime-types/Network/Mime.hs
tarleb: Oh, good point. I noticed that only `application/vnd.apple.mpegurl` has been registered by IANA, so I used that instead of `audio/mpegurl`.

tarleb:
Thanks for the review!
I hope it's ok that I picked one additional low-hanging fruit (`application/x-gzip` to `application/gzip`, [RFC 6713](https://www.rfc-editor.org/rfc/rfc6713.html)).

Vlix:
I'd like one other reviewer to check these changes (at the least superficially) for any issues that could potentially arrise because of the changes of (sub)types.

Come to think of it. If these were only additions, I think a patch version bump would be fine, but seeing as some existing extensions' (sub)types are changed, this might warrant a minor version bump at least. 🤔

snoyberg:
I think this is great, thanks for the contribution @tarleb and the thorough review @Vlix. I'd be happy to merge and release this as is, unless @Vlix has more comments or wants to do the merge/release instead.

Vlix:
@snoyberg If you could merge and release, that'd be great 👍

snoyberg:
Done!

Issue #907, opened 2022-09-05
warp-tls: offer the possibility to provide our own data to the library via some hooks or `Connection` object


jecaro:
I'm working on a low level network application. And basically I'd need to pass some buffered data straight to warp-tls instead of having it fetching from client sockets.

Ideally, I'd need to have some kind of interface or hooks in the like of `Connection` objects so I can pass my data to warp-tls. I understand this is not possible right now. FWIU of the code, warp-tls performs some low level operations on the client socket before creating such `Connection` object and running warp on it.

I'm asking for advice & ideas here. Would that be possible to provide such an interface for warp-tls ? How difficult it would be ? Any other available approaches that I missed to address this issue ?

kazu-yamamoto:
The `tls` package has `Backend`: https://hackage.haskell.org/package/tls-1.6.0/docs/Network-TLS.html#g:2

Would you give a look at `Backend` first? It's quite easy to extend `warp-tls` to specify `Backend` to `tls`, I guess.

jecaro:
Thx for pointing this. 

That looks like a better idea. And for handling plain HTTP, one can create the required `Connection` from the TLS `Backend`, just like it's done for HTTP over TLS. Indeed, I think that could work. 

Would you be interested in a PR on that feature ?

kazu-yamamoto:
Yes, of course!

jecaro:
Closing as #912 solves this issue

PR #909, opened 2022-09-12
make it possible to pass a backend to warp-tls

jecaro:
Here is a WIP of the feature discussed in #907. It works correctly in my tests. 

Basically, this is how it work. `runTLSBackend` takes an `accept` like function that returns a `Backend`. Then that `Backend` :
- is wrapped into another `Backend` for TLS
- is wrapped into a `Connection` for HTTP

then given to `warp`.

The `runTLSSocket` creates a `Backend` from the raw socket then it works as described before.

Here are the things worth to note:
- `safeRecv` cannot be used in this approach as it operates on the socket. I'm not sure how important it is to use it.
- the `HasBackend` instance for the socket provided by the `tls` package cannot be used here as the `backendRecv` function it creates blocks if there is no enough bytes in the queue.
- the `backendToConnection` function is ~~copied from~~ heavily inspired by `Network.Wai.Handler.Warp.Run.socketConnection` but without the buffer pool optimization
- the performances of HTTPS are roughly the same in my tests but are degraded for HTTP. This is prolly due of the removal of the buffer optimization provided by the `recv` package.

Overall I'm not very happy with this approach mostly because it degrades performances. Here are some ideas of different approaches:
- having two paths in the code: one for sockets with the optimizations and one for `Backend`. But that means a lot a duplicated code that does fundamentally the same stuff. 
- having an `accept` like function that returns a `Connection` object. That `Connection` object could include the buffer pool optimization upfront.
- in my app, at the end, I do have a socket behind. So being able to call `warp-tls` with an `accept` like function that returns the socket should also solve my issue.

On the all different approaches, I think my preference would go to the last one: plain simple and offering the same performances. 

Asking for feedback: what do you guys think about this stuff ? What would be the way to move forward ? See #907 for context.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -16,6 +16,7 @@ module Network.Wai.Handler.WarpTLS (
     -- * Runner
       runTLS
     , runTLSSocket
+    , runTLSBackend
     -- * Settings
     , TLSSettings
     , defaultTlsSettings
@@ -44,6 +45,8 @@ module Network.Wai.Handler.WarpTLS (
     , OnInsecure (..)
     -- * Exception
     , WarpTLSException (..)
+    -- * Conversion
+    , sockToBackend
     -- * DH parameters (re-exports)
     --
     -- | This custom DH parameters are not necessary anymore because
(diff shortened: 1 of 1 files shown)


jecaro:
Gently ping @kazu-yamamoto 

Is that what you had in mind ? Any comments on my concerns ?

goertzenator:
I just arrived at the need for something like this today and am delighted to see a PR in-progress.  I also am dealing with sockets underneath, so the approach I had in mind was adding a `settingsAccept` member to `Settings` in the same manner as `settingsFork`.

The [existing call to accept](https://github.com/yesodweb/wai/blob/master/warp-tls/Network/Wai/Handler/WarpTLS.hs#L279-L287) would be swapped out with the one provided in `Settings`, and there would be a `defaultAccept` function installed by default that invokes [current accept code](https://github.com/yesodweb/wai/blob/master/warp-tls/Network/Wai/Handler/WarpTLS.hs#L281-L285).

Thoughts?

My use case is reverse socket connections for a Warp server in an IOT device.  The device establishes a connection to some cloud server, and then the cloud server begins TLS/http negotiation as client and the device as server.  IOT devices are typically not reachable due to NAT and firewalls.

kazu-yamamoto:
I'm trying to understand this conversation.

@goertzenator Would you implement `settingsAccept` to show how it can solves the issues of this PR?

kazu-yamamoto:
@jecaro General comments:

- You can modify Warp's `Settings` if it eases your approach.
- Performance degradation is not acceptable.

jecaro:
Thx for the feedback. Indeed degradation of performances is unacceptable. 

In the case of this PR (implemented from [this idea](/yesodweb/wai/issues/907#issuecomment-1237547723)), I dont think modifying `Settings` would be a good idea for these two reasons:
- we would need to add `tls` as a dependency to `warp` which is not currently the case
- that would introduce different branches in the code making it more complicated than it needs to be

I think @goertzenator  approach would lead to something less convoluted, easier to understand thus easier to maintain which is basically very close to what I had in mind in my comment:

> in my app, at the end, I do have a socket behind. So being able to call `warp-tls` with an `accept` like function that returns the socket should also solve my issue.

jecaro:
Closing in favor of #912

PR #921, opened 2023-02-01
No recvBuf

kazu-yamamoto:
The receiving function which receives exactly N bytes is necessary for the HTTP/2 module in Warp and for the backend of TLS. This function was created by:

```haskell
-- The first ByteString is the leftover in the previous procedure.
makeReceiveN :: ByteString -> Recv -> RecvBuf -> IO RecvN

type Recv = IO ByteString
type RecvBuf = Buffer -> BufSize -> IO Bool
type RecvN = Int -> IO ByteString
```

The first `ByteString` is managed via `IORef`.
`RecvBuf` was necessary for the case where `IORef` does not contain required size.
In this case, a buffer of the required size is created and `RecvBuf` is repeatedly used to fill the buffer.

I realized that `RecvBuf` is not necessary to implement `makeReceiveN`.
We can store received `ByteString` to a builder of `[ByteString]` with the total size counted.
When the total size reaches the required size, we can create a buffer and copy data from the list.

This idea makes our code much.

But the `recv` package will break the backward compatibility.
One concern is that Warp does not set the upper boundary for `recv`.
So, I have already set the upper boundary of Warp v3.3.23 on Hackage.

I have tested this code in the real world for a week.

@snoyberg I hope you like this.


kazu-yamamoto:
@snoyberg Could you take time to review this PR?

snoyberg (approved):
Sorry for the delayed review, LGTM!

kazu-yamamoto:
Merged. Thank you for reviewing!

PR #927, opened 2023-03-24
using unix instead of unix-compat

### LINKED ISSUE #919: Consider dropping `unix-compat` dependency
Hello! I've been a bit blocked on using `warp` with the new `unix-2.8.0.0` that was released over 4 months ago. I believe `unix-compat` is the only dependency of `warp` that is incompatible with `unix-2.8.0.0`, but the maintainer seems absent.

Perhaps it would make sense to shed the `unix-compat` dependency and vendor the cross-platform code we actually use?

Thanks (for reading / considering)!

kazu-yamamoto:
This should fix #919.

@mitchellwrosen Sorry for the delay. Are you satisfied with this PR?

### DIFF
--- a/warp/Network/Wai/Handler/Warp/FileInfoCache.hs
+++ b/warp/Network/Wai/Handler/Warp/FileInfoCache.hs
@@ -6,10 +6,14 @@ module Network.Wai.Handler.Warp.FileInfoCache (
   , getInfo -- test purpose only
   ) where
 
-import qualified UnliftIO (onException, bracket, throwIO)
 import Control.Reaper
 import Network.HTTP.Date
+#if WINDOWS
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
From the CI results, we need to add a condition to `unix` and `unix-compat`.
I will take care of this in the next week.

mitchellwrosen:
Looks good. I will note that since making this issue I've (finally) been granted maintainership of `unix-compat`, so it now has a `unix-2.8`-compatible release on Hackage.

That said, the `unix-compat` package doesn't seem very high quality. I wouldn't recommend using it anywhere, really.

kazu-yamamoto:
Using `unix-compat` only for Windows.
Merged.

PR #937, opened 2023-08-10
Throw exception on "peer closed" event

bgamari:
This is a prototype enabling support for the `evtPeerClosed` event being introduced into GHC's event manager. This allows `warp` to throw a `PeerClosedException` exception to handler threads for connections which have been prematurely closed.

See [ghc!11080](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/11080) and [ghc#23825](https://gitlab.haskell.org/ghc/ghc/-/issues/23825)

## To do

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -8,6 +8,7 @@
 module Network.Wai.Handler.Warp.Run where
 
 import Control.Arrow (first)
+import Control.Concurrent
 import qualified Control.Exception
 import Control.Exception (allowInterrupt)
 import qualified Data.ByteString as S
@@ -21,6 +22,12 @@ import Network.Socket (gracefulClose)
 #endif
 import Network.Socket.BufferPool
 import qualified Network.Socket.ByteString as Sock
+#if MIN_VERSION_base(4,18,0)
+-- For evtPeerClosed
+import Network.Socket (withFdSocket)
+import GHC.Event
+import System.Posix.Types (Fd(Fd))
+#endif
 import Network.Wai
 import System.Environment (lookupEnv)
 import System.IO.Error (ioeGetErrorType)
@@ -59,6 +66,14 @@ socketConnection _ s = do
     bufferPool <- newBufferPool 2048 16384
     writeBuffer <- createWriteBuffer 16384
     writeBufferRef <- newIORef writeBuffer
+#if MIN_VERSION_base(4,18,0)
+    let registerPeerClosedCb = Just $ \cb -> withFdSocket s $ \fd -> do
+            Just mgr <- getSystemEventManager
+            _ <- registerFd mgr (\ _ _ -> cb) (Fd fd) evtPeerClosed OneShot
+            return ()
+#else
+    let registerPeerClosedCb = Nothing
+#endif
     isH2 <- newIORef False -- HTTP/1.x
     return Connection {
         connSendMany = Sock.sendMany s
@@ -80,6 +95,7 @@ socketConnection _ s = do
       , connRecvBuf = \_ _ -> return True -- obsoleted
       , connWriteBuffer = writeBufferRef
       , connHTTP2 = isH2
+      , connRegisterPeerClosedCb = registerPeerClosedCb
       }
   where
     receive' sock pool = UnliftIO.handleIO handler $ receive sock pool
--- a/Repro.hs
+++ b/Repro.hs
@@ -1,0 +1,40 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Main where
+
+import Network.Socket
+import Network.Wai
+import Network.Wai.Handler.Warp
+import Network.HTTP.Types (status200)
+import Network.Wai.Handler.Warp.Internal
+import Data.ByteString.Builder (byteString)
+import Debug.Trace
+import Control.Concurrent
+import qualified Control.Exception as E
+
+main :: IO ()
+main = do
+  let settings =
+        defaultSettings { 
+          settingsOnClose = \_ -> msg "closed!",
+          settingsOnException = \_ e -> msg ("Exception: " ++ show e) >> E.throw e
+        }
+  runSettings settings app
+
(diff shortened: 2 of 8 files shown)


kazu-yamamoto:
In my understanding, the `EPOLLRDHUP` event happens when the peer closes the connection OR shutdown its writing side. My concern is that `PeerClosedException` is also thrown when the client shutdowns its writing side after uploading a file or something.

@bgamari Did you consider the uploading scenario?

bgamari:
Hmm, this is an interesting point, @kazu-yamamoto; I had not considered the uploading case. This indeed complicates matters as we certainly do not want to terminate any handlers until they have at very least had a change to read and process the input sent by the client. Consequently, I suspect the exception should not be delivered until the handler has declared that it is "done" with the connection; the idea here is that then "pure" handlers (e.g. pure queries, not uploads) could signal immediately that they are done with the connection.

I can think of at least two concrete interfaces by which this could be done.

For instance, one could imagine that `Application` could have the form:
```haskell
type Application =
       Request
         -- ^ the request
    -> (Response -> IO ResponseReceieved)
         -- ^ respond to the request
    -> IO ()
         -- ^ signal that the handler is finished reading from the request 
    -> IO ()
```
Alternatively, we could make this a property of the `Request`:
```haskell
requestFinishedReading :: Request -> IO ()
```

Given that changing `Application` would be quite problematic at this juncture, I suspect the latter is the only viable option.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -59,6 +66,14 @@ socketConnection _ s = do
     bufferPool <- newBufferPool 2048 16384
     writeBuffer <- createWriteBuffer 16384
     writeBufferRef <- newIORef writeBuffer
+#if MIN_VERSION_base(4,18,0)
```
kazu-yamamoto, line 69: This condition is not good enough as `error` is called by GHC on macOS.

kazu-yamamoto:
I don't think I understand your idea.
Would you show me how to implement the latter?

kazu-yamamoto:
There are two kinds of applications.

1. Not read the request body: for this applications, `PeerClosedException` is useful.
2. Read the request body: for this applications, `PeerClosedException` is harmful since it is thrown immediately when TCP FIN is received. Pending data (not read yet) is lost. The proper way to detect peer's close/shutdown in this case is just use `read()` until EOF.

kazu-yamamoto:
Since HTTP/1.1 uses persist connections, clients rarely uses `shutdown()`.
But it is OK for clients to use `shutdown()` after uploading with `Connection: close`.
We need to consider how to save this case.

bgamari:
> I don't think I understand your idea. Would you show me how to implement the latter?

Sure. The idea is that distinguishing between your cases (1) and (2) in general
will require feedback from the user. To get this feedback, we equip each
connection with an "active" flag which can be deasserted by a type (1)
handlers. That is, a pure "query" handler might look like:

```haskell
handleQuery :: Application
handleQuery request respond = do
    -- indicate that the handler can be safely killed if the client shuts down
    -- the connection
    connectionIsInactive request

    -- this may be a long computation
    result <- computeQueryResult
    sendResponse respond result
```

When the client closes their end of the connection, `wai` would first wait
until the connection is in the "inactive" state and would then throw a
`PeerClosedException`.

Naturally, ensuring that this is not racy in the face of connection reuse is
rather tricky. We likely want to re-assert the connection-active flag after
each request has been handled to avoid this sort of behavior:

```
Client                                               Server
---------------------------------                    --------------------------------------
opens connection
                                                     starts connection-handling thread
                                                     connection begins with active=True

submits GET query

                                                     starts handler for query
                                                     handler sets active=False

                                                     sends response

reads response

submits a non-query POST request
                                                     starts handler for request
shuts down connection
                                                     request killed with PeerClosedException
```

Note above that I am not terribly familiar with HTTP/2. My understanding is that it allows
multiple concurrent requests over a single connection; this obviously muddies the waters.

kazu-yamamoto:
OK. It seems to me that this is a right direction.

HTTP/2 clients does not use `shudown()`. So, TCP FIN in HTTP/2 connections is just the end of connections.
I think that no special care for HTTP/2 is necessary in this scheme.

@bgamari Would you put this implementation forward?

kazu-yamamoto:
Looks good. 

- [ ] `connectionIsInactive` is not an appearing API name. What about `readingIsDone` or something?
- [ ] Register a CB only on Linux (commented above)
- [ ] Integrate the test case to hspec and remove `Repro.hs` and `Test.hs`

bgamari:
>  Register a CB only on Linux (commented above)

I am reluctant to make this platform dependent since we may implement `evtPeerClosed` on other platforms eventually (I suspect KQueue supports something similar, although I've struggled to figure out exactly how).

Currently GHC will throw an exception if you attempt to register for `evtPeerClosed` on an unsupported platform, which is why I surround the registration with `tryIO`. I suspect this is the right way forward since the implementation will naturally adapt to the capabilities of the host platform.

>  `connectionIsInactive` is not an appearing API name. What about `readingIsDone` or something?

`readingIsDone` sounds fine to me.

kazu-yamamoto:
> I am reluctant to make this platform dependent since we may implement evtPeerClosed on other platforms eventually (I suspect KQueue supports something similar, although I've struggled to figure out exactly how).

Right. The combination of `EV_ADD/EV_EOF` and `EVFILT_READ` would implement this feature with KQueue.

Vlix:
Could we add the flag to the `Request` to then make a helper/wrapper function for users to implement on the path handlers they want the `PeerClosedConnection` exception to be thrown?
Like `requestClientClosedConnectionFlag :: ConnActiveFlag` and:
```haskell
closeOnClientDisconnect :: Application -> Application
closeOnClientDisconnect app req respond = do
    setupDisconnectCheck $ requestClientClosedConnectionFlag req
    app req respond
```
Where the `setupDisconnectCheck` gets the `ThreadId` of that handler and throws the `PeerClosedConnection` exception when the flag switches?

Issue #893, opened 2022-07-26
`waitForZero` tweak


Vlix:
Might it be a good idea to change the check in `waitForZero` to `<= 0`, instead of `== 0`?

Just in case for _some_ reason it dips below 0 (erroneous double `decrease`) and then never goes back up?
I don't see the downside of this, since negative numbers are breaking the invariant anyway, and this way the gracefulShutdown won't have a chance to hang unnecessarily.

Vlix:
Fixed in #940

PR #940, opened 2023-09-09
warp: changed check in 'waitForZero' to never loop indefinitely and arguably be more readable.

Vlix:
Doesn't need a version bump or anything.

Based on #893


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged.
Thanks!

PR #915, opened 2022-09-29
Install settings shutdown handler for TLS servers

qxjit:
Any shutdown handler passing in the settings to 'runTLS' or 'runTLSSocket' was not being installed in the same way that 'runSettingsSocket' and friends do for non-TLS servers.

This updates 'runTLSSocket' to make it install the shutdown handler from the settings so that warp-tls behaves the same way as warp in this regard.

Before submitting your PR, check that you've:

- [x] Bumped the version number

After submitting your PR:

- [ ] Update the Changelog.md file with a link to your PR
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

PR #942, opened 2023-10-02
warp-tls: fixed typo in ChangeLog

Vlix:
Changelog entry for `3.4.2` was copy-pasted and then not changed, so here's the fix.

@kazu-yamamoto Can you publish `warp-tls-3.4.3` this after the merge? (since I've just merged in the `3.4.3` update before this fix)

### DIFF
--- a/warp-tls/ChangeLog.md
+++ b/warp-tls/ChangeLog.md
@@ -4,7 +4,7 @@
 
 * Install shutdown handlers passed via `Settings` to `run...` functions
 
-## 3.4.1
+## 3.4.2
 
 * Requiring warp v3.3.29.
 


kazu-yamamoto (approved):
Nice catch!

Vlix:
@kazu-yamamoto Do you think you could find the time to publish `warp-tls-3.4.3` to hackage as well? Or do you and Michael maybe have certain schedules for doing so?

kazu-yamamoto:
I have time. Would you want to release `warp` as well?

Vlix:
Warp doesn't need to be published, does it?

kazu-yamamoto:
If #940 is not urgent, the answer is "not need to be published".

Vlix:
I just merged in #915 today, so was wondering what the process was for publishing (since only you and Michael have publishing rights to `warp-tls` on hackage)

kazu-yamamoto:
I was talking about `warp`, not `warp-tls`.
I will release `warp-tls` since #915 was merged.
I asked whether or not we should release `warp` because of #940.

kazu-yamamoto:
`warp-tls` v3.4.3 has been released.

Vlix:
> I was talking about `warp`, not `warp-tls`. I will release `warp-tls` since #915 was merged. I asked whether or not we should release `warp` because of #940.

Apologies, `warp` indeed doesn't need to be released. You mentioning it might have confused me, I think. Thanks for the release of `warp-tls` :+1:

kazu-yamamoto:
My apologies for the confusion.

PR #946, opened 2023-10-12
warp: if we're building the outgoing buffer, why not count the bytes?

Vlix:

After submitting your PR:

---

Would this justify a version bump? :thinking: 

... come to think of it, should the `bytesSent` type be in `Integer`, to be able to count more than 2GB on a 32 bit machine?


kazu-yamamoto (approved):
Nice!

kazu-yamamoto:
Rebased and merged. Thanks!

kazu-yamamoto:
Should we release a new version of `warp`?

Vlix:
I'll make a PR with the last two changes added to the Changelog, and I think I'm going to change the counting type to `Integer`, just so there's no chance of hitting the 2,1GB limit on 32-bit machines.
Then we can release `warp-3.3.30` (but I can do that too, if necessary :smile: )

PR #947, opened 2023-10-13
Last update for warp 3.3.30

Vlix:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/IO.hs
+++ b/warp/Network/Wai/Handler/Warp/IO.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE CPP #-}
-
 module Network.Wai.Handler.Warp.IO where
 
 import Control.Exception (mask_)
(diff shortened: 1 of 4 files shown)


warp/Network/Wai/Handler/Warp/IO.hs
```diff
@@ -10,7 +8,7 @@ import Network.Wai.Handler.Warp.Buffer
 import Network.Wai.Handler.Warp.Imports
 import Network.Wai.Handler.Warp.Types
 
-toBufIOWith :: Int -> IORef WriteBuffer -> (ByteString -> IO ()) -> Builder -> IO Int
+toBufIOWith :: Int -> IORef WriteBuffer -> (ByteString -> IO ()) -> Builder -> IO Integer
```
kazu-yamamoto, line 11: Isn't `Int64` good enough?

Vlix: Would also be fine, but then we'd have to `fromIntegral` it to `Integer` at the next step anyway. So thought, why not just do `Integer` immediately.

Vlix:
Removed the `fromIntegral` that changes the length type to `Integer`

kazu-yamamoto (approved):
Now LGTM!

kazu-yamamoto:
Merged. Thanks!

Vlix:
Tagged commit as `warp-3.3.30` and uploaded to hackage.

PR #950, opened 2023-10-19
wai-extra: redact `Authorization` header when logging a request

austin-artificial:
The `Authorization` request header normally contains sensitive data which we probably don't want to log.

This redacts it in a similar way to what is done for cookies. 

Happy to bump the version number if needed and add a changelog if the idea is approved.

After submitting your PR:


Vlix:
I don't see any problem with this change 🤔

When using a default logger, you'd probably want to err on the safe side, so not leak credentials. If anyone wants to log delicate data, they can implement it themselves.

I'd appreciate a bump in version, update to the ChangeLog and maybe find where describing this redaction (and maybe also the cookie redaction) would make sense and add it there?

PR #955, opened 2023-12-10
Make CI work again

Vlix:
Adjusted all `stack.yaml`s and git actions yaml to have CI work, hopefully

Also added `warp-quic` to the `stack.yaml`s so that everything is being built/tested.


kazu-yamamoto (approved):
Great work!

Issue #956, opened 2023-12-11
Connections HEAD requests are not kept alive if they don't have Content-Length


akshaymankar:
Here is a tiny server for testing:

```haskell
{-# LANGUAGE OverloadedStrings #-}

import Network.Wai.Handler.Warp qualified as Warp
import Network.Wai
import Network.HTTP.Types

main :: IO ()
main = Warp.run 3000 testApp

testApp :: Application
testApp req respond =
  respond $ responseLBS status200 [] ""
```

Any requests made to this server keep the connection alive except for HEAD requests.

<details>
For client side I used telnet. Send a `GET` request like this (note the extra new lines):

```
GET / HTTP/1.1


```

The connection will hang on. Then send a `HEAD` request like this:

```
HEAD / HTTP/1.1


```
</details>

I've tracked it down to these two lines:

https://github.com/yesodweb/wai/blob/2eb5339d4b802903460db1e51063d72345752aad/warp/Network/Wai/Handler/Warp/Response.hs#L143

https://github.com/yesodweb/wai/blob/2eb5339d4b802903460db1e51063d72345752aad/warp/Network/Wai/Handler/Warp/Response.hs#L386

So, `isChunked` is always `False` for `HEAD` requests and it depends completely on the `Content-Length` header in response.
This is not what nginx expects and this causes a race between warp actually closing the connection and nginx using it again causing nginx sometimes return `502 Bad Gateway`. It also logs errors like this:

```
[error] 1198525#1198525: *13 upstream prematurely closed connection while reading response header from upstream,
```

I am not sure why the connection is closed on this. But I also noticed that when warp decides to close a connection it doesn't include the `Connection: close` header to tell the client that it shouldn't try to re-use the connection.

Vlix:
Reading through [RFC9110](https://www.rfc-editor.org/rfc/rfc9110#HEAD), it does say "However, a server MAY omit header fields for which a value is determined only while generating the content. [...] Such a response to GET might contain [Content-Length](https://www.rfc-editor.org/rfc/rfc9110#field.content-length) and [Vary](https://www.rfc-editor.org/rfc/rfc9110#field.vary) fields, for example, that are not generated within a HEAD response." 

So I think removing the `not isHead` would be an acceptable change, right?

kazu-yamamoto:
I don't remember why I gave special treatment to the `HEAD` method.
It was probably because of my misunderstanding.
I agree with @Vlix's observation.

Vlix:
[...]

PR #958, opened 2023-12-12
Send conn close

akshaymankar:
I think this not a very good solution to #956, but at least it avoids nginx tripping over closed connections. I think it would be nice if warp doesn't actually close the connection, but I don't know why it was decided to close the connection, so I might be missing something. 

After submitting your PR:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -110,7 +110,7 @@ sendResponse :: Settings
              -> Response -- ^ HTTP response including status code and response header.
              -> IO Bool -- ^ Returing True if the connection is persistent.
 sendResponse settings conn ii th req reqidxhdr src response = do
-    hs <- addAltSvc settings <$> addServerAndDate hs0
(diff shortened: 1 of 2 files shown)


MangoIV:
Hello! Is there anything we can do here to move this forward? Thanks in advance.

elland:
Is there something we can do about getting this addressed? Thanks.

Vlix:
Seems fine. Though I'd rather not duplicate a header, so a check that it isn't set would be appreciated.

elland:
🍨 bump :)

Vlix:
@kazu-yamamoto except the `http3` issue on nightly, this should be fine, right?

kazu-yamamoto:
I'm fine with this.
But if we could add test cases, it would be great.

blackheaven:
@kazu-yamamoto I have rebased and added a test, let me know what you think

Vlix (approved):
Everything passes, change makes sense, ready for merging 👍

blackheaven:
Thank you :)

PR #962, opened 2023-12-21
delete unused ExceptionInsideResponseBody exception

mitchellwrosen:
This PR deletes the unused `ExceptionInsideResponseBody` exception.


kazu-yamamoto:
`ExceptionInsideResponseBody` was introduced in https://github.com/yesodweb/wai/pull/740
But the code throwing `ExceptionInsideResponseBody` was unintentionally removed.
We should recover this feature.

PR #964, opened 2023-12-25
add and use RequestParseException in Network.Wai.Parse

ocramz:
Before submitting your PR, check that you've:

- [x] Bumped the version number
- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [x] ~~Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

closes #963

PR #965, opened 2023-12-27
Fourmolu

kazu-yamamoto:
- Adding `fourmolu.yaml`
- Reformatting most Haskell files with `fourmolu`

### DIFF
--- a/auto-update/Control/AutoUpdate.hs
+++ b/auto-update/Control/AutoUpdate.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+
 -- | In a multithreaded environment, running actions on a regularly scheduled
 -- background thread can dramatically improve performance.
 -- For example, web servers need to return the current time with each HTTP response.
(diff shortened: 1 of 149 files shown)


Vlix (approved):
Straight-forward addition 👍

The following is one of the reasons I prefer trailing function arrows, though:
```hs
-- leading
type CheckCreds =
    ByteString
    -> ByteString
    -> IO Bool
```
compare with trailing:
```hs
-- trailing
type CheckCreds =
    ByteString ->
    ByteString ->
    IO Bool
```
Since this way: they're alligned, and the one that stands out because it doesn't have an arrow is the result of the function, instead of the first argument. (My reasoning is, the first argument isn't special, the result is, so having that be more visually apparent makes more sense)

And with normal functions the diffs are nicer if the first argument moves:
```diff
  -- leading
  myFunc
-    :: Arg1
-    -> Arg2
+    :: Arg2
+    -> Arg1
     -> Result
```
versus
```diff
  -- trailing
  myFunc ::
-    Arg1 ->
     Arg2 ->
+    Arg1 ->
     Result
```

But if you prefer this, it's a nit-pick, so it's totally fine 👍
LGTM.

kazu-yamamoto:
I have no logical reason why I prefer leading.
Anyway, I will merge this after all possible PRs have been merged.

Vlix:
I'd say just add this in. 👍 
The other open PRs can resolve any conflicts without too much hassle. (only one with more hassle is mine, but I'll reformat it after, no problem)

kazu-yamamoto:
I would like to merge #964 first.

kazu-yamamoto:
Rebased and watching CI.

PR #966, opened 2023-12-29
Use `Data.Word8` when possible

Vlix:
Wanted to get rid of literal numbers when we can use values from `Data.Word8` and have it more obvious what's happening in the code. Also feels more robust to see `_colon` instead of `58` or `0x3A`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,8 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         args:
         - "--resolver nightly --stack-yaml stack-nightly.yaml"
-        - "--resolver lts-21"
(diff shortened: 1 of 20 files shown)


Vlix:
The new Stackage Nightly seems to be missing a lot of stuff.
Then there's `quic` and `websockets` that need to accept `bytestring < 0.13`, I think.

@kazu-yamamoto do you think you'll be able to add `http2`, `http3`, `quic`, `websockets`, etc. to the (nightly) stackage in the foreseeable future?

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
I have released a new version of `quic` which allows `bytestring` 0.12.
`http3` does not have the upper boundary.
I don't maintain `websockets`.

PR #967, opened 2024-01-04
prevent thunks from building when calculating size in 'toBufWith'

Vlix:
Small correction that I missed when making #946

### DIFF
--- a/warp/Network/Wai/Handler/Warp/IO.hs
+++ b/warp/Network/Wai/Handler/Warp/IO.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 module Network.Wai.Handler.Warp.IO where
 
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
This PR has no effect since `Strict` is specified in the cabal file.
Rather, we should delete all `!`.

Vlix:
Might be a good idea to double check there aren't any sections that are assuming laziness then. 🤔 
How long has the `Strict` pragma been enabled by default?

kazu-yamamoto:
This one:

```
commit ce607dd19dcd7ed42c28c173667a8d6abfb8d46c
Author: Kazu Yamamoto <kazu@iij.ad.jp>
Date:   Mon Jun 17 10:41:50 2019 +0900

    using strict language extensions.
```

PR #969, opened 2024-01-04
Multiline syntax bug

Vlix:
warp-tls: when making multiline strings, the first `\` on a new line doesn't escape the next character, it's the start of the next part of the string.

This will result in the following response (note the trailing `r`s)
```
HTTP/1.1 426 Upgrade Requiredr
Upgrade: TLS/1.0, HTTP/1.1r
Connection: Upgrader
Content-Type: text/plain

```

I'm not sure why this hasn't been a problem yet? Probably because it's the response when denying something, so if the response is off, it still indicates that it's a 426? :shrug:


kazu-yamamoto (approved):
Nice catch!

PR #968, opened 2024-01-04
Reworked header lines logic

Vlix:
Before submitting your PR, check that you've:

- [x] Bumped the version number

After submitting your PR:

- [ ] Update the Changelog.md file with a link to your PR
- [x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

---

Had a go at making the header lines parsing more efficient.

I want to get some feedback on this first, and I'll want to check the benchmarks as well to see if this doesn't make anything worse. (Maybe add some benchmarks too)

I'm also wondering if we want to support multiline functionality in headers, still? The [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.4) states it deprecates this functionality. This RFC is from almost 10 years ago. It would make the parsing a lot easier.

PR #973, opened 2024-01-14
Format files with CPP

Vlix:
`fourmolu` skipped a few files because of CPP, so I've gone through them and:
* removed the conflicting CPP
* formatted with `fourmolu`
* re-added the CPP sections


Vlix:
Wanted to do this before I continue with the PR #968

kazu-yamamoto (approved):
LGTM

PR #974, opened 2024-01-22
CI

kazu-yamamoto:
@Vlix I have added http2 v5.1.0 and http3 0.0.8 but CI fails.
Do you know how to recover CI?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,10 +1,10 @@
 name: Tests
 
 on:
-    pull_request:
-    push:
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
My bad.
`PackIntSpec.hs` was not added, exits only on my machine...

Vlix (approved):
`abs (minBound :: Int64)` is still negative, but the chances of hitting that is miniscule. So looks good 👍

kazu-yamamoto:
Thanks!

PR #985, opened 2024-04-26
Make reaper workloads modifiable

FinleyMcIlwaine:
It is convenient to be able to modify reaper workloads outside of the `reaperAction` in scenarios where we don't want to wait for `reaperDelay` for the jobs to be updated. This PR adds an atomic `reaperModify` function to the reaper API which enables this.

Since this changes the visible `Reaper` constructor, I did a major version bump (`auto-update` 0.1.6 -> 0.2.0) and added a changelog entry. I also bumped the `warp` dependency on `auto-update` upper bound.

After submitting your PR:

### DIFF
--- a/auto-update/ChangeLog.md
+++ b/auto-update/ChangeLog.md
@@ -1,5 +1,11 @@
 # ChangeLog for auto-update
 
+## 0.2.0
+
+* Add `reaperModify` to the `Reaper` API, allowing workload modification outside
+  of the main `reaperAction` loop.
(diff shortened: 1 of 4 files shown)


kazu-yamamoto:
@snoyberg I'm OK with this PR.
But I don't remember why the `Reaper` constructor is exported.
Should we stop exporting it and export it from `Reaper.Internal`?

kazu-yamamoto:
Merged anyway.
I will probably make `Reaper.Internal` later.

PR #986, opened 2024-04-26
Eagerly remove `Handle`s from the manager

FinleyMcIlwaine:
This PR is based on #985. That should be merged first, then this can be rebased on `main` and merged.

Previously, if many streams were opened on a connection with a very long timeout, the handles would accumulate. This PR causes the handles to be removed from the manager immediately upon `cancel`, instead of relying on timeouts causing the reaper to remove them.

After submitting your PR:

### DIFF
--- a/auto-update/ChangeLog.md
+++ b/auto-update/ChangeLog.md
@@ -1,5 +1,11 @@
 # ChangeLog for auto-update
 
+## 0.2.0
+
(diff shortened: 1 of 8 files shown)


kazu-yamamoto (approved):
Excellent!

kazu-yamamoto:
Rebased and merged.

Issue #628, opened 2017-06-19
`wai` allows invalid HTTP headers


danclien:
### General
`wai` allows HTTP header names with spaces which breaks Chrome (returns an `ERR_SPDY_PROTOCOL_ERROR` error) when using HTTP/2. 

We discovered this while using an AWS Application Load Balancer (ALB) which automatically upgrades the connection between the browser and the ALB to HTTP/2. The connection between the ALB and the server was HTTP/1.1.

### Relevant RFCs
> HTTP header fields, which include general-header (section 4.5), request-header (section 5.3), response-header (section 6.2), and entity-header (section 7.1) fields, follow the same generic format as that given in Section 3.1 of RFC 822
-- https://tools.ietf.org/html/rfc2616#section-4.2

> The  field-name must be composed of printable ASCII characters (i.e., characters that  have  values  between  33.  and  126., decimal, except colon).
-- https://tools.ietf.org/html/rfc822#section-3.1.2

### Example
```
{-# LANGUAGE OverloadedStrings #-}

import Network.Wai (responseLBS, Application)
import Network.Wai.Handler.Warp (run)
import Network.HTTP.Types (status200)
import Network.HTTP.Types.Header (hContentType)

main = do
    let port = 80
    putStrLn $ "Listening on port " ++ show port
    run port app

app :: Application
app req f =
    f $ responseLBS status200 [(hContentType, "text/plain"), ("A A", "foo")] "Hello world!"
```

kazu-yamamoto:
Which is better?
- just not adding broken headers to responses
- throwing an error if broken headers are found

danclien:
I prefer failing fast by throwing an error if broken headers are found, but I know that can also break other people's code.

pbrinkmeier:
@kazu-yamamoto I could set up a PR that adds a middleware to `wai-extra` to address this? This middleware could also validate header values, both keys and values are not allowed to contain any control characters iirc (i can research the RFC, just need to hear if you'd be willing to merge something like that).

kazu-yamamoto:
@pbrinkmeier Sorry for the delay. Yes, please.

pbrinkmeier:
I will do it in the next few days, thanks for getting back to me :)

kazu-yamamoto:
Closing.

PR #990, opened 2024-05-13
Middleware for header validation

pbrinkmeier:
This PR addresses #628 and #770. It adds a middleware that throws a 500 when the application sets invalid response headers.

The definition of "invalid" varies between RFCs. I have chosen to use RFC 9110 as a reference, which is the most recent of the ones I researched that concern HTTP headers:

| Year | RFC                                                                                   | Field Name                                                                                     | Field Value                                                                                                                                                                                            |
| ---- | ------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1982 | [RFC 822 Section 3.1](https://datatracker.ietf.org/doc/html/rfc822#section-3.1)       | Printable ASCII except colon.                                                                  | ASCII except CR or LF (except folding).                                                                                                                                                                |
| 2001 | [RFC 2822 Section 2.2](https://datatracker.ietf.org/doc/html/rfc2822#section-2.2)     | Printable ASCII except colon.                                                                  | ASCII except CR or LF (except folding).                                                                                                                                                                |
| 2008 | [RFC 5322 Section 2.2](https://datatracker.ietf.org/doc/html/rfc5322#section-2.2)     | Printable ASCII except colon.                                                                  | Printable ASCII + SP + HTAB (except folding).                                                                                                                                                          |
| 2022 | [RFC 9110 Section 5.5](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values) | `token`, see e.g. [RFC 9110 Section 5.6.2](https://www.rfc-editor.org/rfc/rfc9110#name-tokens) | ASCII, SHOULD limit to visible ASCII, SP and HTAB. Especially no control characters, CR or LF. ISO-8859-1 is mentioned but not recommended. Recipients should treat octets `0x80-0xFF` as opaque data. |

This is also mentioned in the top level comment of `Network.Wai.Middleware.ValidateHeaders`. The middleware also enforces the SHOULD clause mentioned above, i.e. header values are only allowed to contain visible ASCII, space and horizontal tab.

## To Do List

After submitting your PR:

### DIFF
--- a/wai-extra/ChangeLog.md
+++ b/wai-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for wai-extra
 
+## 3.1.15
+
+* Added `validateHeadersMiddleware` for validating response headers set by the application [#990](https://github.com/yesodweb/wai/pull/990).
+
 ## 3.1.14
 
 * Request parsing throws an exception rather than `error`ing [#972](https://github.com/yesodweb/wai/pull/972):
--- a/wai-extra/Network/Wai/Middleware/ValidateHeaders.hs
+++ b/wai-extra/Network/Wai/Middleware/ValidateHeaders.hs
@@ -11,6 +11,9 @@ module Network.Wai.Middleware.ValidateHeaders
     -- * Settings
     , ValidateHeadersSettings (..)
     , defaultValidateHeadersSettings
+    -- * Types
+    , InvalidHeader (..)
+    , InvalidHeaderReason (..)
     ) where
 
 import Data.CaseInsensitive (original)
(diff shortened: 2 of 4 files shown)


kazu-yamamoto (approved):
Excellent!

kazu-yamamoto:
Rebased and merged.
Thank you for your contribution.
Should I release a new version?

pbrinkmeier:
Wow, that was quick! And yes, feel free to release the code. :rocket:

kazu-yamamoto:

ocramz:
Hi, I realize it's been a while since this was merged but I was thinking how to move https://github.com/scotty-web/scotty/issues/385 forward.

In this implementation, only _response_ headers are analyzed as far as I can see. So this safeguards against other parts of the server being implemented against the standard. But wouldn't it also make sense to sanitize the request headers in this same way?

Vlix:
I feel like `warp` would make sure only valid headers make it into the middleware and user space. I've reworked the header parse logic a while back to at least not accept newlines in headers (since RFC 9110 has done away with that), and wanted to do more in the header parsing area as well.

Don't know when I'll have time to do so, but the obvious question would also be: "what should the response to an invalid request header be?"
If RFC 9110 has a specification for it, I'd like to just stick to that. Otherwise, we'll have to decide what to do with invalid headers, but dropping offending headers and keeping the rest is probably a bad idea. So my intuition would say "return an appropriate 4XX response".

Vlix:
That or a `400 Bad Request`.
A 422 implies the syntax is fine, and unsupported characters in headers seems to me like "bad syntax", but maybe others have a different opinion?

PR #991, opened 2024-05-15
ensuring to receive 4 bytes.

### LINKED ISSUE #989: "is http2" detection doesn't seem quite correct
https://github.com/yesodweb/wai/blob/2a7399f5229f6a8ea675a20cf0cf8f69ba8b9f72/warp/Network/Wai/Handler/Warp/Run.hs#L390

Here we seem to determine whether this is an HTTP/2 request by reading some bytes off the wire, and if they are `PRI `, it is.

But there's no guarantee we read 4 bytes, of course. We could read `PR`, determine this isn't an HTTP/2 request, go down the HTTP/1 code path and proceed to read the remaining `I `!

kazu-yamamoto:
Fix #989.

@mitchellwrosen Would you review this PR?

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -386,7 +386,7 @@ serveConnection conn ii th origAddr transport settings app = do
         if isHTTP2 transport
             then return (True, "")
             else do
-                bs0 <- recv4
+                bs0 <- recv4 ""
                 if "PRI " `S.isPrefixOf` bs0
                     then return (True, bs0)
                     else return (False, bs0)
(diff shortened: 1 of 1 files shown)


mitchellwrosen (approved):

Vlix:
As long as `connRecv` always gets an amount of bytes > 0, this should be fine.
If it can produce 0 bytes, then this loop might be infinite (would that result in a timeout?)

What if only 3 bytes were sent?

---

And the [L389](https://github.com/yesodweb/wai/pull/991/files#diff-f6863cde0b4e33012b8c901c00d5c2aab1a58f877bfdb2f159fcf11a6a148812R389) `bs0 <- recv4` could be replaced with:

```hs
bs0 <- connRecv conn >>= loop
```

So you don't need the `recv4` function.

Vlix:
I'd use `S.null bs1` to check for empty ByteString, but it looks fine :+1: 

Also, when doing `<>` on ByteStrings where one is empty, it just drops it, so it's pretty much as close to efficient as can be. :+1: 

Using a literal `""` will do some unnecessary memory allocation though, it seems, so using `S.empty` would be more efficient (will just be a static `nullForeignPtr` which will be dropped at the first `<>`)

kazu-yamamoto:
During modification, `len` is used twice.
But in the current code, I agree that `null` is better.

Vlix (approved):

kazu-yamamoto:
Merged.
Thank you for reviewing!

PR #994, opened 2024-05-26
Update all stack files to fix CI

Vlix:
Trying to make CI work again. Also removed outdated stack files


Vlix:
Ok, few things:

- Ubuntu and Windows LTS 18 fails on a certain haddock parse (bit weird)
- `macos-latest` with LTS 18-19 doesn't work because of an LLVM version mismatch, it seems.
- `macos-latest` with LTS >= 20 fails on `gauge-0.2.5` (`/private/var/folders/2f/8t5k6yr535sdw0s4glnpxrzm0000gn/T/stack-9e612839270a55ff/gauge-0.2.5/cbits/cycles.c:55:2: error: 55 | #error Unsupported OS/architecture/compiler!`)

(I've turned off `--bench` on `macos-latest` to hopefully fix the 3rd issue)

kazu-yamamoto:
Nice!

Issue #980, opened 2024-03-21
auto-update: mkDebounce leaking the thread


Yuras:
`mkDebounce` internally [forks a thread](https://github.com/yesodweb/wai/blob/4d2f12c5b5812d64f9dae9872b56c9cc2fd51ec2/auto-update/Control/Debounce/Internal.hs#L88), but never kills it. To reproduce, run the following with `+RTS -s -RTS` and observe memory usage.

```haskell
import Control.Debounce
import Control.Monad

main :: IO ()
main = forM_ [(0::Int) .. 100000] $ \_ -> do
  r <- mkDebounce defaultDebounceSettings
  r
```

Output:
```
     250,129,600 bytes allocated in the heap
     417,814,184 bytes copied during GC
      47,018,288 bytes maximum residency (8 sample(s))
       3,325,648 bytes maximum slop
             132 MiB total memory in use (0 MB lost due to fragmentation)
```

132MiB memory is too much for the above program.

As a possible solution, consider attaching a finalizer to the debounced action that will kill the thread. I'd be happy to prepare a patch if you agree with such a solution.
Other possibility is to use a closable channel instead of `MVar`. I didn't try it though.

A bit of context: [fast-logger](https://hackage.haskell.org/package/fast-logger) uses `mkDebounce` to flush logs, so each logger comes with a thread which never exits.

Thanks!

Vlix:
I wonder how this could be fixed.

I guess `mkDebounce` could also provide a finalizer?

Or create a function that captures its own MVar and doesn't need to fork a thread?

```hs
mkDebounce :: IO () -> IO (IO ())
mkDebounce action = do
  baton <- newMVar ()
  pure $ do
    mBaton <- tryTakeMVar baton
    case mBaton of
      Nothing -> pure ()
      Just () -> do
        -- handle Leading/Trailing
        -- and `putMVar ()` when done with action + delay
```

@kazu-yamamoto would this idea be an ok replacement for `Control.Debounce.mkDebounce`? Or do you see any issues with it? (a `finally putMVar` might also be needed, come to think of it)

kazu-yamamoto:
This module was created by @snoyberg.
I don't have strong opinions but yes, cleaning up threads is a MUST.

Vlix:
Will try to reproduce with my PR #996 at the earliest convenience.

kazu-yamamoto:
Note that the recent GHC has `GHC.Conc.Sync.listThreads`.
[...]

PR #996, opened 2024-06-08
Reworking `mkDebounce`

Vlix:
Ok, so I've tried to rework the debouncing, because of issue #980, but it seems to fail on a test because the trigger `MVar` doesn't seem to be correctly filled unless I put a `trace` in a specific spot!?

Does anyone know how to force/evaluate (or whatever is happening here) the `tryPutMVar` without using `trace`?

- auto-update: reworked 'mkDebounce', and it should work... but 'Leading' only works if I put a 'trace' in a specific spot???

---

Still TODO:

After submitting your PR:


Vlix:
Hmmm, seeing the random failing in CI jobs, this is probably a timing issue? And the `trace "..."` printing to the screen is just takes enough time with the GHC version I was developing with? 🤔 

Now to figure out if the test should be adjusted, or if there is actually an error in the implementation...

Vlix:
I figured it out. It was the `trigger` MVar that should be emptied BEFORE forking.

I think this fixes the leaking of threads and adds two styles of debouncing that are also pretty common.

Vlix:
Windows nightly fails with:
`doctest > C:/Users/runneradmin/AppData/Local/Programs/stack/x86_64-windows/ghc-9.8.3/mingw/bin/strip.exe: error: Permission denied`

... who's fault would this be? :thinking:

Vlix:
Oh god, `data-default` breaks literally every CI... why is this even still a thing!?

kazu-yamamoto:
@Vlix Is this PR ready for reviewing?

Vlix:
@kazu-yamamoto Yes, I hope with this last commit (and `quic` building again) that CI is back to working condition.

Vlix:
Apparently my new `auto-update` is a bit flaky :/

Vlix:
When this is merged, I'll update the CI in #1011 to include `quic` again.

auto-update/Control/Debounce.hs
```diff
@@ -23,23 +23,27 @@
...
+
+    -- * Settings
     DI.DebounceSettings,
     defaultDebounceSettings,
 
-    -- * Accessors
+    -- ** Accessors
     DI.debounceFreq,
     DI.debounceAction,
     DI.debounceEdge,
```
kazu-yamamoto, line 36: I wonder why `DebounceEdge` is NOT exported while `deboundeEdge` is exported.

kazu-yamamoto: Never mind.
`leadingEdge` should be used instead of `Leading`.

kazu-yamamoto:
Rebased and merged.
Nice work!

PR #1002, opened 2024-08-28
cabal.project: Remove allow-newer and SRP stanzas

erikd:
Also enable building of test code by default and improve test output.

Before these changes the code in Git would not even compile with `ghc-9.6` (using `cabal`). With these changes it builds and all test pass both with `ghc-9.6` and `ghc-9.10`.

I would have updated the tests, but I do not use `stack` and have zero knowledge of it.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,53 +12,5 @@ packages:
     wai-conduit
     time-manager
 
--- These need hackage revisions but otherwise test fine in the repo
-allow-newer:
-      cryptohash-md5:base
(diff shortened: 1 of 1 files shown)


Vlix:
Thanks. We should probably add builds with cabal...

But yeah, the explicit `allow-newer` were from 2 years ago and for GHC 9.4 at the time, so if it works without it, I'd accept this change without issue.

Vlix (approved):

PR #1004, opened 2024-09-14
delay new connection in case of exausted files

### LINKED ISSUE #1003: busy loop on too many open files
```haskell
acceptNewConnection = do
    ex <- UnliftIO.tryIO getConnMaker
    case ex of
        Right x -> return $ Just x
        Left e -> do
            let eConnAborted = getErrno eCONNABORTED
                getErrno (Errno cInt) = cInt
            if ioe_errno e == Just eConnAborted
                then acceptNewConnection
                else do
                    settingsOnException set Nothing $ toException e
                    return Nothing
```

The issue is that this function only checks for the `eCONNABORTED` error specifically. If any other error occurs, including a "too many open files" error, it will call `settingsOnException` and then return `Nothing`, which causes `acceptLoop` to exit.

However, `acceptLoop` is called in a loop from `acceptConnection`, so it will immediately try to accept a new connection again. If the "too many open files" error persists, this will result in a busy loop, constantly trying to accept new connections and failing.

To fix this, you could:

1. Add a delay before retrying after encountering errors other than `eCONNABORTED`.
2. Implement a backoff strategy for repeated errors.
3. Handle the "too many open files" error specifically and take appropriate action (like waiting for some connections to close).

domenkozar:
Fixes #1003

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -13,7 +13,8 @@ import qualified Control.Exception
 import qualified Data.ByteString as S
 import Data.IORef (newIORef, readIORef)
 import Data.Streaming.Network (bindPortTCP)
-import Foreign.C.Error (Errno (..), eCONNABORTED)
+import Foreign.C.Error (Errno (..), eCONNABORTED, eMFILE)
+import GHC.Conc (threadDelay)
 import GHC.IO.Exception (IOErrorType (..), IOException (..))
 import Network.Socket (
     SockAddr,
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @kazu-yamamoto

kazu-yamamoto:
In the case of `eMFILE`, we should wait for the number of connections decreases using `counter` (`TVar Int`) as @nh2 suggested. (It seems to me that the delay approach is ad-hoc.)

@nh2 I don't understand what we should do for the case of `eCONNABORTED` yet. In my understanding, `accept` throws `eCONNABORTED` when the creation of a new connection fails. We can use the listen socket safely again. Is the current code good enough for the case of `eCONNABORTED`?

domenkozar:
Note that number of open files might not just depend on warp, but it can also be due to other things running in the same space.

Either way, I'm happy as long as it doesn't busy-loop or just crash.

PR #1011, opened 2024-11-03
No using data-default

Vlix:

After submitting your PR:

---

So I really want to get rid of any `data-default`, and now with `tls-2.1.3` we can at least get rid of the dependency on `warp-tls`. Though it requires `tls >= 2.1.3`, which will be a bit more restrictive, though I don't think many people will have problems because of this.

I've also tried my best to obviously deprecate the `Default` instances that are still in this repo. I've made a synonym for `defaultGzipSettings`, since `def` was also directly (re-)exported, so now we can give an actual Deprecation warning. Though we can't on the other instances... unfortunately.

I'm hoping come `wai-extra-3.3` these instances can be removed as well.

### DIFF
--- a/stack-lts-19.yaml
+++ b/stack-lts-19.yaml
@@ -12,7 +12,7 @@ packages:
 - ./wai-http2-extra
 - ./wai-websockets
 - ./warp
-# - ./warp-quic
+- ./warp-quic
 - ./warp-tls
 flags:
   wai-extra:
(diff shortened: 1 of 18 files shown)


kazu-yamamoto (approved):
I agree with this change.
Since this PR includes CI stuff, @Vlix should merge this PR by himself, I think.

Vlix:
Released new minor version for `wai-extra`, but aren't a maintainer for `warp-tls`. @kazu-yamamoto could you upload version `3.4.10`?

kazu-yamamoto:
I have released `warp-tls-3.4.10`.

@Vlix You should push the tag for `wai-extra` to github.

Vlix:
@kazu-yamamoto I did do that, didn't I?

Or do you mean the 3.1.16 tag? (seems I forgot that one)

PR #1012, opened 2024-11-07
Removing unliftio

kazu-yamamoto:
This PR removes `unliftio` from WAI families.
This is because TLS `bye` and network `gracefulClose` cannot used with `bracket` of `unliftio`.

In my field testing for a week, I have not found any problems.
This PR is just to run CI.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -56,7 +56,6 @@ module Network.Wai.Handler.WarpTLS (
 ) where
 
 import Control.Applicative ((<|>))
-import Control.Concurrent (newEmptyMVar, putMVar, takeMVar, forkIOWithUnmask)
 import Control.Exception (
     Exception,
     IOException,
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -56,6 +56,20 @@ module Network.Wai.Handler.WarpTLS (
 ) where
 
 import Control.Applicative ((<|>))
+import Control.Concurrent (newEmptyMVar, putMVar, takeMVar, forkIOWithUnmask)
+import Control.Exception (
+    Exception,
+    IOException,
+    SomeException (..),
+    bracket,
+    finally,
+    fromException,
+    handle,
+    handleJust,
+    onException,
+    throwIO,
+    try,
+ )
 import Control.Monad (guard, void)
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as L
@@ -73,6 +87,7 @@ import Network.Socket (
 #endif
     withSocketsDo,
  )
+import qualified Control.Exception as E
 import Network.Socket.BufferPool
 import Network.Socket.ByteString (sendAll)
 import qualified Network.TLS as TLS
@@ -82,23 +97,7 @@ import Network.Wai.Handler.Warp
 import Network.Wai.Handler.Warp.Internal
 import Network.Wai.Handler.WarpTLS.Internal
 import System.IO.Error (ioeGetErrorType, isEOFError)
-import UnliftIO.Exception (
-    Exception,
-    IOException,
-    SomeException (..),
-    bracket,
-    finally,
-    fromException,
-    handle,
-    handleAny,
-    handleJust,
-    onException,
-    throwIO,
-    try,
- )
-import qualified UnliftIO.Exception as E
-import UnliftIO.Concurrent (newEmptyMVar, putMVar, takeMVar, forkIOWithUnmask)
-import UnliftIO.Timeout (timeout)
+import System.Timeout (timeout)
 
 ----------------------------------------------------------------
 
@@ -366,7 +365,7 @@ httpOverTls TLSSettings{..} set s bs0 params =
         case mconn of
           Nothing -> throwIO IncompleteHeaders
           Just conn -> return conn
-    wrappedRecvN recvN n = handleAny (const mempty) $ recvN n
+    wrappedRecvN recvN n = handle (\(SomeException _) -> mempty) $ recvN n
     backend recvN =
         TLS.Backend
             { TLS.backendFlush = return ()
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -322,12 +322,8 @@ mkConn
     -> params
     -> IO (Connection, Transport)
 mkConn tlsset set s params = do
-    var <- newEmptyMVar
-    _ <- forkIOWithUnmask $ \umask -> do
-        let tm = settingsTimeout set * 1000000
-        mct <- umask (timeout tm recvFirstBS)
-        putMVar var mct
-    mbs <- takeMVar var
+    let tm = settingsTimeout set * 1000000
+    mbs <- timeout tm recvFirstBS
     case mbs of
       Nothing -> throwIO IncompleteHeaders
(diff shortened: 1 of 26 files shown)


warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -323,12 +321,8 @@ mkConn
...
     -> IO (Connection, Transport)
 mkConn tlsset set s params = do
-    var <- newEmptyMVar
-    _ <- forkIOWithUnmask $ \umask -> do
-        let tm = settingsTimeout set * 1000000
-        mct <- umask (timeout tm recvFirstBS)
-        putMVar var mct
-    mbs <- takeMVar var
+    let tm = settingsTimeout set * 1000000
+    mbs <- timeout tm recvFirstBS
```
Vlix, line 325: I think this makes sense, but if we want to keep the same behaviour we'd have to change it to:
```hs
mbs <- unsafeUnmask $ timeout tm recvFirstBS
```

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -366,7 +360,7 @@ httpOverTls TLSSettings{..} set s bs0 params =
         case mconn of
           Nothing -> throwIO IncompleteHeaders
           Just conn -> return conn
-    wrappedRecvN recvN n = handleAny (const mempty) $ recvN n
+    wrappedRecvN recvN n = handle (\(SomeException _) -> mempty) $ recvN n
```
Vlix, line 363: Again, this will now also catch (and ignore) async exceptions.

warp/Network/Wai/Handler/Warp/Conduit.hs
```diff
@@ -2,10 +2,10 @@
 
 module Network.Wai.Handler.Warp.Conduit where
 
+import Control.Exception (assert, throwIO)
```
Vlix, line 5: This `throwIO` is fine if we're only throwing syncronous exceptions.

warp/Network/Wai/Handler/Warp.hs
```diff
@@ -141,7 +141,7 @@ module Network.Wai.Handler.Warp (
 
 import Data.Streaming.Network (HostPreference)
 import qualified Data.Vault.Lazy as Vault
-import UnliftIO.Exception (SomeException, throwIO)
+import Control.Exception (SomeException, throwIO)
```
Vlix, line 144: This `throwIO` is fine if we're only throwing syncronous exceptions.

warp/Network/Wai/Handler/Warp/FdCache.hs
```diff
@@ -26,7 +26,7 @@ import System.Posix.IO (
     openFd,
     setFdOption,
  )
-import UnliftIO.Exception (bracket)
+import Control.Exception (bracket)
```
Vlix, line 29: The `UnliftIO` bracket is different in that it uses `uninterruptibleMask_` on the `after` actions. How important are the `terminate` actions?
They are cleaning up file descriptors, so that might be important enough to absolutely make sure it happens, right?

warp/Network/Wai/Handler/Warp/FileInfoCache.hs
```diff
@@ -7,14 +7,14 @@ module Network.Wai.Handler.Warp.FileInfoCache (
     getInfo, -- test purpose only
 ) where
 
+import Control.Exception (bracket, onException, throwIO)
```
Vlix, line 10: `UnliftIO.onException` also runs with `uninterruptibleMask`. Making sure async exceptions wait until after the cleanup is done.

warp/Network/Wai/Handler/Warp/Types.hs
```diff
@@ -7,7 +7,7 @@ module Network.Wai.Handler.Warp.Types where
 import qualified Data.ByteString as S
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
 import Data.Typeable (Typeable)
-import qualified UnliftIO
+import qualified Control.Exception as E
```
Vlix, line 10: `import Control.Exception as E (Exception, SomeException)`

Vlix (commented):
Overall, this change seems to me as a bit too invasive.

Changing to `Control.Exception` will change behaviour with regards to async exceptions. If we'd want to keep it similar without `unliftio`, we'd need to use `safe-exceptions`.
I'd only fall back to `Control.Exception` if we make sure the masking stays the same and async exceptions are handled the same. 

Unless anyone has a good argument (in every spot where `catch`, `throwIO`, `bracket`, etc. is used) as to why we don't have to.

kazu-yamamoto:
@Vlix Thank you for reviewing.
I don't want to use `safe-exception` as alternative `unliftio`.
This is because each library has its own semantics which I don't understand completely.
One example is `timeout` cannot be used in `bracket` of `unliftio`  which suffered me very much.
I would like to use only `Control.Exception` since it is well-documented and I know it well.
Even we would adopt `safe-exception`, we cannot let all other libraries to introduce it.
The mixed situation is anxious.
I would like to set up a proper handler for `Control.Exception` even though it is boring.
I believe it saves my time In the long run.
I will check your comments above tomorrow.

Vlix:
> I would like to set up a proper handler for `Control.Exception` even though it is boring.

I can understand that.
I'd argue that we should try to mimic what `unliftio` did and change it in the situations where the `unliftio` style is not what we want.

kazu-yamamoto:
I remembered one this.
So, I write this up just for record.

Some people believe that `unliftio` makes our libraries less-buggy.
But the reality is that `unliftio` can introduce new bugs.
An example is that `unliftio` made `makeConn` of `warp-tls` uninterruptible.
The fix was 371279f.

Vlix:
Aaaah, that's why!

Ok, so what I think we should probably do:
* identify the sections we want to be uninterruptible
* identify which catches should let async exceptions through

Adding comments at those sections describing why we're using regular `Control.Exception.catch` or `uninterruptibleMask` or whatever would also be good for posterity.

warp-tls/Network/Wai/Handler/WarpTLS.hs
kazu-yamamoto: The original code before unsafeUnmask does not have `unsafeUnmask`.
Is it really necessary?

warp/Network/Wai/Handler/Warp/FdCache.hs
kazu-yamamoto: Resouce-release actions in `bracket` (i.e. `mask`ed actions) are iterruptible only if they are blocked (such as `takeMVar`).
Non-blocking actions are NOT iterruptible.
In other words, asynchronous exceptions are not delivered to resource-release actions if they are non-blocking.

`terminate` is non-blocking, so it's free from asynchronous exceptions, I believe.

warp-tls/Network/Wai/Handler/WarpTLS.hs
Vlix: The `umask` in the diff is basically `unsafeUnmask`. But as you described earlier, that `mkConn` should be interruptable, this is unnecessary if the `bracket` (or whatever is used) is from `Control.Exception` instead of from `UnliftIO`.

Because the whole `forkWithUnmask` was a hack to get around the `unliftio` functions that `mask` more than necessary.

So no, I have changed my stance and adding `unsafeUnmask` is indeed not necessary.

warp/Network/Wai/Handler/Warp/FdCache.hs
Vlix: Ah, yes. Never mind. I've re-read the `bracket` docs, and it does. I misread, I think.

Issue #1014, opened 2024-11-20
Asynchronous exeptions


kazu-yamamoto:
## Rule 1: don't catch asynchronous exceptions of others

Exceptions of others have their own semantics. If you catch them, your code does not work well. For instance, if you use the `async` library and catch `AsyncCancelled`, a thread tree created by `concurrently`  cannot be destroyed.

If you use `SomeException` for `catch`, `handle` or `try`, check if the caught exception is asynchronous. And re-throw it via `throwIO` if asynchrnous. You can use the following code to check asynchronous or not.

```haskell
isAsyncException :: Exception e => e -> Bool
isAsyncException e =
    case fromException (toException e) of
        Just (SomeAsyncException _) -> True
        Nothing -> False
```

## Rule2: catch the asynchronous exception which you are using

If you don't catch, the exceptions are leaked from your threads. GHC RTS catches them and displays them to sdout.

If you are using `time-manager` and don't want to leak `TimeoutThread`, catch it.

PR #1013, opened 2024-11-13
Rethrow async exception

kazu-yamamoto:
@Vlix This is the first step to fix the stuffs in #1012.

PR #1015, opened 2024-11-21
withHandle

kazu-yamamoto:
This implements rule 2 of #1014.
Relating to #1013.

### DIFF
--- a/time-manager/System/TimeManager.hs
+++ b/time-manager/System/TimeManager.hs
@@ -14,15 +14,19 @@ module System.TimeManager (
     withManager,
     withManager',
 
-    -- ** Registration
-    register,
-    registerKillThread,
-    cancel,
+    -- ** Registering a timeout action
+    withHandle,
+    withHandleKillThread,
 
     -- ** Control
     tickle,
     pause,
     resume,
+
+    -- ** Low level
+    register,
+    registerKillThread,
+    cancel,
 
     -- ** Exceptions
     TimeoutThread (..),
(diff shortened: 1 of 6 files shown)


time-manager/System/TimeManager.hs
```diff
@@ -96,6 +100,22 @@ killManager = reaperKill
...
+--   when the body action is finished.
+withHandle :: Manager -> TimeoutAction -> (Handle -> IO a) -> IO a
+withHandle mgr onTimeout action =
+    E.bracket (register mgr onTimeout) cancel action
+
+-- | Registering a timeout action of killing this thread and
+--   unregister its handle when the body action is killed or finished.
+withHandleKillThread :: Manager -> TimeoutAction -> (Handle -> IO ()) -> IO ()
+withHandleKillThread mgr onTimeout action =
+    E.handle handler $ E.bracket (registerKillThread mgr onTimeout) cancel action
```
Vlix, line 113: Would it be better/worse to do the following? :thinking: 

    E.bracket (registerKillThread mgr onTimeout) cancel $ E.handle handler action

warp/Network/Wai/Handler/Warp/HTTP2.hs
```diff
@@ -106,10 +106,6 @@ http2server label settings ii transport addr app h2req0 aux0 response = do
             logResponse req st msiz
             mapM_ (logPushPromise req) pps
         Left e
-          -- killed by the local worker manager
-          | Just E.ThreadKilled  <- E.fromException e -> return ()
```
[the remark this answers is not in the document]
kazu-yamamoto: I think this is very old code.
Currenlty, `killThread` is not used anywhere.

time-manager/System/TimeManager.hs
kazu-yamamoto: Interesting.
I don't know which is better.

warp/Network/Wai/Handler/Warp/HTTP2.hs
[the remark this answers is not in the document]
Vlix: But people can implement their own forking behaviour with `settingsFork`. So it COULD happen that a `ThreadKilled` is thrown to this thread.

Though now that I think about it, this is actually better, right? Since if someone would want to kill an HTTP thread that's handling a request, this would just ignore the signal. So I now also believe this is the right decision.

time-manager/System/TimeManager.hs
Vlix: I don't think it will matter much... since the `cancel` will be run either way.

The difference is this, effectively:

```hs
-- Here's the body of 'bracket'

-- either here instead of:
-- mask $ \restore -> do
E.handle handler $ mask $ \restore -> do
  a <- before
  -- or here instead of:
  -- r <- restore (thing a) `onException` after a
  r <- restore (E.handle handler $ thing a) `onException` after a
  _ <- after a
  return r
```

Vlix (approved):

kazu-yamamoto:
Thank you for reviewing.

PR #1016, opened 2024-12-04
Export `KilledByThreadManager` exception

edsko:


edsko:
I think CI failures must be unrelated to this PR, since I literally just added an export.

Vlix:
Yes, the `time-manager` was bumped to 2.0, and `http2`, the dependency of `warp`, uses it but that version hasn't been added to the stack.yml files used for the tests.
If you could add `http2-5.3.9` to the `extra-deps` of all the `stack*.yml`s (except `stack-nightly.yml`), that'd be great 😅

kazu-yamamoto:
@edsko Would you tell me why this exception should be exported? It is used only internally...

@Vlix I noticed this failure of CI and am trying to fix it. The fix should be done independent on this PR. So, @edsko does not have to tackle the CI issue.

kazu-yamamoto:
CI has been fixed.

edsko:
> Would you tell me why this exception should be exported? It is used only internally...

Specifically, in the `http2` case, this exception is thrown to server handlers, which are running user-defined code (my code), and `http2` sometimes kills those handlers when it considers them no longer required (when a client disappears), but I need to allow them to continue to run and finish what they were doing, so instead I catch the exception, mark the client as having disappeared in case the handler still tries to communicate, but allow the handler to continue to run to completion.

kazu-yamamoto:
OK!
Rebased and merged.
Also, a new version has been released.

edsko:
Thanks @kazu-yamamoto !

PR #1017, opened 2024-12-11
Providing NoManager

kazu-yamamoto:
"initialize" with non positive integer creates "NoManager".

@edsko What do you think?


edsko:
I'll give this a shot and let you know :)

kazu-yamamoto:
Thanks.

PR #1018, opened 2024-12-12
Thread less autoupdate

kazu-yamamoto:
@khibino I'm trying to merge your version of `AutoUpdate`.

- Which module should export `mkClosableAutoUpdate`? (`AutoUpdate` or `Internal`?)
- Could you implement `mkAutoUpdateWithModify`, too?
- Any other concerns?


khibino:
> Could you implement `mkAutoUpdateWithModify`, too?

If my understanding is correct,
the action passed additionally to `mkAutoUpdateWithModify` will always be used
in place of the `updateAction` in `UpdateSettings` for update action calls starting from the second call.

Based on that premise, I implemented it as follows.

https://github.com/khibino/wai/compare/extend-thread-less~3..extend-thread-less

https://github.com/khibino/wai/tree/extend-thread-less

khibino:
> Which module should export `mkClosableAutoUpdate`? (`AutoUpdate` or `Internal`?)

Considering exporting the combination of `mkAutoUpdateWithModify` and `mkClosableAutoUpdate`,
it seems good to keep the exports from `AutoUpdate` simple,
and handle complex cases through exports from `Internal`.

On the other hand, while "thread-less" eliminates the thread leak problem when stopping auto-update midway,
there is a minor issue where timeout registration in GHC's `TimeManager` remains if `mkClosableAutoUpdate` is not used.

kazu-yamamoto:
@khibino I cherry-picked your commits.
Do you agree with merging this branch as is?

khibino:
Yes. I think that's good.

kazu-yamamoto:
@khibino The current implementation cannot be built on Windows because `GHC.Event` cannot be imported.

kazu-yamamoto:
@khibino I will use the old implementation for Windows.

kazu-yamamoto:
@khibino CI passes. Would you give one more look?

khibino:
informative link:
original code of thread-less auto-udpate
https://github.com/khibino/dnsext/blob/44cc84ae472d75c8fcdd5ecda083c99e8aeb1cca/dnsext-utils/DNS/Utils/AutoUpdate.hs

PR #1020, opened 2024-12-20
Using the thread version of AutoUpdate for non-threaded RTS

### LINKED ISSUE #1019: `auto-update-0.2.5` - a warning
With the update to `0.2.5` we've seen some errors like 
```
 test: GHC.Internal.Event.Thread.getSystemTimerManager: the TimerManager requires linking against the threaded runtime
             error, called at libraries/ghc-internal/src/GHC/Internal/Event/Thread.hs:290:13 in ghc-internal:GHC.Internal.Event.Thread
           HasCallStack backtrace:
             collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:92:13 in ghc-internal:GHC.Internal.Exception
             toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
             throwIO, called at libraries/ghc-internal/src/GHC/Internal/Control/Exception/Base.hs:195:43 in ghc-internal:GHC.Internal.Control.Exception.Base
```
during stackage test-suite execution - however not directly on the `auto-update` tests, but on dependent projects like `http-reverse-proxy` - unfortunately we lack a clear reproduction, for now we've noticed the problem only on the stackage server (also only in complete build cycles, running tests of a single package succeed) - also I failed to reproduce things like that on my local machine. But at least the problems occur consistent and on lts and nightly, so with ghc-9.8.4 and ghc-9.10.1. Also I believe that this might be the root cause for other problems that start to occur the same time like:
```
(ConnectionFailure Network.Socket.connect: <socket: 6>: unsupported operation (Cannot assign requested address))
           HasCallStack backtrace:
             collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:92:13 in ghc-internal:GHC.Internal.Exception
             toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
             throwIO, called at ./Network/HTTP/Client/Core.hs:214:29 in http-client-0.7.17-IncG67QVij97hnYFCn3f1x:Network.HTTP.Client.Core
```

For now we "resolved" the problem by expecting all affected test-suites to fail (https://github.com/commercialhaskell/stackage/commit/ef39034f068b37bff2d68f31b735c1627a3cb2ce).

Sorry that we couldn't provide a clearer picture of the problem.

kazu-yamamoto:
This should fix #1019.

### DIFF
--- a/auto-update/Control/AutoUpdate.hs
+++ b/auto-update/Control/AutoUpdate.hs
@@ -45,9 +45,26 @@ module Control.AutoUpdate (
 )
 where
 
+import Control.AutoUpdate.Types
 #ifdef mingw32_HOST_OS
 import Control.AutoUpdate.Thread
 #else
-import Control.AutoUpdate.Event
+import qualified Control.AutoUpdate.Event as Event
+import qualified Control.AutoUpdate.Thread as Thread
+
+import GHC.Event
+
+mkAutoUpdate :: UpdateSettings a -> IO (IO a)
+mkAutoUpdate settings = do
+    mmgr <- getSystemEventManager
+    case mmgr of
+      Nothing -> Thread.mkAutoUpdate settings
+      Just _m -> Event.mkAutoUpdate settings
+
+mkAutoUpdateWithModify :: UpdateSettings a -> (a -> IO a) -> IO (IO a)
+mkAutoUpdateWithModify settings f = do
+    mmgr <- getSystemEventManager
+    case mmgr of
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
@alaendle Could you review this PR?

Vlix (approved):
From what I can make out, this would also work, yes.

Issue #859, opened 2021-10-21
warp-tls: Issues with "ability to load TLS credentials from IORefs"


bitc:
@vdukhovni left some very insightful comments on the pull request *Add the ability to load TLS credentials from IORefs*:

https://github.com/yesodweb/wai/pull/806#pullrequestreview-483039614

The comments were not addressed by the authors of warp-tls, but I believe that that they are valid points that reveal shortcomings in the current implementation. I recommend reading the original comments since it is very well explained. My understanding is that there are 2 issues:

1. A single `IORef` should be used instead of multiple `IORef`s (to avoid race condition by allowing atomic update)
2. The API should encourage proper use of certificate chains

Thank you

tmcl:
As far as I know, the current design is totally unusable. Aside from the issues mentioned above, also @vdukhovni's solution will be problematic because it would require parsing the certificate/key/chain on every single load - IORef is not event-driven, so there's no way to say "whenever the IORef is updated, parse it and store the results in this IORef". Instead, you would want something that takes an `IORef TLS.Credentials` and you'd need to change the tls module in order to filter it through. 

But fortunately, there's already an ever so slightly more powerful interface basically where you need it - in the tls onServerNameIndication hook. `tls` currently consults the SNI in the ClientHello, calls the onServerNameIndication and prepends the results to the preloaded credentials. Then it picks the first credential compatible with the client's needs.

I would therefore recommend deprecating this whole interface with a view to removing it, and directing users to either use `tlsSettings(Chain)Mem` (if they don't care about reloading certificates, the current experience) or following a pattern like this if they do:

```
makeMyReloadableTlsSettings :: MyServerSettings -> IO TLSSettings
makeMyReloadableTlsSettings serverConfig = do
  iorServerCreds <- newIORef =<< either fail (pure . Network.TLS.Credentials . pure) =<< credentialLoadX509 serverConfig.serverCert serverConfig.serverKey
  setUpMyCredentialReloading iorServerCreds serverConfig
  pure defaultTlsSettings 
    { tlsCredentials = Just (Network.TLS.Credentials []) -- necessary to ignore the CertSettings
    , tlsServerHooks = defaultTlsSettings.tlsServerHooks
      { onServerNameIndication =  \_mHostname -> readIORef iorServerCreds
      }
    }
```

vdukhovni:
[...]

PR #1024, opened 2025-03-11
deprecating functions relating to CertFromRef

kazu-yamamoto:
See #859

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -176,6 +176,8 @@ tlsSettingsRef cert key =
         { certSettings = CertFromRef cert [] key
         }
 
+{-# DEPRECATED tlsSettingsRef "Will be removed in the next major release" #-}
+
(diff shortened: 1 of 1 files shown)


warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -176,6 +176,8 @@ tlsSettingsRef cert key =
         { certSettings = CertFromRef cert [] key
         }
 
+{-# DEPRECATED tlsSettingsRef "Will be removed in the next major release" #-}
```
Vlix, line 179: I'd like to read _why_ something is deprecated when I get warnings and possibly also what to do instead of the deprecated functionality.

Could the reason that this function is deprecated be added, at least?
(Same with the other change)

kazu-yamamoto: Indeed. But I think that @tmcl or @vdukhovni are better to write up the background than me.

kazu-yamamoto:
Yes. Let's close.

PR #1026, opened 2025-04-07
build: fix depend on stack build and update

ncaq:
- **build: add crypton-x509-validation dependency to stack.yaml**
- **build: move to lts-22 config file to `stack-lts-22.yaml`**
- **build: update stack.yaml to use lts-23**
- **ci: add lts-23 and update lts-22 configuration**


stack-lts-22.yaml
```diff
@@ -0,0 +1,38 @@
+resolver: lts-22.40
```
Vlix, line 1: Would prefer this to be the latest iteration of `lts-22` (i.e. 22.43) but it's not that big of a deal. We mostly use the LTS-es for making sure it builds with different GHC versions.

Vlix (approved):

stack-lts-22.yaml
ncaq: I was also considering updating this. I've updated it.

PR #1029, opened 2025-04-09
Use `:seti` instead of `:set` in doctests

sol:
I'm not going into details, but the short story is that `:set` frequently leads to issues, as it may unload modules.

You should always use `:seti` in doctests, unless you have a very specific reason not to.

### DIFF
--- a/auto-update/Control/AutoUpdate/Event.hs
+++ b/auto-update/Control/AutoUpdate/Event.hs
@@ -65,7 +65,7 @@ mkAutoUpdateThingsWithModify mk settings update1 = do
 --------------------------------------------------------------------------------
 
 -- $setup
--- >>> :set -XNumericUnderscores
+-- >>> :seti -XNumericUnderscores
 -- >>> import Control.Concurrent
 
 -- |


kazu-yamamoto:
@khibino What do you think?

khibino:
What I've experienced was when resolving mysterious "not in scope" errors.
https://github.com/kazu-yamamoto/dnsext/commit/74a978e1646e0bff4a68bbc1bbd7794d9746f535

At that time, I discovered https://github.com/sol/doctest/issues/327#issuecomment-1405603806 and
started using `:seti -XOverloadedStrings`.

kazu-yamamoto:
@khibino So, do you mean you approve this PR?

khibino:
I think it's better to approve this PR for the following reasons.
Until I read the following description in the user's guide, I wasn't aware that
the `:set` command also changes options for modules that have already been loaded.
In other words, if there is no intention to change options for already loaded modules, it can be considered always safer to use `:seti`.

https://downloads.haskell.org/ghc/9.6.7/docs/users_guide/ghci.html?highlight=seti#setting-options-for-interactive-evaluation-only

"It is often useful to change the interactive options, without having that option apply to loaded modules too."

"If you are setting language options in your `.ghci` file, it is good practice to use `:seti` rather than `:set`, unless you really do want them to apply to all modules you load in GHCi."

kazu-yamamoto (approved):
by @khibino

kazu-yamamoto:
@sol If you need a new release, please let me know.

sol:
Thank you for the quick review and for approving the PR, @kazu-yamamoto and @khibino.

I don't need a new release—my main goal here was simply to raise awareness about the potential pitfalls of using `:set` in doctests. There are a few more occurrences of `:set` in the codebase that would benefit from being switched to `:seti`.

Going forward, I encourage using `:seti` consistently in examples and tests to help set the right pattern for others, especially less experienced developers who might learn by reading this code.

Thanks again!

PR #1031, opened 2025-04-21
Fix recv

### LINKED ISSUE #1030: recv: when the size is insufficient, the last chunk is skipped
When the read size is insufficient, isn't the skipping of the last chunk not as intended?

script to observe the behavior, under `recv` dir:

```
 % ghci
:load Network/Socket/BufferPool/Recv.hs
:seti +m
:seti -XOverloadedStrings
import Data.IORef
import Data.ByteString
import Data.String
:{
mockRecv ini = do
    store <- newIORef ini
    let eof        = writeIORef store (error "closed") >> return ""
        chunk x xs = writeIORef store xs               >> return x
    return $ readIORef store >>= \xxs -> case xxs of { []    -> eof ;  x:xs  -> chunk x xs }
:}
```

```
ghci> tryRecvN "a" 3 =<< mockRecv ["bcd"]
("abc","d")
ghci> tryRecvN "a" 3 =<< mockRecv ["bc"]
("abc","")
ghci> tryRecvN "a" 3 =<< mockRecv ["b"]
("","")
ghci>
```

this is the section that seems problematic

https://github.com/yesodweb/wai/blob/1b68663c2c3f5706939d8774be85ca1dbbd1180e/recv/Network/Socket/BufferPool/Recv.hs#L57-L59

kazu-yamamoto:
Should fix #1030.

### DIFF
--- a/recv/Network/Socket/BufferPool/Recv.hs
+++ b/recv/Network/Socket/BufferPool/Recv.hs
@@ -30,6 +30,14 @@ receive sock pool = withBufferPool pool $ \ptr size -> recvBuf sock ptr size
 --   When N is less than equal to 4096, the buffer pool is used.
 --   Otherwise, a new buffer is allocated.
 --   In this case, the global lock is taken.
+--
+-- >>> :seti -XOverloadedStrings
+-- >>> tryRecvN "a" 3 =<< _iorefRecv ["bcd"]
+-- ("abc","d")
+-- >>> tryRecvN "a" 3 =<< _iorefRecv ["bc"]
+-- ("abc","")
+-- >>> tryRecvN "a" 3 =<< _iorefRecv ["b"]
+-- ("ab","")
 makeRecvN :: ByteString -> Recv -> IO RecvN
 makeRecvN bs0 recv = do
     ref <- newIORef bs0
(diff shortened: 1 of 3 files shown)


recv/Network/Socket/BufferPool/Recv.hs
```diff
@@ -69,9 +79,29 @@ tryRecvN init0 siz0 recv
...
+    ref <- newIORef ini
+    return $ recv ref
+  where
+    recv ref = do
+        xxs <- readIORef ref
+        case xxs of
+            [] -> do
+                writeIORef ref $ error "closed"
+                return ""
+            x : xs -> do
```
Vlix, line 105: I know this is only used in a doctest, but technically, this might return an `empty ByteString` prematurely if one of the elements is empty.

```hs
    case xxs of
        [] -> do
            writeIORef ref $ error "closed"
            return ""
        x : xs -> do
            writeIORef ref xs
            if BS.null x
                -- if an element is empty, keep going
                then recv ref
                else return x
```

Vlix (approved):
Checked it again to be sure, and looks good 👍 
Did find something in `_iorefRecv`, but fixing that could also be skipped since it's only for doctests.

Vlix:
This will also need a minor version bump and an entry to the ChangeLog before we can release this fix, though.

recv/Network/Socket/BufferPool/Recv.hs
kazu-yamamoto: A comment is added.

kazu-yamamoto:
`ChangeLog.md` is added and version bumps up.

kazu-yamamoto:
Thank you all!

PR #1032, opened 2025-05-06
Added ability to have IO action returning (LBS.ByteString, LBS.ByteString) to OnInsecure

liamzee:

- [✅] Bumped the version number
- [✅] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html)
- [✅] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock

After submitting your PR:

- [✅] Update the Changelog.md file with a link to your PR
- [✅] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

I'd like to be able to force the user on the Warp-TLS (instead of at reverse proxy or DNS) level to send a redirect response to users. Adding IO also makes it possible to provide logging behavior on the Warp-TLS level, although reverse proxies and firewalls would be superior at this.

### DIFF
--- a/warp-tls/ChangeLog.md
+++ b/warp-tls/ChangeLog.md
@@ -1,5 +1,10 @@
 # ChangeLog
 
+* Introduced additional constructor `DenyInsecureWithAction` for `OnInsecure`,
+  allowing the insertion of a custom function to consume the request bytestring,
+  returning a strict header bytestring and a lazy body bytestring for the given request.
+  [#1032](https://github.com/yesodweb/wai/pull/1032)
+
 ## 3.4.13
 
 * Introduced new smart constructor `tlsSettingsSni` to make it more convenient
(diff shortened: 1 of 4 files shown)


liamzee:
Actually, if a maintainer is interested in this work, maybe a pause on the PR would be nice, since CI seems a bit sparse.

liamzee:
Builds with a bit of hacking on system (there are no tests). Haddocks seem okay upon review.

The only question is whether the split return type (strict header, lazy bytestring in Strict.ByteString -> IO (Strict.ByteString, Lazy.ByteString) ) is desirable (it matches the procedure used for DenyInsecure). Please advise.

liamzee:
Closing. I'd like to poll what the maintainers think; perhaps WAI-style Request-Response might be better and might provide for better integration with the rest of the ecosystem.

liamzee:
See: https://github.com/yesodweb/wai/issues/1033

Vlix:
By the by, you can also turn this PR into a draft to put it on hold, instead of closing it 😉

PR #1035, opened 2025-05-28
correct seconds -> microseconds

chris-martin:
These arguments are microseconds, not seconds; the comments and identifiers are misleading.


Vlix (approved):
Looks good. These were indeed wrongly described as seconds 👍

kazu-yamamoto (approved):
Nice catch!

kazu-yamamoto:
Merged.

PR #1038, opened 2025-06-27
Handle quoted strings within Content-Disposition

### LINKED ISSUE #1037: Semicolons in quoted content-disposition attributes parse incorrectly
`parseRequestBody` doesn't properly handle semicolons within quoted strings in multi-part forms.  Consider this body:

```
------WebKitFormBoundaryB1pWXPZ6lNr8RiLh
Content-Disposition: form-data; name="yaml"; filename="semi; colon;"
Content-Type: application/octet-stream
etc
------WebKitFormBoundaryB1pWXPZ6lNr8RiLh
```

This results in a `fileName` of `semi;`, but it should result in `semi; colon;`.

spencerjanssen:
Resolves #1037

After submitting your PR:


wai-extra/Network/Wai/Parse.hs
```diff
@@ -796,17 +796,35 @@ sinkTillBound bound iter seed0 src max' = do
...
+ where
+    nonTokenChars = [_semicolon, _equal]
+    dropSpace = S.dropWhile (== _space)
+    parseTokenValues input | S.null input = []
+    parseTokenValues input =
+        let (token, rest) = parseToken $ dropSpace input
+         in case S.uncons rest of
+            Just (c, rest')
+                | c == _equal -> 
+                    let (value, rest'') = parseValue $ dropSpace rest'
```
Vlix, line 810: The `dropSpace` here is superfluous, since `parseValue` will also do `dropSpace`.

wai-extra/Network/Wai/Parse.hs
```diff
@@ -796,17 +796,35 @@ sinkTillBound bound iter seed0 src max' = do
...
+            _ -> S.break (`elem` nonTokenChars) $ dropSpace input
+    parseQuotedString acc input =
+        let (prefix, rest) = S.break (`elem` [_quotedbl, _backslash]) input
+         in case S.uncons rest of
+            Just (c, rest')
+                | c == _quotedbl -> (S.concat $ reverse (prefix:acc), rest')
+                | c == _backslash ->
+                    let (slashed, postSlash) = S.splitAt 1 rest'
+                     in parseQuotedString (slashed:prefix:acc) postSlash
+            _ -> (prefix, rest)
```
Vlix, line 827: This doesn't use the acc, so there's potential for loss of information.

I think you'll lose information if the quoted string doesn't terminate and there's a backslash in there, but it doesn't terminate with a double quote.
Now this is obv bad, but trying to be lenient, and at least just giving the entire information back would be better than only giving the string after the backslash. (i.e. if I read this right, the following header example will only give "only this will be returned, I think" as the `filename`)

`Content-Disposition: form-data; name="yaml"; filename="this will be dropped, \!only this will be returned\r\n`
Please add a test that makes sure the filename from the above header results in:
`this will be dropped, !only this will be returned`

Vlix (changes requested):
Better than the current implementation. That's certain.
I have a few parts that could be improved, but on the whole, looks good.

Vlix:
Sorry, could you change all the `http3-0.0.22` entries in the `extra-deps` of the `stack*.yaml` files to `http3-0.1.0`?

kazu-yamamoto:
@spencerjanssen `quic` v0.2.16 has been released. This should fix your issue.

spencerjanssen:
This last commit attempts to fix the older LTS Windows builds.

Vlix:
Thank you for all the effort and patience 🙏

Vlix:
```
quic > C:\Users\runneradmin\AppData\Local\Temp\stack-c3bda9f70685c8f5\quic-0.2.15\cbits\sysevent.c:1:10: error:
quic >      fatal error: 'sys/socket.h' file not found
quic >   |
quic > 1 | #include <sys/socket.h>
quic >   |          ^
quic > 
quic > #include <sys/socket.h>
quic >          ^~~~~~~~~~~~~~
quic > 1 error generated.
quic > `clang.exe' failed in phase `C Compiler'. (Exit code: 1)
```
🤦 This just keeps going, doesn't it

kazu-yamamoto:
I ran the failed test again.
But the result is the same.
@Vlix The test failed but I cannot see the issue  of `sys/socket.h`.
Do you still see this issue?

Vlix:
> Do you still see this issue?

No, they all fail because of the following now:

```
zlib > Error: Cabal-simple_O_XXXXX_X.X.X.X_ghc-X.X.X.exe: The program 'pkg-config'
zlib > version >=0.9.0 is required but it could not be found.
```

kazu-yamamoto:
I have merge this PR at 0999faa.
@spencerjanssen Thank you for your contribution!
If you want a new release, please let us know.

PR #1039, opened 2025-07-16
Network.Wai.Middleware.Approot: use lookupEnv instead of getEnvironment

mpscholten:
`lookupEnv` should be more efficient then retrieving all headers using `getEnvironment` and then using `lookup` on the list of all headers.


Vlix (approved):

mpscholten:
@kazu-yamamoto is this good to go?

kazu-yamamoto:
@mpscholten I don't touch the internal of `wai-extra`.
Probably, @Vlix could answer.

Vlix:
Looks good, no downsides.

PR #1041, opened 2025-08-25
Fix build ci failures

Vlix:
Trying to get CI working again, so had to bump all the versions in `stack*.yaml`s again. And now that `warp-quic` requires `random >= 1.3`, it clashes with `wai-websockets`, since it requires `websockets`, which still has `random < 1.3`.

@kazu-yamamoto was it _necessary_ to limit `random >= 1.3`? Or could we have allowed a wider range? Because now our CI is broken until `websockets` allows `random < 1.4`. And that makes me kind of sad :(

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,11 @@ jobs:
   stack:
     name: Stack
     runs-on: ${{ matrix.os }}
+    env:
+      STACK_PATH_GLOBAL_WINDOWS: "C:\\sr"
+      STACK_PATH_GLOBAL_UNIX: "~/.stack"
+      STACK_PATH_LOCAL: ".stack-work"
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


Vlix:
Only error I couldn't fix is on `windows` runners:

```
zlib > Error: Cabal-simple_O_vy6YIf_3.10.3.0_ghc-9.6.7.exe: The program 'pkg-config'
zlib > version >=0.9.0 is required but it could not be found.
```

Vlix:
Ok, most are finishing in like ~6 min because they only need to do tests and haddock building 👍 
Changing stack YAMLs will invalidate cache, and for local files, all the *.hs files are hashed, so any change to code will also make a new cache. (local files cache will be restored of a cache that is at least the same `OS/LTS/STACKYAML` so even then most libraries will probably not need to be rebuilt :) )

kazu-yamamoto:
Sorry for the delay. `random` 1.3 is necessary for its new APIs.
@Vlix Would you merge this PR by yourself?

PR #1042, opened 2025-08-27
Use Cabal PackageInfo_warp instead of Paths_warp

lrworth:
Use of `Paths_warp` causes a large dependency closure when building with nix. See https://github.com/NixOS/nixpkgs/pull/264434 for example.

This reduces the closure size of our application from 756Mb to 53Mb by virtue of not including GHC and GCC.

It makes a significant change to `cabal-version` and I'm not sure what compatibility we're trying to maintain in this project; also I made some incidental changes so that it would build with GHC 9.12.

This is not merge-ready — the PR is meant to open a discussion about using `PackageInfo_`.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP2.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2.hs
@@ -152,7 +152,7 @@ wrappedRecvN th slowlorisSize readN bufsize = do
     handler = throughAsync (return "")
 
 -- connClose must not be called here since Run:fork calls it
-goaway :: Connection -> H2.ErrorCodeId -> ByteString -> IO ()
+goaway :: Connection -> H2.ErrorCode -> ByteString -> IO ()
 goaway Connection{..} etype debugmsg = connSendAll bytestream
   where
     einfo = H2.encodeInfo id 0
--- a/warp/Network/Wai/Handler/Warp/HTTP2/Request.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP2/Request.hs
@@ -85,13 +85,13 @@ toRequest' ii settings addr ref (reqths, reqvt) bodylen body th transport =
             Nothing -> case mAuth of
                 Just auth -> (tokenHost, auth) : reqths
                 _ -> reqths
-    !mPath = getHeaderValue tokenPath reqvt -- SHOULD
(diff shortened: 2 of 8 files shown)


Vlix (changes requested):
Looks fine, though does the `cabal-version` bump mean it won't build with GHC 9.10 anymore? That might be unacceptable.

We'd need to add conditionals etc. to add the `Path/PackageInfo` modules.

Vlix:
Also please merge the current `master` branch, as it has fixed some CI issues.

lrworth:
The lts-21 / GHC 9.4 configuration listed in CI doesn't work:

```
$ stack test --system-ghc --resolver lts-21 --stack-yaml stack-lts-21.yaml

Warning: nix (Nix) 2.29.0 is on the PATH (at /nix/var/nix/profiles/default/bin/nix) but Stack's Nix integration is disabled. To mute this message in future, set notify-if-nix-on-path: false in Stack's
         configuration.

Selected snapshot: lts-21.25.
[...]
Populating package index cache ...
Package index cache populated

Error: [S-242]
       Unable to parse Cabal file from package $PWD/wai/warp/warp.cabal:
       * 0:0: Unsupported cabal-version 3.12. See https://github.com/haskell/cabal/issues/4899.

       The Cabal file uses the Cabal specification version 3.12, but we only support up to version 3.8. Recommended action: upgrade your build tool (e.g. stack upgrade).
```

If in principle this work might be merged at a later date, we are happy to maintain a fork in a mergeable state. What do you say, @Vlix?

lrworth:
https://github.com/haskell/cabal/pull/9481 guards against using `PackageInfo_*` with `cabal-version: >= 1.10`, so that change would work with old versions of Cabal but not the ones that actually support `PackageInfo_*`. Building https://github.com/bellroy/wai/commit/5b6eb2cc536ae503ab813226c6ed46f313d16d70 with Cabal 3.14.2 gives me:

```
> [autogen-guard] To use the autogenerated module PackageInfo_* you need to specify `cabal-version: 3.12` or higher.
```

lrworth:
https://github.com/yesodweb/wai/pull/1044 is a much more conservative approach — optionally removing dependency on `Paths_*` by disabling runtime version support altogether.

PR #1044, opened 2025-08-30
Version optional

lrworth:

After submitting your PR:

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -31,13 +31,11 @@ import Data.Streaming.ByteString.Builder (
     newByteStringBuilderRecv,
     reuseBufferStrategy,
  )
-import Data.Version (showVersion)
 import Data.Word8 (_cr, _lf, _space, _tab)
 import qualified Network.HTTP.Types as H
 import qualified Network.HTTP.Types.Header as H
 import Network.Wai
 import Network.Wai.Internal
-import qualified Paths_warp
 import qualified System.TimeManager as T
 
 import Network.Wai.Handler.Warp.Buffer (toBuilderBuffer)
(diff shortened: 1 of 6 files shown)


Vlix (approved):
I think this is acceptable. I can also imagine some people would not want to broadcast the `warp` version for security reasons, so this would kill two birds with one stone 👍 .

If all the CI runs correctly and you add the `default` to the flag, this can be merged :)

warp/warp.cabal
```diff
@@ -34,6 +34,14 @@ flag network-bytestring
 flag allow-sendfilefd
     description: Allow use of sendfileFd (not available on GNU/kFreeBSD)
 
+flag include-warp-version
+    description:
+        Exposes the version in Network.Wai.Handler.Warp.warpVersion.
+        This can cause a large runtime dependency closure due to depending on
+        files shipped with GHC itself.
```
JackKelly-Bellroy, line 41: I think this is a Nix-specific concern, and should be flagged as such lest we spook non-Nix-using people.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -7,6 +7,7 @@
 module Network.Wai.Handler.Warp.Response (
     sendResponse,
     sanitizeHeaderValue, -- for testing
+    --  Provided here for backwards compatibility.
     warpVersion,
```
JackKelly-Bellroy, line 11: Question for the maintainers: should we deprecate this and remove it in a future version?

```haskell
{-# DEPRECATED warpVersion "some message" #-}
warpVersion = Network.Wai.Handler.Warp.Settings.warpVersion
```

kazu-yamamoto: I don't have a strong opinion.

Vlix:
Published as [warp-3.4.9](https://hackage.haskell.org/package/warp-3.4.9)

PR #1046, opened 2025-10-31
using GHC.Event.TimerManager

kazu-yamamoto:
This is a PoC. Please don't merge this at this moment.

@khibino This PR replaces `wai`'s `TimeManager` with `GHC.Event.TimerManager`.
Some functionalities are lost but I guess it's OK.

### DIFF
--- a/time-manager/System/TimeManager.hs
+++ b/time-manager/System/TimeManager.hs
@@ -86,7 +86,7 @@ data State
 
 ----------------------------------------------------------------
 
--- | Creating timeout manager which works every N micro seconds
+-- | Creating timeout manager which works every N microseconds
 --   where N is the first argument.
 initialize :: Int -> IO Manager
 initialize timeout


Vlix:
We're testing GHC 9.2 and above right now. And I feel that's a broad enough spectrum to support.
When GHC 9.12 turns into an LTS on stackage, I'd say we should drop 9.2.

The API seems to be stable for GHC <9.0 even, so I don't expect that to be an issue. Though I would like to keep both the GHC-specific time manager and the original one we have right now. Keeping the API the same, but maybe in separate modules that we make work using cabal conditionals? 🤔 (And otherwise behind a build flag?)

kazu-yamamoto:
The `old-manager` flag is provided.
@Vlix Please review this again.

Vlix:
I'll try to review this tomorrow. And see why all the CI is failing.

time-manager/System/TimeManager.hs
```diff
@@ -84,7 +86,7 @@ data State
 
 ----------------------------------------------------------------
 
--- | Creating timeout manager which works every N microseconds
+-- | Creating timeout manager which works every N micro seconds
```
Vlix: I think [`microsecond`](https://en.wikipedia.org/wiki/Microsecond) is one word in English. (I'm not an English native speaker either, btw, so 🤷 )

time-manager/System/TimeManager.hs
```diff
@@ -266,3 +268,227 @@ ignoreSync :: E.SomeException -> IO ()
...
+            mtid <- deRefWeak wtid
+            case mtid of
+                Nothing -> return ()
+                -- FIXME: forkIO to prevent blocking TimerManger
+                Just tid' -> E.throwTo tid' TimeoutThread
+
+----------------------------------------------------------------
+
+-- | Call the inner function with a timeout manager.
+withManager
```
Vlix: The `withManager` and `withManager'` functions are also redundant now, no?
If the `stopManager` and `killManager` functions don't do anything, then why bother wrapping in a bracket?

These can also be removed, potentially

warp/warp.cabal
```diff
@@ -114,7 +114,7 @@ library
         stm >=2.3,
         streaming-commons >=0.1.10,
         text,
-        time-manager >=0.2 && <0.3,
+        time-manager >=0.2 && <0.4,
```
Vlix: If we change the API, then we'll need to adjust `warp` as such as well.

If we support both `time-manager 0.2 && 0.3`, which is something we should do, then we'll need to adjust `withII` in `Network.Wai.Handler.Warp.Run`. But not much more, I expect.

Vlix (changes requested):
It looks ok, and it does simplify the module.

If we're going to change the API (which I think is the right decision), we should probably ditch my earlier plan of keeping the old functionality. Having two APIs is not worth it, and I expect we won't really maintain the old code, so we'll keep `v0.2` as the old time-manager, and `v0.3` as the new one.

If we remove the `USE_OLD_MANAGER` flag, we can also drop the `auto-update` dependency.

Vlix:
Also, it seems `http2` and `http3` both depend on `time-manager` as well:
```
       In the dependencies for http2-5.3.10:
         * time-manager must match >=0.2 && <0.3, but time-manager-0.3.0 is in the Stack
           configuration (latest matching version is 0.2.3).
       The above is/are needed due to warp-3.4.9 -> http2-5.3.10
       
       In the dependencies for http3-0.1.1:
         * time-manager must match >=0.2.3 && <0.3, but time-manager-0.3.0 is in the Stack
           configuration (latest matching version is 0.2.3).
       The above is/are needed due to warp-quic-0.0.4 -> http3-0.1.1
```
That's why the builds are failing.

time-manager/System/TimeManager.hs
kazu-yamamoto: Thanks.

kazu-yamamoto:
I have allowed time-manager v0.3 for `http2` and `http3` in Hackage but still failing, sigh.

warp/warp.cabal
kazu-yamamoto: In warp, most `pause ` and `resume` are used correctly as a pair.

Unfortunately,  `pause` is used solely `timeoutBody`.
We should understand cd82d85b060b720e2cd89b035b2a101b943c8273.

time-manager/System/TimeManager.hs
```diff
@@ -266,3 +268,227 @@ ignoreSync :: E.SomeException -> IO ()
...
+-- | Creating timeout manager which works every N microseconds
+--   where N is the first argument.
+initialize :: Int -> IO Manager
+initialize timeout
+    | timeout <= 0 = error "initialize"
+initialize timeout = return $ Manager timeout
+
+----------------------------------------------------------------
+
+-- | Obsoleted.
```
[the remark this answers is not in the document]
kazu-yamamoto: `DEPRECATE` is better because these functions are actually used in the field.

PR #1048, opened 2025-11-19
using GHC.Event.TimerManager

kazu-yamamoto:
@Vlix This should take over #1046

- Change APIs
- Only new code is provided
- `Manager 0` as `defaultManager`
- No `error`s
- Unnecessary `bracket`s are removed

Todo:

- [x] Documentation
- [x] Changelog

Note that `time-manager` v0.2.4 has been released but this version is not relating to this discussion.

### DIFF
--- a/time-manager/System/TimeManager.hs
+++ b/time-manager/System/TimeManager.hs
@@ -94,7 +94,10 @@ isEmptyHandle Handle{..} = handleTimeout == 0
 
 ----------------------------------------------------------------
 
--- | Creating timeout manager with a timeout value.
+-- | Creating timeout manager with a timeout value in microseconds.
+--
+--   Setting the timeout to zero or lower (<= 0) will produce a
+--   `defaultManager`.
 initialize :: Int -> IO Manager
 initialize = pure . Manager . max 0
 


time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
-            Inactive -> do
-                onTimeout <- I.readIORef handleActionRef
-                onTimeout `E.catch` ignoreSync
-                return Nothing
-            _ -> return $ Just m
-
-    inactivate Active = Inactive
-    inactivate x = x
+    | timeout <= 0 = return defaultManager
+    | otherwise = return $ Manager timeout
```
Vlix: `initialize = pure . Manager . max 0`

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
 
 ----------------------------------------------------------------
 
--- | Stopping timeout manager with onTimeout fired.
+-- | Obsoleted.
```
Vlix:
```hs
-- | Obsoleted since version 0.3.0
--
-- Is now equivalent to @pure ()@.
```

time-manager/System/TimeManager.hs
```diff
@@ -210,62 +208,32 @@ registerKillThread m onTimeout = do
...
-withManager timeout f =
-    E.bracket
-        (initialize timeout)
-        stopManager
-        f
+withManager timeout f = initialize timeout >>= f
 
 -- | Call the inner function with a timeout manager.
---   'killManager' is used after that.
+--   This is identical to 'withManager'.
```
Vlix: `--   This is identical to 'withManager' since version 0.3.0`

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
+    mgr <- getTimerManager
+    key <- I.readIORef handleKeyRef
+#if defined(mingw32_HOST_OS)
+    EV.updateTimeout mgr key $ fromIntegral (handleTimeout `div` 1000000)
+#else
+    EV.updateTimeout mgr key handleTimeout
+#endif
+
+-- | This is identical to 'cancel'.
+--   To resume timeout with the same 'Handle', 'resume' MUST be called.
```
Vlix: I feel like this line should be next to `tickle`?

```hs
-- | Extending the timeout. 'tickle' does __NOT__ resume a timeout. 
--
--   To resume a timeout, 'resume' must be called on a canceled 'Handle'.
tickle :: Handle -> IO ()
...

-- | This is identical to 'cancel'.
pause :: Handle -> IO ()
pause = cancel
```

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
+import qualified GHC.Event as EV
+#endif
+
 ----------------------------------------------------------------
 
 -- | A timeout manager
-data Manager = Manager (Reaper [Handle] Handle) | NoManager
+data Manager = Manager Int
 
--- | No manager.
```
Vlix:
```hs
-- | 'Manager' that does not timeout any actions.
--
-- To create a 'Manager' that does, use 'initialize'.
defaultManager :: Manager
```

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
-    = Active -- Manager turns it to Inactive.
-    | Inactive -- Manager removes it with timeout action.
-    | Paused -- Manager does not change it.
+isEmptyHandle :: Handle -> Bool
+isEmptyHandle Handle{..} = handleTimeout == 0
 
 ----------------------------------------------------------------
 
 -- | Creating timeout manager which works every N microseconds
 --   where N is the first argument.
```
Vlix:
```hs
-- | Creating timeout manager which works every N microseconds 
--   where N is the first argument.
--
--   Setting a timeout of 0 or less (or using 'defaultManager')
--   will act as if no timeout was set, so __NO__ 'TimeoutAction'
--   is run when using 'withHandle', and 'register' returns a
--   'Handle' that doesn't do anything.
```

Vlix: Wait... does the timeout manager run the `TimeoutAction` _EVERY_ N microseconds? Or only once?
I'd imagine it would only do it once, because that would be the timeout.

Or would something like the following be more accurate:
```
Create a timeout 'Manager' that takes the first argument as
microseconds. This is how long the timeout will take on first
invocation and which will be used to refresh the timer after
using 'tickle'.
```

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
+    EV.unregisterTimeout mgr key
+
+-- | Extending the timeout.
+tickle :: Handle -> IO ()
+tickle h | isEmptyHandle h = return ()
+tickle Handle{..} = do
+    mgr <- getTimerManager
+    key <- I.readIORef handleKeyRef
+#if defined(mingw32_HOST_OS)
+    EV.updateTimeout mgr key $ fromIntegral (handleTimeout `div` 1000000)
```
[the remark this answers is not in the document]
kazu-yamamoto: Yes. I believe so.

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
 killManager :: Manager -> IO ()
-killManager NoManager = return ()
-killManager (Manager mgr) = reaperKill mgr
+killManager _ = return ()
 
 ----------------------------------------------------------------
 
 -- | Registering a timeout action and unregister its handle
 --   when the body action is finished.
 --   'Nothing' is returned on timeout.
```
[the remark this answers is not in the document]
kazu-yamamoto: This code really sucks. I will fix it. Thank you for pointing out.

time-manager/System/TimeManager.hs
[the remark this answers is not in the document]
Vlix: Wow, I checked the source code... and it's true. That seems so weird to me. Someone didn't pay attention when making that API, I guess? 😅

time-manager/System/ThreadManager.hs
```diff
@@ -209,7 +210,7 @@ labelMe l = do
     labelThread tid l
 
 withHandle
-    :: ThreadManager -> T.TimeoutAction -> (T.Handle -> IO a) -> IO (Maybe a)
```
Vlix: Maybe also add this change (and the one in `System.TimeManager`) to the `ChangeLog.md`? It technically is a breaking change (which is fine since we're bumping a major version)

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +39,158 @@ module System.TimeManager (
...
-        Just <$> action th
+withHandle mgr onTimeout action
+    | isNoManager mgr = Just <$> action emptyHandle
+    | otherwise = do
+        E.handle ignore $ E.bracket (register mgr onTimeout) cancel $ \th ->
+            Just <$> action th
   where
     ignore TimeoutThread = return Nothing
 
 -- | Registering a timeout action of killing this thread and
```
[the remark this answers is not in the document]
Vlix: Also, I now understand the difference between `withHandle` and `withHandleKillThread` better, so this I understand this is not a synonym for `withHandle`.

Though I think a better name could be `withHandleKillAction`?
If you agree, we can rename and add
```hs
withHandleKillThread = withHandleKillAction
```
for backwards compatibility?

time-manager/System/ThreadManager.hs
kazu-yamamoto: > Yes, but the throwing of TimeoutThread is expected and caught with handle, so it doesn't actually kill the current thread, it kills the action and continues with the current thread. Right?

There are only tow threads: TimerManager (a) and the thread to call `withHandleKillThread` (b).
(a) throws `TimeoutThread` to (b).
(b)'s `ignore` catches it.
At this moment, (b)'s `action` has been terminated already.
`ignore` does not re-throw the exception and finished.
So, (b) dies.

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +43,156 @@ module System.TimeManager (
...
-withHandle :: Manager -> TimeoutAction -> (Handle -> IO a) -> IO (Maybe a)
-withHandle mgr onTimeout action =
-    E.handle ignore $ E.bracket (register mgr onTimeout) cancel $ \th ->
-        Just <$> action th
-  where
-    ignore TimeoutThread = return Nothing
+withHandle :: Manager -> TimeoutAction -> (Handle -> IO a) -> IO a
+withHandle mgr onTimeout action
+    | isNoManager mgr = action emptyHandle
+    | otherwise = E.bracket (register mgr onTimeout) cancel action
```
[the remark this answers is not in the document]
kazu-yamamoto: If we delete `>> throwTo tid CancelTimeout`, your impression gets right.

kazu-yamamoto:
> Though I think a better name could be `withHandleKillAction`?

I guess you did not understand the details described above.
I would like to hear your opinion again when you understand the details.

time-manager/System/TimeManager.hs
```diff
@@ -210,62 +210,33 @@ registerKillThread m onTimeout = do
...
-isAsyncException e =
-    case E.fromException (E.toException e) of
-        Just (E.SomeAsyncException _) -> True
-        Nothing -> False
-
-ignoreSync :: E.SomeException -> IO ()
-ignoreSync se
-    | isAsyncException se = E.throwIO se
-    | otherwise = return ()
+withManager' timeout f = initialize timeout >>= f
```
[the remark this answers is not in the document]
kazu-yamamoto: Done. Thanks!

time-manager/ChangeLog.md
```diff
@@ -1,5 +1,19 @@
...
+
+* New architecture. The backend is switched from the designated thread
+  to GHC's System TimerManager. From this version, this library is
+  just wrapper APIs of GHC's System TimerManager. Unlike v0.2 or
+  earlier, callbacks are executed at the exact time. System
+  TimerManager uses a PSQ (a tree) while v0.2 or earlier uses a list.
+  So, this version hopefully scales better.
+* Deprecated functions: `stopManager`, `killManager` and `withManager'`.
+* `tickle` sets the specified timeout from now.
+* `pause` is now idential to `cancel`.
```
Vlix: identical

time-manager/System/TimeManager.hs
```diff
@@ -210,62 +208,32 @@ registerKillThread m onTimeout = do
             mtid <- deRefWeak wtid
             case mtid of
                 Nothing -> return ()
+                -- FIXME: forkIO to prevent blocking TimerManger
```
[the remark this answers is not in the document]
Vlix: @kazu-yamamoto

time-manager/System/TimeManager.hs
```diff
@@ -37,162 +43,156 @@ module System.TimeManager (
...
-    | Inactive -- Manager removes it with timeout action.
-    | Paused -- Manager does not change it.
+isEmptyHandle :: Handle -> Bool
+isEmptyHandle Handle{..} = handleTimeout == 0
 
 ----------------------------------------------------------------
 
--- | Creating timeout manager which works every N microseconds
---   where N is the first argument.
+-- | Creating timeout manager with a timeout value.
```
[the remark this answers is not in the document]
kazu-yamamoto: Done.

kazu-yamamoto:
> Also, I now understand the difference between `withHandle` and `withHandleKillThread` better, so this I understand this is not a synonym for `withHandle`.

Sorry but I don't understand the difference in nuance.

Vlix (approved):
The remaining comments are mostly questions about documentation/naming, so I'll approve it now.

Issue #651, opened 2017-10-30
Connection Counter


Daniel-Diaz:
Hello.

In an application I wanted to have available the number of open connections to the server. So I used `setOnOpen` and `setOnClose` together with a variable. This worked well.

For some reason, I started reading warp's code, and realized that there is already a counter set up in those same places, but this counter is hidden from the user. I know keeping an additional counter is not a big deal, but maybe I just don't like redundant work.

In my opinion, the best way to solve this is to export the `Counter` type, but only with construction and accessing functions so that the user can't mess up the internals, and then add `setWithCounter :: Maybe Counter -> Settings -> Settings` so that the user of the library can create a counter and keep track of it. When `Nothing` is provided, warp would create the counter just like it does today.

Well, this was my thought today. If it's regarded as a good idea, I can go ahead and write a patch.

Thanks.

snoyberg:
I'd defer to @kazu-yamamoto on this one. My gut feeling is that this is an internal structure that we wouldn't really want to expose, but I don't have a strong feeling here.

kazu-yamamoto:
Generally, I don't mind exporting `Counter`.

> so that the user of the library can create a counter and keep track of it.

I don't understand this. So, please show me your patch.

Vlix:
@Daniel-Diaz Could you explain what you meant with the following?

> so that the user of the library can create a counter and keep track of it.

Do you mean you want the user to be able to provide a TVar which will be adjusted exactly the same as the internal one? I don't think we want to expose the internal one, since it is relied on for graceful shutdowns.

Daniel-Diaz:
> @Daniel-Diaz Could you explain what you meant with the following?
> 
> > so that the user of the library can create a counter and keep track of it.
> 
> Do you mean you want the user to be able to provide a TVar which will be adjusted exactly the same as the internal one? I don't think we want to expose the internal one, since it is relied on for graceful shutdowns.

Hello!

Wow, this is really old. I totally forgot about it!

I'm not completely sure what I meant by that sentence, but I definitely meant for the counter to be read-only. I guess a function to read it without exposing the type would also do.

domenkozar:
This is going to be fixed after 8 years in #1050

PR #1050, opened 2025-12-06
warp: expose open connection count via getOpenConnectionCount

domenkozar:
Adds ability to monitor the number of currently open connections by storing the connection counter in Settings and exposing it through a new getOpenConnectionCount function.

Fixes https://github.com/yesodweb/wai/issues/651

After submitting your PR:

### DIFF
--- a/warp/warp.cabal
+++ b/warp/warp.cabal
@@ -1,6 +1,6 @@
 cabal-version:      >=1.10
 name:               warp
-version:            3.4.10
+version:            3.4.11
 license:            MIT
 license-file:       LICENSE
 maintainer:         michael@snoyman.com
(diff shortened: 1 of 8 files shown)


Vlix (changes requested):
First of all, thank you for the PR 🙏 contributions are always appreciated!

I explain in the comment why I'm requesting changes. I'd prefer the settings to not use `unsafePerformIO` and `NOINLINE`, so that there's no magic shenanigans.
Adding a function to get the `Counter` and exposing the `getCount` function would make more sense to me.

I'd love to hear if there's anything I'm missing or overlooking, though.

warp/warp.cabal
```diff
@@ -290,6 +290,7 @@ benchmark parser
     hs-source-dirs:   bench .
     other-modules:
         Network.Wai.Handler.Warp.Conduit
+        Network.Wai.Handler.Warp.Counter
```
Vlix: Why does the benchmark need this module suddenly? 🤔

Vlix (changes requested):
Great! All tests pass and this looks pretty much ready.

Found some language "issues" that I'd like to see addressed before we publish this, but we can consider that the cleanup, I guess :)

Vlix (approved):
Thank you for your contribution 🎉 

We'll merge this when we can publish it immediately after.

kazu-yamamoto (approved):
LGTM

PR #1056, opened 2026-01-18
Time manager tests

Vlix:

---

Added tests to make sure the `System.TimeManager` works as expected.

We can publish and tag `time-manager-0.3.1` after this, I think. AND deprecate `time-manager-0.3.0` on hackage.


kazu-yamamoto (approved):
This PR is awesome!
I re-run the failed test but failed again.
But it's not an objection to merge this PR.

PR #1064, opened 2026-03-11
Handle graceful shutdown for keep-alive connections

### LINKED ISSUE #853: Graceful shutdown not working with keep-alive
I encountered unexpected behaviour with the graceful shutdown method. I have the following code:

```haskell

import Control.Monad

import Data.Time

import Network.Wai (responseLBS, Application)
import Network.HTTP.Types.Header (hContentType)

import System.Signal

logM :: String -> IO ()
logM s = do
  t <- getCurrentTime
  putStrLn $ show t ++ ": " ++ s

main = do
    let port = 3000
    logM $ "Listening on port " ++ show port
    let shutdownHandler closeSocket =
          installHandler sigINT $ const (logM "Closing webserver" >> closeSocket >> logM "Socket closed")
    let settings =
          setPort port $
          setInstallShutdownHandler shutdownHandler $
          defaultSettings
    runSettings settings app
    logM "Server exited"

app :: Application
app req f =
    f $ responseLBS status200 [(hContentType, "text/plain")] "Hello world!"
```

Now when I run this and send a sigint, the program will shut down properly. However, when I request `localhost:3000` in my browser, and then ask the server to shutdown, it will always take up the full grace period:

```
2021-07-13 07:11:06.6979399 UTC: Listening on port 3000
^C2021-07-13 07:11:16.0393217 UTC: Closing webserver
2021-07-13 07:11:16.0398661 UTC: Socket closed
2021-07-13 07:11:21.0430659 UTC: Server exited
```

However, this is unexpected, as there were no request coming in at this moment at all, so the server could have shut down immediately. As this does not happen when I make requests in other ways, I assume it is related to the fact that the browser sends a `Connection: keep-alive` header and keeps the connection alive. This alive connection probably stops the graceful shutdown from terminating, even though no requests are being made.

Is this by design, or can this be solved? With this behaviour I think the graceful shutdown has little value, while it could be a quite useful tool.

mkleczek:
Implementing graceful shutdown for keep-alive sessions (HTTP 1.x). I would like to discuss the possibility to merge it.

Fixes #853 

TODO:
Not sure if the following needs to be done (Internal API was changed).

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -427,9 +427,10 @@ attachConn mysa ctx = do
     writeBufferRef <- I.newIORef writeBuffer
     -- Creating a cache for leftover input data.
     tls <- getTLSinfo ctx
-    return (conn writeBufferRef isH2, tls)
+    connActiveApps <- newCounter
+    return (conn writeBufferRef isH2 connActiveApps, tls)
   where
-    conn writeBufferRef isH2 =
+    conn writeBufferRef isH2 connActiveApps' =
         Connection
             { connSendMany = TLS.sendData ctx . L.fromChunks
             , connSendAll = sendall
@@ -440,6 +441,7 @@ attachConn mysa ctx = do
             , connWriteBuffer = writeBufferRef
             , connHTTP2 = isH2
             , connMySockAddr = mysa
+            , connActiveApps = connActiveApps'
             }
       where
         sendall = TLS.sendData ctx . L.fromChunks . return
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -100,6 +100,7 @@ import Network.Wai.Handler.Warp.Internal
 import Network.Wai.Handler.WarpTLS.Internal
 import System.IO.Error (ioeGetErrorType, isEOFError)
 import System.Timeout (timeout)
+import Control.Concurrent.STM (TVar, readTVarIO, readTVar)
 
 ----------------------------------------------------------------
 
@@ -330,11 +331,12 @@ getter
     -> Settings
     -> Socket
     -> params
+    -> TVar Bool
     -> IO (IO (Connection, Transport), SockAddr)
-getter tlsset set@Settings{settingsAccept = accept'} sock params = do
+getter tlsset set@Settings{settingsAccept = accept'} sock params shuttingDown = do
     (s, sa) <- accept' sock
     setSocketCloseOnExec s
-    return (mkConn tlsset set s params, sa)
+    return (mkConn tlsset set s params shuttingDown, sa)
 
 mkConn
     :: TLS.TLSParams params
@@ -342,8 +344,9 @@ mkConn
     -> Settings
     -> Socket
     -> params
+    -> TVar Bool
     -> IO (Connection, Transport)
-mkConn tlsset set s params = do
+mkConn tlsset set s params shuttingDown = do
     let tm = settingsTimeout set * 1000000
     mbs <- timeout tm recvFirstBS
     case mbs of
@@ -353,8 +356,8 @@ mkConn tlsset set s params = do
     recvFirstBS = safeRecv s 4096 `onException` close s
     switch firstBS
         | S.null firstBS = close s >> throwIO ClientClosedConnectionPrematurely
-        | S.head firstBS == 0x16 = httpOverTls tlsset set s firstBS params
-        | otherwise = plainHTTP tlsset set s firstBS
+        | S.head firstBS == 0x16 = httpOverTls tlsset set s firstBS params shuttingDown
+        | otherwise = plainHTTP tlsset set s firstBS shuttingDown
 
 ----------------------------------------------------------------
 
@@ -376,13 +379,15 @@ httpOverTls
     -> Socket
     -> S.ByteString
     -> params
+    -> TVar Bool
     -> IO (Connection, Transport)
-httpOverTls TLSSettings{..} set s bs0 params =
+httpOverTls TLSSettings{..} set s bs0 params shuttingDown = do
     makeConn `onException` close s
   where
     makeConn = do
+        activeApps <- newCounter
         pool <- newBufferPool 2048 16384
-        rawRecvN <- makeRecvN bs0 $ receive s pool
+        rawRecvN <- makeRecvN bs0 $ makeRecv s (readTVar shuttingDown) activeApps pool
         let recvN = wrappedRecvN rawRecvN
         ctx <- TLS.contextNew (backend recvN) params
         TLS.contextHookSetLogging ctx tlsLogging
@@ -390,7 +395,7 @@ httpOverTls TLSSettings{..} set s bs0 params =
         mconn <- timeout tm $ do
             TLS.handshake ctx
             mysa <- getSocketName s
-            attachConn mysa ctx
+            attachConn mysa ctx shuttingDown activeApps
         case mconn of
           Nothing -> throwIO IncompleteHeaders
           Just conn -> return conn
@@ -419,16 +424,15 @@ httpOverTls TLSSettings{..} set s bs0 params =
 
 -- | Get "Connection" and "Transport" for a TLS connection that is already did the handshake.
 -- @since 3.4.7
-attachConn :: SockAddr -> TLS.Context -> IO (Connection, Transport)
-attachConn mysa ctx = do
+attachConn :: SockAddr -> TLS.Context -> TVar Bool -> Counter -> IO (Connection, Transport)
+attachConn mysa ctx shuttingDown activeApps = do
     h2 <- (== Just "h2") <$> TLS.getNegotiatedProtocol ctx
     isH2 <- I.newIORef h2
     writeBuffer <- createWriteBuffer 16384
     writeBufferRef <- I.newIORef writeBuffer
     -- Creating a cache for leftover input data.
     tls <- getTLSinfo ctx
-    connActiveApps <- newCounter
-    return (conn writeBufferRef isH2 connActiveApps, tls)
+    return (conn writeBufferRef isH2 activeApps, tls)
   where
     conn writeBufferRef isH2 connActiveApps' =
         Connection
@@ -442,6 +446,7 @@ attachConn mysa ctx = do
             , connHTTP2 = isH2
             , connMySockAddr = mysa
             , connActiveApps = connActiveApps'
+            , connShuttingDown = readTVarIO shuttingDown
             }
       where
         sendall = TLS.sendData ctx . L.fromChunks . return
@@ -509,10 +514,10 @@ tryIO = try
 ----------------------------------------------------------------
 
 plainHTTP
-    :: TLSSettings -> Settings -> Socket -> S.ByteString -> IO (Connection, Transport)
-plainHTTP TLSSettings{..} set s bs0 = case onInsecure of
+    :: TLSSettings -> Settings -> Socket -> S.ByteString -> TVar Bool -> IO (Connection, Transport)
+plainHTTP TLSSettings{..} set s bs0 shuttingDown = case onInsecure of
     AllowInsecure -> do
-        conn' <- socketConnection set s
+        conn' <- socketConnection set s shuttingDown
         cachedRef <- I.newIORef bs0
         let conn'' =
                 conn'
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -1,5 +1,10 @@
 # ChangeLog for warp
 
+## 3.4.12.1
+
+* Graceful shutdown waited for all connections to close, resulting in potentially indefinite waiting on keep-alive connections. It now proactively closes connections after waiting for current request processing to be completed.
+  [#1064] (https://github.com/yesodweb/wai/pull/1064)
+
 ## 3.4.12
 
 * Respond with `Connection: close` header if connection is to be closed after a request.
--- a/warp/Network/Wai/Handler/Warp/HTTP1.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP1.hs
@@ -120,6 +120,7 @@ http1server settings ii conn transport app addr th istatus src =
         -- See comment below referencing
         -- https://github.com/yesodweb/wai/issues/618
         | Just NoKeepAliveRequest <- fromException e = return ()
+        | Just ShutdownInProgress <- fromException e = return ()
         -- No valid request
         | Just (BadFirstLine _) <- fromException e = return ()
         | isAsyncException e = throwIO e
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -120,7 +120,9 @@ sendResponse
     -> IO Bool
     -- ^ Returing True if the connection is persistent.
 sendResponse settings conn ii th req reqidxhdr src response = do
-    hs <- addConnection . addAltSvc settings <$> addServerAndDate hs0
+    -- handle graceful shutdown
+    shuttingDown <- connShuttingDown conn
+    hs <- addConnection shuttingDown . addAltSvc settings <$> addServerAndDate hs0
     if hasBody s
         then do
             -- The response to HEAD does not have body.
@@ -134,12 +136,12 @@ sendResponse settings conn ii th req reqidxhdr src response = do
                 Nothing -> return ()
                 Just realStatus -> logger req realStatus mlen
             T.tickle th
-            return ret
+            return $ ret && not shuttingDown
         else do
             _ <- sendRsp conn ii th ver s hs rspidxhdr maxRspBufSize method RspNoBody
             logger req s Nothing
             T.tickle th
-            return isPersist
+            return $ isPersist && not shuttingDown
   where
     defServer = settingsServerName settings
     logger = settingsLogger settings
@@ -148,7 +150,7 @@ sendResponse settings conn ii th req reqidxhdr src response = do
     s = responseStatus response
     hs0 = sanitizeHeaders $ responseHeaders response
     rspidxhdr = indexResponseHeader hs0
-    addConnection hs = if (hasBody s && not ret) || (not (hasBody s) && not isPersist)
+    addConnection shuttingDown hs = if shuttingDown || (hasBody s && not ret) || (not (hasBody s) && not isPersist)
                        then (H.hConnection, "close") : hs
                        else hs
     getdate = getDate ii
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -64,6 +64,7 @@ socketConnection _ s = do
     writeBufferRef <- newIORef writeBuffer
     isH2 <- newIORef False -- HTTP/1.x
     mysa <- getSocketName s
+    connActiveApps' <- newCounter
     return
         Connection
             { connSendMany = Sock.sendMany s
@@ -87,6 +88,7 @@ socketConnection _ s = do
             , connWriteBuffer = writeBufferRef
             , connHTTP2 = isH2
             , connMySockAddr = mysa
+            , connActiveApps = connActiveApps'
             }
   where
     receive' sock pool = E.handle handler $ receive sock pool
@@ -366,10 +368,16 @@ fork set mkConn addr app counter ii = settingsFork set $ \unmask -> do
             $ \goingon ->
                 -- Actually serve this connection.  bracket with closeConn
                 -- above ensures the connection is closed.
-                when goingon $ serveConnection conn ii th addr transport set app
+                when goingon $ serveConnection conn ii th addr transport set (updateConnActiveApps conn)
 
     onOpen adr = increase counter >> settingsOnOpen set adr
     onClose adr _ = decrease counter >> settingsOnClose set adr
+
+    updateConnActiveApps conn req resp =
+        E.bracket_
+            (increase $ connActiveApps conn)
+            (decrease $ connActiveApps conn)
+            (app req resp)
 
 serveConnection
     :: Connection
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -27,6 +27,7 @@ import Network.Socket (
     getSocketName,
     setSocketOption,
     withSocketsDo,
+    waitReadSocketSTM,
  )
 #if MIN_VERSION_network(3,1,1)
 import Network.Socket (gracefulClose)
@@ -51,13 +52,30 @@ import Network.Wai.Handler.Warp.Imports hiding (readInt)
 import Network.Wai.Handler.Warp.SendFile
 import Network.Wai.Handler.Warp.Settings
 import Network.Wai.Handler.Warp.Types
+import Control.Concurrent.STM (atomically, writeTVar, check, readTVar, readTVarIO, newTVarIO, TVar, STM)
+import Data.Functor (($>))
+import Control.Exception (throwIO)
+
+makeRecv :: Socket -> STM Bool -> Counter -> BufferPool -> Recv
+makeRecv sock shuttingDown activeApps pool = do
+    sockWait <- waitReadSocketSTM sock
+    join . atomically $ do
+        isShuttingDown <- shuttingDown
+        appsInProgress <- getCountSTM activeApps
+        -- when shutting down and no apps in progress throw
+        check (isShuttingDown && appsInProgress < 1) $> throwIO ShutdownInProgress
+        <|>
+        -- else wait for socket readiness and do non-blocking read
+        sockWait $> recv
+    where
+        recv = receive sock pool
 
 -- | Creating 'Connection' for plain HTTP based on a given socket.
-socketConnection :: Settings -> Socket -> IO Connection
+socketConnection :: Settings -> Socket -> TVar Bool -> IO Connection
 #if MIN_VERSION_network(3,1,1)
-socketConnection set s = do
+socketConnection set s shuttingDown = do
 #else
-socketConnection _ s = do
+socketConnection _ s shuttingDown = do
 #endif
     bufferPool <- newBufferPool 2048 16384
     writeBuffer <- createWriteBuffer 16384
@@ -83,15 +101,16 @@ socketConnection _ s = do
 #else
             , connClose = close s
 #endif
-            , connRecv = receive' s bufferPool
+            , connRecv = receive' s bufferPool connActiveApps'
             , connRecvBuf = \_ _ -> return True -- obsoleted
             , connWriteBuffer = writeBufferRef
             , connHTTP2 = isH2
             , connMySockAddr = mysa
             , connActiveApps = connActiveApps'
+            , connShuttingDown = readTVarIO shuttingDown
             }
   where
-    receive' sock pool = E.handle handler $ receive sock pool
+    receive' sock pool activeApps = E.handle handler $ makeRecv sock (readTVar shuttingDown) activeApps pool
       where
         handler :: E.IOException -> IO ByteString
         handler e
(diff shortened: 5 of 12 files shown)


kazu-yamamoto:
I tested your patch but it does not solve #853.
I don't understand why your logic can solve #853.

Moreover, if you don't want to use graceful shutdown, why not setting `settingsGracefulShutdownTimeout` to 0?

mkleczek:
> I tested your patch but it does not solve #853. I don't understand why your logic can solve #853.
> 

@kazu-yamamoto thanks for taking a look.

#853 really is about two (related but separate) issues:

1. `warp` not closing idle _keep-alive_ connections
2. As described in https://github.com/yesodweb/wai/issues/853#issuecomment-963409019 - `warp` happily accepting new requests after shutdown signal.

Indeed - my patch initially handled only case 2 (ie. after processing the request the connection was closed) but failed to address case 1 (idle _keep-alive_ connections).

I updated the patch so that shutdown signals are handled while waiting for next request.

> Moreover, if you don't want to use graceful shutdown, why not setting `settingsGracefulShutdownTimeout` to 0?

The issue is that in reality graceful shutdown does not work for _keep-alive_ connections: if you set a limit then it is not going to be graceful (because `warp` accepts new requests). If you don't set a limit then `warp` will not shutdown.

kazu-yamamoto:
> As described in https://github.com/yesodweb/wai/issues/853#issuecomment-963409019 - warp happily accepting new requests after shutdown signal.

I was not able to confirm this. With the test code in #853, attempts to create new connections are rejected after `kill -INT`.

mkleczek:
> > As described in [#853 (comment)](https://github.com/yesodweb/wai/issues/853#issuecomment-963409019) - warp happily accepting new requests after shutdown signal.
> 
> I was not able to confirm this. With the test code in #853, attempts to create new connections are rejected after `kill -INT`.

New connections are rejected but requests on existing open connections are accepted.

We have a reproducer in PostgREST: https://github.com/mkleczek/postgrest/blob/babc1206febe6acbc8a3636f8bbb6b88b6112d64/test/io/test_io.py#L151

This test starts a client that issues requests in a loop re-using HTTP 1.1 _keep-alive_ connections. This keeps the terminated PostgREST alive and accepting requests.

With a fix from this PR `warp` finishes handling in-flight requests, closes all connections and quits.


Looking at the `warp` code in [http1server function](https://github.com/yesodweb/wai/blob/e2a6756dfeb5bf910d88976fe1f6adf9a0228078/warp/Network/Wai/Handler/Warp/HTTP1.hs#L138): there is no code here that would stop request handling loop when the listening socket is closed.

kazu-yamamoto:
@mkleczek OK. I understand what your problem is. I will review this PR again.

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -135,9 +136,12 @@ http1server settings ii conn transport app addr th istatus src =
...
             throwIO e
 
-    loop firstRequest = do
-        (req, mremainingRef, idxhdr, nextBodyFlush) <-
-            recvRequest firstRequest settings conn ii th addr src transport
+    loop firstRequest =
+        -- graceful shutdown:
+        -- exit loop before reading the request on connFinishWait signal
+        -- but once we received the request let it be processed
+        race (connFinishWait conn) (recvRequest firstRequest settings conn ii th addr src transport) >>=
```
kazu-yamamoto: `race` is a mechanism that relies on asynchronous exceptions, which introduces new problems. Consider whether race can be achieved by representing  `connFinishWait` and the return values ​​of `recvRequest` with `STM` and composing them.

I call it *event polling* style.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -120,6 +121,30 @@ sendResponse
     -> IO Bool
     -- ^ Returing True if the connection is persistent.
 sendResponse settings conn ii th req reqidxhdr src response = do
+    -- handle graceful shutdown
+    shouldKeepAlive <- connKeepAlive conn
+    let
+        (isPersist, isChunked0) = DB.first (shouldKeepAlive &&) $ infoFromRequest req reqidxhdr
```
kazu-yamamoto: The combinator style might look cool to the writer, but it's painful for the reader. Please write in a more straightforward style.

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -135,9 +136,12 @@ http1server settings ii conn transport app addr th istatus src =
...
+        foldMap (\(req, mremainingRef, idxhdr, nextBodyFlush) -> do
```
kazu-yamamoto: I don't understand this `foldMap` intuitively. Please write it in a more straightforward way.

kazu-yamamoto:
CI on `master` is green but this PR is red.
We need to determine whether this pull request is the cause of the CI failure.

warp/Network/Wai/Handler/Warp/HTTP1.hs
mkleczek: > `race` is a mechanism that relies on asynchronous exceptions, which introduces new problems. Consider whether race can be achieved by representing the return values ​​of `connFinishWait` and `recvRequest` with `STM` and composing them.
> 
> I call it _event polling_ style.

Hmm... `recvRequest` is a blocking IO action (at the bottom it uses `Recv` from `recv` library which is blocking). I don't know of any other way to interrupt it (except asynchronous exceptions used by `race`).

We would need to change `warp` to use non-blocking IO but that would be a major rewrite, I guess.

kazu-yamamoto: Prepare a `STM` variable, then write it and return.

kazu-yamamoto: Also, `waitReadSocketSTM` from `network` to prevent blocking.
I will consider that we can apply this style later.

mkleczek: > Also, `waitReadSocketSTM` from `network` to prevent blocking. I will consider that we can apply this style later.

Thanks, I learned something new!
Should be better now.

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: I find it easier to read myself but happy to adjust.
Fixed.

warp/Network/Wai/Handler/Warp/HTTP1.hs
mkleczek: > I don't understand this `foldMap` intuitively. Please write it in a more straightforward way.

`foldMap` == `either (const $ pure ())` here. Anyway: fixed now - there is no `race` usage anymore and previous code was restored.

mkleczek:
> CI on `master` is green but this PR is red. We need to determine whether this pull request is the cause of the CI failure.

@kazu-yamamoto Thanks for your patience, that was trickier than I thought.
`cabal test warp:spec` is green for me locally now. Also added GracefulShutdown test case to the test suite. I am afraid I cannot start GitHub CI.
I still need to dig into one failing TimeManager test (`TimeManager resume also works as tickle with pauses`).

mkleczek:
> > CI on `master` is green but this PR is red. We need to determine whether this pull request is the cause of the CI failure.
> 
> @kazu-yamamoto Thanks for your patience, that was trickier than I thought. `cabal test warp:spec` is green for me locally now. Also added GracefulShutdown test case to the test suite. I am afraid I cannot start GitHub CI. I still need to dig into one failing TimeManager test (`TimeManager resume also works as tickle with pauses`).

Still `warp-tls` to fix. I am not sure how it works though and how much space for reuse there is.

kazu-yamamoto:
I took some time today to review this PR.

* `sendResponse` doesn't need changing. Checking for `connKeepAlive` in `http1server`'s `loop` is sufficient.
* `recvRequest` needs to make the header processing part non-blocking. This will require significant modifications.

I'll take over the rest of the work.

mkleczek:
> I took some time today to review this PR.
> 
> * `sendResponse` doesn't need changing. Checking for `connKeepAlive` in `http1server`'s `loop` is sufficient.

Hmm... `sendResponse` sets `Connection: close` header - I guess we want to set it during shutdown?

> * `recvRequest` needs to make the header processing part non-blocking. This will require significant modifications.

That's indeed something I am not sure as I don't know how buffering works. I assume wrapping `Recv` in non-blocking `waitSocketReadSTM` is not enough?

> 
> I'll take over the rest of the work.

OK, cool, thx. ( hope to be mentioned in changelog though :) ).
Right now the whole `stack test` passes.

kazu-yamamoto:
> Hmm... `sendResponse` sets `Connection: close` header - I guess we want to set it during shutdown?

I overlooked that. However, I think the changes will be very minor.

Vlix:
Please add me as a reviewer when the changes look final. @kazu-yamamoto :pray:

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -170,17 +204,17 @@ runSettings set app =
...
-    closeListenSocket = close socket
+    shutdownState <- newShutdownState
+    let getConn = do
+            (s, sa) <- accept' socket
+            setSocketCloseOnExec s
+            -- NoDelay causes an error for AF_UNIX.
+            setSocketOption s NoDelay 1 `E.catch` throughAsync (return ())
+            conn <- socketConnection set s shutdownState
+            return (conn, sa)
+    runSettingsConnection set shutdownState getConn (ensureGracefulShutdown shutdownState app)
```
mkleczek: Left original code here but I think the above 3 lines might be simplified:
```
            return (socketConnection set s shutdownState, sa)
    runSettingsConnectionMaker set shutdownState getConn (ensureGracefulShutdown shutdownState app)
```

And the function `runSettingsConnection` can be removed altogether.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -120,6 +120,32 @@ sendResponse
...
+        -- Make sure we don't hang on to 'response' (avoid space leak)
+        !ret = case response of
+            ResponseFile{} -> isPersist
+            ResponseBuilder{} -> isKeepAlive
+            ResponseStream{} -> isKeepAlive
+            ResponseRaw{} -> False
+        addConnection hs =
+            if (hasBody s && not ret) || (not (hasBody s) && not isPersist)
+            then (H.hConnection, "close") : hs
+            else hs
```
Vlix: This whole section can be replaced with:

```hs
-- handle graceful shutdown
shouldKeepAlive <- connKeepAlive conn
let keepAliveWithBody =
      shouldKeepAlive && if hasBody s then ret else isPersist
    addConnection hs =
      if not keepAliveWithBody
        then (H.hConnection, "close") : hs
        else hs
```

And remove the `addConnection` definition in the `where`.
Then also in the final part of this function, factor out the `tickle` and `return` from the if block:

```hs
if hasBody s
  then do
    ...
    (ms, mlen) <- sendRsp ...
    case ms of { ... }
  else do
    _ <- sendRsp ...
    logger ...
T.tickle th
return keepAliveWithBody
```

and we'll have made the code a little more robust along the way.

warp/warp.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      >=1.10
 name:               warp
-version:            3.4.12
+version:            3.4.12.1
```
Vlix: I feel like a PATCH bump is a not enough for a change in shutdown behaviour.
Wouldn't a MINOR bump be more appropriate?

It's not like it didn't work before, it just shutdown on the connection level, not the request level.
Would it be a good idea to change the "graceful shutdown" setting so that users can decide whether they want the current behaviour or the "per request" behaviour?

warp/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # ChangeLog for warp
 
+## 3.4.12.1
+
+* Fix: [#853] Graceful shutdown not working with keep-alive.
```
Vlix: I wouldn't say it did not work, it just didn't work as some people expected it.

We should describe the difference in behaviour so users know what changed.

Vlix (commented):
Since I've already started the review before the new PRs landed, I'll just comment them now. I'm not requesting these changes at the moment, we'll see what the other #1065 and #1066 PRs are like first.

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: I agree with you but I didn't want to perform non-related refactoring here and leave as much original code as possible.

warp/ChangeLog.md
mkleczek: > I wouldn't say it did not work, it just didn't work as some people expected it.
>
> We should describe the difference in behaviour so users know what changed.

We have a reproducer that shows `warp` will not shutdown at all if there are new requests coming. I would say it falls under common sense definition of _shutdown not working_ .

But let's not into philosophical discussion - I am happy to treat it as change (not a fix). Not sure what should be put into the changelog though - existing sentence is exact title of #853 .

warp/Network/Wai/Handler/Warp/Response.hs
Vlix: I'd say my suggestion would leave way more original code, because the diff now is a large chunk and checking that it's moved code instead of changed code is a lot harder.

warp/ChangeLog.md
Vlix:
```markdown
Graceful shutdown waited for all connections to close, resulting in potentially 
indefinite waiting on keep-alive connections. It now proactively closes connections
after waiting for current request processing to be completed.
  [#1064] (https://github.com/yesodweb/wai/pull/1064)
```

warp/warp.cabal
Vlix: What do you think, @kazu-yamamoto

Vlix (changes requested):
As I've said in the other PRs, make the "isShuttingDown" `TVar Bool` a field in the `Settings`, so that it's available everywhere, because it's a "server-wide" flag.
(Make it `Maybe (TVar Bool)`, with the `defaultSettings` initializing `Nothing`, and then the `runSettingsConnectionMakerSecure` creating the TVar and inserting it in the `Settings`)

It will cut down on the diffs a lot and make it a lot more obvious what's going on.
You can also take the `ShutdownState`'s "isInProgress" TVar and just put that in the `Connection`, so it's just a flag for "is the `Application` handling a request". That way it could also be used for other things, without having the shutdown connotation. (though do mention the shutdown behaviour its used for in the comments)

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: > I'd say my suggestion would leave way more original code, because the diff now is a large chunk and checking that it's moved code instead of changed code is a lot harder.

Unfortunately, I had to move the code that now depends on dynamic `shouldKeepAlive`  to the `let` binding. Didn't want to introduce any other change.
But I am happy to simplify the code if you find it better. Let me analyze your proposal and apply it.

warp/ChangeLog.md
mkleczek: > ```
> Graceful shutdown waited for all connections to close, resulting in potentially 
> indefinite waiting on keep-alive connections. It now proactively closes connections
> after waiting for current request processing to be completed.
>   [#1064] (https://github.com/yesodweb/wai/pull/1064)
> ```

No mention of #853 ? So maybe it should be closed as "won't fix"?

mkleczek:
> As I've said in the other PRs, make the "isShuttingDown" `TVar Bool` a field in the `Settings`, so that it's available everywhere, because it's a "server-wide" flag. (Make it `Maybe (TVar Bool)`, with the `defaultSettings` initializing `Nothing`, and then the `runSettingsConnectionMakerSecure` creating the TVar and inserting it in the `Settings`)
> 
> It will cut down on the diffs a lot and make it a lot more obvious what's going on.

I actually tried that first and, while it might help avoid changing function signatures, `Settings` is a public interface and IMHO it would be a really bad idea to expose private internal state there. Exposing **mutable** internal state (ie. a `TVar`) there would be even worse.
We could add encapsulated `TVar` (ie. `ShutdownManager`) as a field to settings but that would widen public API surface and only confuse users suggesting it is somehow configurable.

I also tried to add it to `InternalInfo` but it doesn't work well with `WarpTLS` as `Internalnfo` is not shared.

I am afraid there is no way to do it without wider signature changes because the logic needs to be shared with `WarpTLS`.

> You can also take the `ShutdownState`'s "isInProgress" TVar and just put that in the `Connection`, so it's just a flag for "is the `Application` handling a request". That way it could also be used for other things, without having the shutdown connotation. (though do mention the shutdown behaviour its used for in the comments)

I also tried that but, mind you, `ShutdownState` is not a single variable but **two** variables: global `shuttingDown` and per-connection `isInProgress`. **Both** are needed as `Connection` members.
What makes things more complex is that `WarpTLS` has its own lifecycle and its own way of initializing connections. The way it is done now enables reusing the same logic for both `Warp` and `WarpTLS`.

(actually encapsulating `isInProgress` inside `ShutdownState` payed off really quickly because I had to change it from `IORef Bool` to `TVar Int` in latest version to account for the fact that in `HTTP2` there may be multiple requests in progress on a single connection).

Please, review the latest version, I tried to minimize the code churn to make review easier.

Vlix:
I've been thinking and if I'm not mistaken:

If we disconnect the "isShuttingDown" TVar and the "isInProgress" part, the "isShuttingDown" only has to be created in the `acceptConnection` helper function:

> EDIT: Oh, right, `sendResponse` also needs the `ShutdownManager`... I guess we could keep the `ShutdownManager` in the `ShutdownState` as well, initializing it with a`ShutdownManager` in `newShutdownState` just to satisfy the type checker, but then override it in `adjustConnection`, so that the `ShutdownManager` inside the `Connection` is the one used by `gracefulShutdown`. And after that just look in the `Connection` in `sendResponse`?

```hs
-- We redefine what the ShutdownState is

data ShutdownState =
    ShutdownState
        (STM ())   -- ^ sockWait
        (TVar Int) -- ^ appsInProgress

newShutdownState :: Socket -> IO ShutdownState
newShutdownState sock =
    ShutdownState
        <$> waitReadSocketSTM sock
        <*> newTVarIO 0

---------------------------------------------------------------

-- Wrapping the 'recv' just needs both things to check

wrapRecv :: ShutdownState -> ShutdownManager -> IO a -> IO a
wrapRecv (ShutdownState sockWait appsInProgress) (ShutdownManager shuttingDown) recv =
    join . atomically $
        -- !!! This should be '> 0' NOT '>= 0' !!!
        (check . (> 0) =<< readTVar appsInProgress) $> recv
        <|>
        (check =<< readTVar shuttingDown) $> throwIO ShutdownInProgress
        <|>
        sockWait $> recv

---------------------------------------------------------------

-- We don't touch the 'recv' yet, we adjust it in 'runSettingsConnectionMakerSecure'

socketConnection _ s = do
    ...
    shutdownState <- newShutdownState s
    ...
        , connShutdownState = shutdownState
        }

-- We create the ShutdownManager here, and wrap the 'recv' in 'adjustConn'

acceptConnection set getConnMaker app counter ii = do
    shutdownManager <- newShutdownManager
    void $ E.mask_ (acceptLoop shutdownManager)
    gracefulShutdown set shutdownManager counter
  where
    acceptLoop shutdownManager = do
        E.allowInterrupt
        mx <- acceptNewConnection
        case mx of
            Nothing -> return ()
            Just (mkConn, addr) -> do
                let newMkConn = adjustConn shutdownManager mkConn
                fork set newMkConn addr app counter ii
                acceptLoop
    adjustConn :: ShutdownManager -> IO (Connection, Transport) -> IO (Connection, Transport)
    adjustConn mgr mkConn = do
        (conn, transport) <- mkConn
        let sdState = connShutdownState conn
            recv = connRecv conn
            newConn = conn {connRecv = wrapRecv sdState mgr recv}
        pure (newConn, transport)            
```

warp/ChangeLog.md
Vlix: The ChangeLog always references the PR that made the change, the issue giving rise to that PR will be found in said PR.

Your initial post of this PR literally has this line:

  - [x] Update the Changelog.md file with a link to your **PR**

warp/warp.cabal
kazu-yamamoto: Version should be 3.4.13 at least.

Setting is a good idea if possible.

warp/Network/Wai/Handler/Warp/Response.hs
kazu-yamamoto: @mkleczek Why don't you just copy and modify `Response.hs` from #1066?

kazu-yamamoto:
I cannot build the current code:

```
[24 of 34] Compiling Network.Wai.Handler.Warp.Response ( Network/Wai/Handler/Warp/Response.hs, dist/build/Network/Wai/Handler/Warp/Response.o, dist/build/Network/Wai/Handler/Warp/Response.dyn_o )
Network/Wai/Handler/Warp/Response.hs:171:81: error: [GHC-88464]
    Variable not in scope: isPersist :: Bool
    |
171 |     addConnection ret hs = if (hasBody s && not ret) || (not (hasBody s) && not isPersist)
    |                                                                                 ^^^^^^^^
```

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -51,19 +52,55 @@ import Network.Wai.Handler.Warp.Imports hiding (readInt)
...
+newShutdownManger = ShutdownManager <$> newTVarIO False
+
+-- TODO naming
+handleShutdown :: Socket -> ShutdownManager -> IO a -> IO (IO a, ShutdownState)
+handleShutdown sock mgr@(ShutdownManager shuttingDown) recv = do
+    -- we have a counter of requests in progress
+    -- because HTTP2 multiplexing allows simiultaneous
+    -- requests on a single TCP connection
+    appsInProgress <- newTVarIO 0
+    sockWait <- waitReadSocketSTM sock
```
kazu-yamamoto, line 61: `waitReadSocketSTM` needs to be called every time before the receiving function is called. In this code, it appears to be called only once per connection.

kazu-yamamoto:
@mkleczek The counter in `appsInProgress` is a very good idea. We can change our graceful shutdown from connection-base to application-base. This solves the graceful shutdown problem in the case where all connections are inactive.

@Vlix What about accepting `appsInProgress` as the first step? If you agree, I would like to ask @mkleczek to make a minimum patch to introduce `appsInProgress` only. This should make use of the functions in `Counter.hs`.

After merging this patch, let's consider the case where some connections are still active.

kazu-yamamoto:
I'm sorry but I would be inactive due to business trip from tomorrow. I should be active from 30th March.

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: An example code:

```haskell
func sock otherSTM = do
    wait <- waitReadSocketSTM sock
    x <- atomically $ do
        wait or otherSTM
    case x of
      Ready -> recv sock
      Timeout -> throwIO TimedOut
```

Don't *reuse* `wait`.

Vlix: Ah, that's good to know. I didn't understand what `waitReadSocketSTM` actually does, apparently.

Vlix:
> @Vlix What about accepting `appsInProgress` as the first step? If you agree, I would like to ask @mkleczek to make a minimum patch to introduce `appsInProgress` only. This should make use of the functions in `Counter.hs`.

Having a counter in the connection with the amount of active application handlers is definitely useful. That in itself would be a good addition. `connActiveApps :: Counter`?

warp/Network/Wai/Handler/Warp/Run.hs
mkleczek: > `waitReadSocketSTM` needs to be called every time before the receiving function is called. In this code, it appears to be called only once per connection.

Thanks, didn't know that. Fixed.

mkleczek:
> I've been thinking and if I'm not mistaken:
> 
> If we disconnect the "isShuttingDown" TVar and the "isInProgress" part, the "isShuttingDown" only has to be created in the `acceptConnection` helper function:
> 
> > EDIT: Oh, right, `sendResponse` also needs the `ShutdownManager`... I guess we could keep the `ShutdownManager` in the `ShutdownState` as well, initializing it with a`ShutdownManager` in `newShutdownState` just to satisfy the type checker, but then override it in `adjustConnection`, so that the `ShutdownManager` inside the `Connection` is the one used by `gracefulShutdown`. And after that just look in the `Connection` in `sendResponse`?
> 
> ```haskell
> -- We redefine what the ShutdownState is
> 
> data ShutdownState =
>     ShutdownState
>         (STM ())   -- ^ sockWait
>         (TVar Int) -- ^ appsInProgress
[snip]

It looked like a great idea but we also need to pass the global `shuttingDown` `TVar` to `sendResponse` function in `Response` module. Either you change signature here (and handle cascading changes in dependent code) or you pass `ShutdownState` pointing to `ShutdownManager` in `Connection`.

Or you have and idea how to resolve it?

mkleczek:
>         -- !!! This should be '> 0' NOT '>= 0' !!!
>         (check . (> 0) =<< readTVar appsInProgress) $> recv

Argh... Off by one... Thanks, fixed!

warp/ChangeLog.md
mkleczek: Fixed

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -440,6 +445,8 @@ attachConn mysa ctx = do
             , connWriteBuffer = writeBufferRef
             , connHTTP2 = isH2
             , connMySockAddr = mysa
+            , connActiveApps = connActiveApps'
```
Vlix: Can just be `activeApps`.

mkleczek:
@kazu-yamamoto @Vlix - I've restructured the PR:
* extracted adding `connActiveApps` counter to a separate commit
* reworked code to minimize code churn - especially in `Response.hs`

I also made sure that blocking is avoided in `WarpTLS` by wrapping `recv` below TLS wrapper (so TLS code always calls non-blocking `recv`)

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -51,19 +52,38 @@ import Network.Wai.Handler.Warp.Imports hiding (readInt)
...
 import Network.Wai.Handler.Warp.Types
+import Control.Concurrent.STM (atomically, writeTVar, check, readTVar, readTVarIO, newTVarIO, TVar, STM)
+import Data.Functor (($>))
+import Control.Exception (throwIO)
+
+makeRecv :: Socket -> STM Bool -> Counter -> BufferPool -> Recv
+makeRecv sock shuttingDown activeApps pool = do
+    sockWait <- waitReadSocketSTM sock
+    join . atomically $
+        -- when any Application is in progress perform blocking read
```
kazu-yamamoto: This blocking `recv` is not acceptable because we would compose a timeout event here in the future. In other words, `recv` MUST be used as non-blocking.

mkleczek: Hmm... Not sure I follow: graceful shutdown timeout is already handled in https://github.com/mkleczek/wai/blob/57d7ab33d8a8bf326396bd4f865ae9280f5a3b28/warp/Network/Wai/Handler/Warp/Run.hs#L467 - isn't it?

Is your comment related to graceful shutdowns or you have something else (not yet implemented) in mind?

kazu-yamamoto: I'm not talking about that timeout.
What I'm talking about is timeout by the timeout manager to kill the worker (application) thread.
In event-polling programming, asynchronous exception should not be used.
This means we don't use `throwTo` nor `killThread`.
The timeout manger should turn an STM variable to tell the timeout to the worker.

mkleczek: > The timeout manger should turn an STM variable to tell the timeout to the worker.

Right, but this is not implemented yet in `warp`? (ie. this PR does not change anything in this regard).

Anyway - I've updated the code to never perform blocking read: https://github.com/mkleczek/wai/blob/c6da9a5e6bfcf1fe7d77212047904ea8cb4cc252/warp/Network/Wai/Handler/Warp/Run.hs#L62

Vlix:

PR #1065, opened 2026-03-17
Shutting active connection

kazu-yamamoto:
@mkleczek This is rough sketch on what I'm thinking.
Many test cases fail at this moment.
I will try to resolve them tomorrow.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/HTTP1.hs
+++ b/warp/Network/Wai/Handler/Warp/HTTP1.hs
@@ -8,8 +8,16 @@ module Network.Wai.Handler.Warp.HTTP1 (
     http1,
 ) where
 
-import qualified Control.Concurrent as Conc (yield)
-import Control.Exception (SomeException, catch, fromException, throwIO, try)
+import qualified Control.Concurrent as Conc
+import Control.Concurrent.STM
+import Control.Exception (
+    SomeException,
+    bracket,
+    catch,
+    fromException,
+    throwIO,
+    try,
+ )
 import qualified Data.ByteString as BS
 import Data.Char (chr)
 import Data.IORef (IORef, newIORef, readIORef, writeIORef)
@@ -102,6 +110,13 @@ http1 settings ii conn transport app origAddr th bs0 = do
 
     decodeAscii = map (chr . fromEnum) . BS.unpack
 
+checkLoop :: TVar Bool -> TBQueue a -> STM Bool
+checkLoop shuttingDown reqQ = do
+    isShuttingDown <- readTVar shuttingDown
+    reqIsReady <- not <$> isEmptyTBQueue reqQ
+    check (isShuttingDown || reqIsReady)
+    return reqIsReady
+
 http1server
     :: Settings
     -> InternalInfo
@@ -113,8 +128,11 @@ http1server
     -> IORef Bool
     -> Source
     -> IO ()
-http1server settings ii conn transport app addr th istatus src =
-    loop FirstRequest `catch` handler
+http1server settings ii conn transport app addr th istatus src = do
+    reqQ <- newTBQueueIO 10 -- fixme: hard coding
+    let reader = recvRequest settings conn ii th addr src transport reqQ
+    bracket (Conc.forkIO reader) Conc.killThread $ \_ ->
+        loop reqQ `catch` handler
   where
     handler e
         -- See comment below referencing
@@ -135,39 +153,43 @@ http1server settings ii conn transport app addr th istatus src =
                     e
             throwIO e
 
-    loop firstRequest = do
-        (req, mremainingRef, idxhdr, nextBodyFlush) <-
-            recvRequest firstRequest settings conn ii th addr src transport
-        keepAlive <-
-            processRequest
-                settings
-                ii
-                conn
-                app
-                th
-                istatus
-                src
-                req
-                mremainingRef
-                idxhdr
-                nextBodyFlush
-                `catch` \e -> do
-                    settingsOnException settings (Just req) e
-                    -- Don't throw the error again to prevent calling settingsOnException twice.
-                    return CloseConnection
+    loop reqQ = do
+        ready <- atomically $ checkLoop (connShuttingDown conn) reqQ
+        when ready $ do
+            ex <- atomically $ readTBQueue reqQ
+            case ex of
+                Left e -> throwIO e
+                Right (req, mremainingRef, idxhdr, nextBodyFlush) -> do
+                    keepAlive <-
+                        processRequest
+                            settings
+                            ii
+                            conn
+                            app
+                            th
+                            istatus
+                            src
+                            req
+                            mremainingRef
+                            idxhdr
+                            nextBodyFlush
+                            `catch` \e -> do
+                                settingsOnException settings (Just req) e
+                                -- Don't throw the error again to prevent calling settingsOnException twice.
+                                return CloseConnection
 
-        -- When doing a keep-alive connection, the other side may just
-        -- close the connection. We don't want to treat that as an
-        -- exceptional situation, so we pass in SubsequentRequest to http1 (which
-        -- in turn passes in SubsequentRequest to recvRequest), indicating that
-        -- this is not the first request. If, when trying to read the
-        -- request headers, no data is available, recvRequest will
-        -- throw a NoKeepAliveRequest exception, which we catch here
-        -- and ignore. See: https://github.com/yesodweb/wai/issues/618
+                    -- When doing a keep-alive connection, the other side may just
+                    -- close the connection. We don't want to treat that as an
+                    -- exceptional situation, so we pass in SubsequentRequest to http1 (which
+                    -- in turn passes in SubsequentRequest to recvRequest), indicating that
+                    -- this is not the first request. If, when trying to read the
+                    -- request headers, no data is available, recvRequest will
+                    -- throw a NoKeepAliveRequest exception, which we catch here
+                    -- and ignore. See: https://github.com/yesodweb/wai/issues/618
 
-        case keepAlive of
-            ReuseConnection -> loop SubsequentRequest
-            CloseConnection -> return ()
+                    case keepAlive of
+                        ReuseConnection -> loop reqQ
+                        CloseConnection -> return ()
 
 data ReuseConnection = ReuseConnection | CloseConnection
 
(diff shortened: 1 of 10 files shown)


mkleczek:
> @mkleczek This is rough sketch on what I'm thinking. Many test cases fail at this moment. I will try to resolve them tomorrow.

OK, so you decoupled request reading from processing by introducing a separate socket reading thread and a queue.
You implemented a workaround for blocking socket reading by introducing a worker thread to handle it.

In my patch I tried to avoid it as it:
* introduces additional latency of queuing and dequeuing
* only handles HTTP1 case

My approach is to make `connRecv` perform both `shutdown` flag and socket readiness check and return `null` on shutdown.
It is fine because according to the spec we don't have to process partially read requests.
I also check for the request being in progress to avoid partial reads of request body by application.

See: https://github.com/mkleczek/wai/blob/703723ea93a84dfcfc771f167dd7e63d4341c3ae/warp/Network/Wai/Handler/Warp/Run.hs#L68 for `connRecv` wrapper used by both TLS and non-TLS server. I think (and correct me if I'm wrong) it is non-blocking as it will not read until `waitSocketReadSTM` signals there is something to read.

That way we should handle both HTTP1 and HTTP2 (since we are handling shutdowns in a layer below).

It is your call which approach is better, thanks for promptly looking at the issue and spending your time on it!


EDIT: it surely would be better if we could have `connRecv` be `IO (Maybe ByteString)` instead but that would require major refactoring of the codebase to handle partiallity. I guess returning empty `ByteString` works as well, doesn't it?

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -113,8 +128,11 @@ http1server
     -> IORef Bool
     -> Source
     -> IO ()
-http1server settings ii conn transport app addr th istatus src =
-    loop FirstRequest `catch` handler
+http1server settings ii conn transport app addr th istatus src = do
+    reqQ <- newTBQueueIO 10 -- fixme: hard coding
+    let reader = recvRequest settings conn ii th addr src transport reqQ
```
mkleczek: I think there is a problem with `reader` not checking shutdown flag.

`checkLoop` returns `True` if there is something in the queue, regardless of shutdowns. So the request processing loop will never stop if new requests keep coming in even if shutdown flag is True.

kazu-yamamoto:
The tests now pass. However, I suspect the performance is poor due to context switching between threads.

kazu-yamamoto:
@Vlix GHC's lightweight threads are so well-designed that a programming style of casually throwing asynchronous exceptions between threads became popular in the Haskell community. The `async` library is a prime example.

From years of experience in network programming, I've learned that using asynchronous exceptions makes protocol termination difficult because they can be thrown without any context.

To solve this problem, I've recently started moving towards an event polling model proposed by my colleague @khibino. Events such as packet inputs and timeouts are each represented in an STM, composed and inspected at the beginning of the loop.

To achieve this, all components must be *non-blocking*.

Packet input is very difficult to handle because it often doesn't arrive all at once. If you can implement a parser from scratch, you can use a method to save the intermediate state of parsing. However, if you are using existing *blocking* code, you will need to separate the thread that handles the input and have it write the results to the STM.

warp/Network/Wai/Handler/Warp/HTTP1.hs
kazu-yamamoto: Right. The current code solves this, I believe.

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -102,6 +111,16 @@ http1 settings ii conn transport app origAddr th bs0 = do
 
     decodeAscii = map (chr . fromEnum) . BS.unpack
 
+checkLoop
+    :: H1Event
+    -> STM (Maybe (Either SomeException H1Request))
+checkLoop H1Event{..} = do
+    isShuttingDown <- readTVar h1evShuttingDown
+    mreq <- readTVar h1evRequest
+    check (isShuttingDown || isJust mreq)
```
Vlix, line 120: `check (not isShuttingDown || isJust mreq)` ?

Also, `checkLoop` will _never_ return a `Nothing`, right? Because of the `isJust mreq`? Should it, at some point? Like the following?

```hs
if isShuttingDown
  then pure Nothing
  else do
    check (isJust mreq) 
    writeTVar ...
    pure mreq
```

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -135,39 +156,45 @@ http1server settings ii conn transport app addr th istatus src =
...
+                        -- throw a NoKeepAliveRequest exception, which we catch here
+                        -- and ignore. See: https://github.com/yesodweb/wai/issues/618
 
-        case keepAlive of
-            ReuseConnection -> loop SubsequentRequest
-            CloseConnection -> return ()
+                        Conc.putMVar (h1evSync $ connH1Event conn) ()
+                        case keepAlive of
+                            ReuseConnection -> loop
+                            CloseConnection -> return ()
```
Vlix: Shouldn't the `putMVar` go here?

```hs
case keepAlive of
  ReuseConnection -> do
    Conc.putMVar (h1evSync $ connH1Event conn) ()
    loop
  CloseConnection -> return ()
```

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -120,7 +121,8 @@ sendResponse
     -> IO Bool
     -- ^ Returing True if the connection is persistent.
 sendResponse settings conn ii th req reqidxhdr src response = do
-    hs <- addConnection . addAltSvc settings <$> addServerAndDate hs0
+    isShuttingDown <- atomically $ readTVar $ h1evShuttingDown $ connH1Event conn
```
Vlix: `atomically . readTVar` === `readTVarIO`

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -217,13 +220,13 @@ runSettingsConnectionMaker x y =
 --
 -- Since 2.1.4
 runSettingsConnectionMakerSecure
-    :: Settings -> IO (IO (Connection, Transport), SockAddr) -> Application -> IO ()
-runSettingsConnectionMakerSecure set getConnMaker app = do
+    :: Settings -> IO (IO (Connection, Transport), SockAddr) -> Application -> H1Event -> IO ()
+runSettingsConnectionMakerSecure set getConnMaker app h1ev = do
```
Vlix, line 224: There is no guarantee that this `h1ev` will be the same one that will be in the `IO (IO (Connection, Transport), SockAddr)`

These functions are also exposed in the public API through `Network.Wai.Handler.Warp.Internal`, so if possible, we should try to not require an external `H1Event`.

Come to think of it... isn't the `isShuttingDown` IORef more of a server-wide variable? I think it would make sense to put it in the `Settings`. `http1` and `http2` also get the settings, so I don't think it has to be in the `Connection`.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -169,15 +171,16 @@ runSettings set app =
 -- 'serverPort' record.
 runSettingsSocket :: Settings -> Socket -> Application -> IO ()
 runSettingsSocket set@Settings{settingsAccept = accept'} socket app = do
+    h1ev <- newH1Event
     settingsInstallShutdownHandler set closeListenSocket
-    runSettingsConnection set getConn app
+    runSettingsConnection set (getConn h1ev) app h1ev
```
Vlix, line 176: If we put the `isShuttingDown` IORef in the `Settings`, then the `newH1Event` can be moved to inside `socketConnection` and this avoids a lot of noise.

This will basically be a non-exposed field of `Settings`, so that a `defaultSettings` has a `Nothing :: IORef Bool` and it gets filled with a `Just ref` after using `newH1Event` at the start of `runSettingsConnectionMakerSecure`.

warp/test/GracefulShutdownSpec.hs
```diff
@@ -0,0 +1,84 @@
...
+                signalQSemN receivedRequests 1
+                -- block until signaled
+                readMVar allowResponse
+                respond $ responseLBS status200 [("Content-Length", "0")] ""
+            
+            client sendRequest = do
+                -- first request should return OK
+                sendRequest >>=
+                    (responseStatus <&> (`shouldBe` ok200)) <>
+                    (lookup "Connection" . responseHeaders <&> (`shouldBe` Just "close"))
```
Vlix:
```hs
client sendRequest = do
  -- first request should return OK
  res <- sendRequest
  responseStatus res `shouldBe` ok200
  lookup hConnection (responseHeaders res) `shouldBe` Just "close"
```

Vlix (changes requested):
I understand what the goal is, and at least for HTTP1, this does seem like an ok change. Though, if we'd move the `isShuttingDown` IORef to the `Settings`, would the `H1Event` even still be needed?

If we put the `isShuttingDown` check in the `http1server.loop` instead of the `checkLoop`, then it will also always do an `isShuttingDown` check before asking `recvRequest`, which it also does this way.
The sync `MVar` is basically making sure it is not running concurrently, so why not just do it in the loop we already have?

warp/Network/Wai/Handler/Warp/Run.hs
Vlix: After reviewing mkleczek's PR, I also don't think adding it to the `Settings` is a good idea. (though it could work)

I think instead of passing the `h1ev` around just to get the "isShuttingDown" var into `acceptConnection`, just make a new one in `acceptConnection` and wrap the `getConnMaker` to use the "isShuttingDown" var created in `acceptConnection`?

warp/Network/Wai/Handler/Warp/HTTP1.hs
kazu-yamamoto: You are right.

warp/Network/Wai/Handler/Warp/Response.hs
kazu-yamamoto: Thanks.

warp/Network/Wai/Handler/Warp/HTTP1.hs
kazu-yamamoto: Thanks.
This should be:

```haskell
checkLoop
    :: H1Event
    -> STM (Maybe (Either SomeException H1Request))
checkLoop H1Event{..} = do
    isShuttingDown <- readTVar h1evShuttingDown
    mreq <- readTVar h1evRequest
    check (isShuttingDown || isJust mreq)
    writeTVar h1evRequest Nothing
    if isShuttingDown
        then return Nothing
        else return mreq
```

warp/test/GracefulShutdownSpec.hs
kazu-yamamoto: Much better! Thanks.

Vlix:

PR #1066, opened 2026-03-18
Mkleczek shutdown

kazu-yamamoto:
@mkleczek #1064 has an unnecessarily large difference, making it difficult to review.
This is minimum-diff version of #1064.

As you've probably noticed, this code will block if not the entire request has been received. For example, during a slowloris attack, it won't shut down as intended, but is that something you've decided is unavoidable?

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Internal.hs
+++ b/warp/Network/Wai/Handler/Warp/Internal.hs
@@ -14,6 +14,13 @@ module Network.Wai.Handler.Warp.Internal (
     runSettingsConnection,
     runSettingsConnectionMaker,
     runSettingsConnectionMakerSecure,
+    -- TODO naming and order tidy-up
+    ShutdownManager,
+    ShutdownState,
+    newShutdownManger,
+    newShutdownState,
+    handleShutdown,
+    ensureGracefulShutdown,
     Transport (..),
 
     -- * Connection
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -11,7 +11,7 @@ module Network.Wai.Handler.Warp.Run where
 import Control.Arrow (first)
 import qualified Control.Exception as E
 import qualified Data.ByteString as S
-import Data.IORef (newIORef, readIORef)
+import Data.IORef (newIORef, readIORef, writeIORef)
 import Data.Streaming.Network (bindPortTCP)
 import Foreign.C.Error (Errno (..), eCONNABORTED, eMFILE)
 import GHC.Conc.Sync (labelThread, myThreadId)
@@ -27,6 +27,7 @@ import Network.Socket (
     getSocketName,
     setSocketOption,
     withSocketsDo,
+    waitReadSocketSTM,
  )
 #if MIN_VERSION_network(3,1,1)
 import Network.Socket (gracefulClose)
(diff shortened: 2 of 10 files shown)


kazu-yamamoto:
@Vlix We should compare #1065 and this PR.

mkleczek:
> @mkleczek #1064 has an unnecessarily large difference, making it difficult to review. This is minimum-diff version of #1064.
> 
> As you've probably noticed, this code will block if not the entire request has been received.

I believe it won't block before control is handed over to `Application` (ie. `isInProgress` is set to `True` in the below code) - see inline comments in the below code:
```hs
    isInProgress <- newIORef False
    sockWait <- waitReadSocketSTM sock
    -- recv' is a wrapper around `recv` which blocks when there is no data to read
    -- we "unblock" it here using `waitReadSocketSTM`
    let recv' = do
            inProgress <- readIORef isInProgress
            if inProgress then
                -- we can block here - this is by design
                -- because Application might have already started processing
                -- and we want to make sure that once started it does not receive partial data
                recv
            else
                 -- the below should be non-blocking
                join . atomically $
                    -- if shutting down return empty result - this means "EOF"
                    (check =<< readTVar shuttingDown) $> mempty
                    <|>
                    -- recv should not block here - socket has data to read
                    sockWait $> recv
    pure (recv', ShutdownState mgr isInProgress)-- not <$> readTVarIO shuttingDown)
```

> For example, during a slowloris attack, it won't shut down as intended, but is that something you've decided is unavoidable?

I believe the above code protects against that during reading of headers before handing over to `Application`, because `headerLines` throws when reading empty `ByteString` from `connRecv`: https://github.com/yesodweb/wai/blob/e2a6756dfeb5bf910d88976fe1f6adf9a0228078/warp/Network/Wai/Handler/Warp/Request.hs#L127-L135

Once `Application` takes over we still have slowloris protection that will kill the thread and hence let shutdown proceed.

mkleczek:
> @mkleczek #1064 has an unnecessarily large difference, making it difficult to review. This is minimum-diff version of #1064.
> 

@kazu-yamamoto @Vlix 

I am still pushing commits to #1064 branch in my repo. Rebasing my commits in your repository @kazu-yamamoto makes it impossible for me to contribute updates to this PR.
I can try to split #1064 into multiple commits so that you can review them one by one, if you prefer.
What do you think?

I've pushed a very important change to implementation of `handleShutdown` wrapper. Instead of returning empty result it now throws `ShutdownInProgress` exception that is handled in `http1server`. That makes the whole thing much cleaner and not dependent on `headerLines` throwing on empty read result. See: https://github.com/mkleczek/wai/blob/6f9135c43823de3fdcc050e0c507a46ea4d2559a/warp/Network/Wai/Handler/Warp/Run.hs#L79

To be honest, I think `handleShutdown` wrapper might also be used for slowloris protection to gracefully exit request processing loop instead of killing the thread.

warp/Network/Wai/Handler/Warp/Run.hs
Vlix, line 1: Most of the changes in this file are unnecessary if we put the `isShuttingDown` flag in the `Settings`.

warp/Network/Wai/Handler/Warp/Internal.hs
```diff
@@ -14,6 +14,13 @@ module Network.Wai.Handler.Warp.Internal (
     runSettingsConnection,
     runSettingsConnectionMaker,
     runSettingsConnectionMakerSecure,
+    -- TODO naming and order tidy-up
+    ShutdownManager,
+    ShutdownState,
+    newShutdownManger,
+    newShutdownState,
+    handleShutdown,
```
Vlix, line 22: I actually don't think we want to export any of these.

Vlix (changes requested):
Seems to do what we want.

The `isInProgress` IORef mainly ensures that when `warp` is done parsing (and receiving) the headers of the `Request`, the application will be able to receive the body if it has started handling the request.

The shutdown flag ensures `sendResponse` results in a `False`, so that connections get closed. (at least in HTTP1)

All in all a lot of changes that would be WAY less if the `isShuttingDown` IORef was in `Settings`.

kazu-yamamoto:
> I believe the above code protects against that during reading of headers before handing over to Application, because headerLines throws when reading empty ByteString from connRecv: 

I don't understand this claim.

`headerLines` calls `Source` multiple times, potentially blocking each time. This happens before `Application` is called.

kazu-yamamoto:
`headerLines` is a blocking function. The core of the problem is how to avoid this blocking.

kazu-yamamoto:
Never mind. I begin to understand your idea.

Vlix:
Yes, the idea of @mkleczek is to wrap the `connRecv` function into one that will just be (potentially blocking) `connRecv` when the given `Application` is handling the already parsed `Request`, and that in the part before the `Application`, it first uses `sockWait` (created from `waitReadSocketSTM`) before calling `connRecv`, so that it is in practice non-blocking. (IIUC)

mkleczek:
> Yes, the idea of @mkleczek is to wrap the `connRecv` function into one that will just be (potentially blocking) `connRecv` when the given `Application` is handling the already parsed `Request`, and that in the part before the `Application`, it first uses `sockWait` (created from `waitReadSocketSTM`) before calling `connRecv`, so that it is in practice non-blocking. (IIUC)

That's exactly the idea.

Vlix:

PR #1067, opened 2026-03-23
App based counter

kazu-yamamoto:
@Vlix This makes graceful shutdown immediate if all connections are inactive.
Cc: @mkleczek

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -265,7 +265,7 @@ acceptConnection
     -> InternalInfo
     -> IO ()
 acceptConnection set getConnMaker app ii = do
-    counter <- case settingsConnectionCounter set of
+    connCounter <- case settingsConnectionCounter set of
         Just c -> pure c
         Nothing -> newCounter
     -- First mask all exceptions in acceptLoop. This is necessary to
@@ -273,13 +273,13 @@ acceptConnection set getConnMaker app ii = do
     -- acceptNewConnection and the registering of connClose.
     --
     -- acceptLoop can be broken by closing the listening socket.
-    void $ E.mask_ $ acceptLoop counter
+    void $ E.mask_ $ acceptLoop connCounter
     -- In some cases, we want to stop Warp here without graceful shutdown.
     -- So, async exceptions are allowed here.
     -- That's why `finally` is not used.
-    gracefulShutdown set counter
+    gracefulShutdown set connCounter
   where
-    acceptLoop counter = do
+    acceptLoop connCounter = do
         -- Allow async exceptions before receiving the next connection maker.
         E.allowInterrupt
 
@@ -290,25 +290,25 @@ acceptConnection set getConnMaker app ii = do
         -- expensive work should not be performed in the main event
         -- loop. An example of something expensive would be TLS
         -- negotiation.
-        mx <- acceptNewConnection counter
+        mx <- acceptNewConnection connCounter
         case mx of
             Nothing -> return ()
             Just (mkConn, addr) -> do
-                fork set mkConn addr app counter ii
-                acceptLoop counter
+                fork set mkConn addr app connCounter ii
+                acceptLoop connCounter
 
-    acceptNewConnection counter = do
+    acceptNewConnection connCounter = do
         ex <- E.try getConnMaker
         case ex of
             Right x -> return $ Just x
             Left e -> do
                 let getErrno (Errno cInt) = cInt
                     isErrno err = ioe_errno e == Just (getErrno err)
-                if | isErrno eCONNABORTED -> acceptNewConnection counter
+                if | isErrno eCONNABORTED -> acceptNewConnection connCounter
                    | isErrno eMFILE -> do
                        settingsOnException set Nothing $ E.toException e
-                       waitForDecreased counter
-                       acceptNewConnection counter
+                       waitForDecreased connCounter
+                       acceptNewConnection connCounter
                    | otherwise -> do
                        settingsOnException set Nothing $ E.toException e
                        return Nothing
@@ -323,7 +323,7 @@ fork
     -> Counter
     -> InternalInfo
     -> IO ()
-fork set mkConn addr app counter ii = settingsFork set $ \unmask -> do
+fork set mkConn addr app connCounter ii = settingsFork set $ \unmask -> do
     tid <- myThreadId
     labelThread tid "Warp just forked"
     -- Call the user-supplied on exception code if any
@@ -367,8 +367,8 @@ fork set mkConn addr app counter ii = settingsFork set $ \unmask -> do
                 -- above ensures the connection is closed.
                 when goingon $ serveConnection conn ii th addr transport set app
 
-    onOpen adr = increase counter >> settingsOnOpen set adr
-    onClose adr _ = decrease counter >> settingsOnClose set adr
+    onOpen adr = increase connCounter >> settingsOnOpen set adr
+    onClose adr _ = decrease connCounter >> settingsOnClose set adr
 
 serveConnection
     :: Connection
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -220,10 +220,7 @@ runSettingsConnectionMakerSecure
     :: Settings -> IO (IO (Connection, Transport), SockAddr) -> Application -> IO ()
 runSettingsConnectionMakerSecure set getConnMaker app = do
     settingsBeforeMainLoop set
-    counter <- case settingsConnectionCounter set of
-        Just c -> pure c
-        Nothing -> newCounter
-    withII set $ acceptConnection set getConnMaker app counter
+    withII set $ acceptConnection set getConnMaker app
 
 -- | Running an action with internal info.
 --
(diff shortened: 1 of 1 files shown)


mkleczek:
> @Vlix This makes graceful shutdown immediate if all connections are inactive. Cc: @mkleczek

Looks like a very simple and elegant solution!

Do you think it is worth incorporating `GracefulShutdownTest` from #1064 ?

kazu-yamamoto:
@mkleczek The next step is here:

https://github.com/kazu-yamamoto/wai/commits/graceful-shutdown2/

mkleczek:
> @mkleczek The next step is here:
> 
> https://github.com/kazu-yamamoto/wai/commits/graceful-shutdown2/

Having looked at this, I am not so sure anymore if this PR is worth doing TBH.
It only covers a very specific case (ie. no traffic) - any app in progress will disable shutdown. And since `warp` accepts requests on existing connections there is a very low probability of zero active apps.

Vlix:
@kazu-yamamoto Looks like a good first step to get to a better graceful shutdown.

I like that it checks running `Application`s instead of connections, but I do wonder if this (this PR on its own) would cause issues just around the `Application`?

Because the `runSettingsConnectionMakerSecure` will just "finish" with `()`. So any still running forked actions will continue until the OS kills the Haskell runtime, right? (or at least not immediately)
AND if it does stop immediately, then the running `fork`s could be in the middle of parsing a request (which I think is ok?), but could they also be in the middle of sending a `Response`? Or does the `descrease appCounter` not run until the `Response` is completely sent?

> Do you think it is worth incorporating `GracefulShutdownTest` from #1064 ?

I think having this test is already a good idea, even if it might fail with just this PR. It is what we're working towards.

kazu-yamamoto:
> Having looked at this, I am not so sure anymore if this PR is worth doing TBH.
It only covers a very specific case (ie. no traffic) - any app in progress will disable shutdown. And since `warp` accepts requests on existing connections there is a very low probability of zero active apps.

@mkleczek I just showed that our problem can be divided into two parts. This first step is mandatory for the second step. If we don't merge this first step, shutdown cannot be done immediately even if the second step is solely merged.

kazu-yamamoto:
> Because the `runSettingsConnectionMakerSecure` will just "finish" with `()`. So any still running forked actions will continue until the OS kills the Haskell runtime, right? (or at least not immediately)

I don't think so. On shutting down,  `connRecv` returns "", and this eventually triggers to throw an exception, resulting kill the thread.

> AND if it does stop immediately, then the running `fork`s could be in the middle of parsing a request (which I think is ok?), but could they also be in the middle of sending a `Response`? Or does the `descrease appCounter` not run until the `Response` is completely sent?

I believe that `descrease appCounter` is called then sending `Response` is completed. But we should double-check this.

mkleczek:
> > Having looked at this, I am not so sure anymore if this PR is worth doing TBH.
> > It only covers a very specific case (ie. no traffic) - any app in progress will disable shutdown. And since `warp` accepts requests on existing connections there is a very low probability of zero active apps.
> 
> @mkleczek I just showed that our problem can be divided into two parts. This first step is mandatory for the second step. If we don't merge this first step, shutdown cannot be done immediately even if the second step is solely merged.

@kazu-yamamoto @Vlix 
I believe this first step is not necessary. In #1064 the shutdown sequence is as follows:

1. The listening socket is closed - no new connections are accepted.
2. The `shuttingDown` flag is set to `True` and we start waiting for open connections to close.
3. Each open connection can be in one of the following states (see: https://github.com/mkleczek/wai/blob/57d7ab33d8a8bf326396bd4f865ae9280f5a3b28/warp/Network/Wai/Handler/Warp/Run.hs#L60):
    - idle waiting for request - `connRecv` will throw `ShutdownInProgress` which will exit request processing loop and release the connection (see: https://github.com/mkleczek/wai/blob/57d7ab33d8a8bf326396bd4f865ae9280f5a3b28/warp/Network/Wai/Handler/Warp/HTTP1.hs#L123)
    - inside an `Application` - we allow reading from the socket and finishing the application. Once finished, `sendResponse` will check `shuttingDown` and, if it is set, add `Connection: close` header to the response and return `False` - this will exit request processing loop and release the connection (see: https://github.com/mkleczek/wai/blob/57d7ab33d8a8bf326396bd4f865ae9280f5a3b28/warp/Network/Wai/Handler/Warp/Response.hs#L139)

There is per connection `connActiveApps` that is used to check for active applications in wrapped `recv` but there is no need for global application counter and it seems redundant - am I missing something?
(Maybe the best thing to do would be to write a test that shows a hole in this reasoning)

> I like that it checks running Applications instead of connections

IMHO waiting for connections is cleaner because...

> I believe that descrease appCounter is called then sending Response is completed. But we should double-check this.

As you pointed out - waiting for connections instead of global application counter does not impose risk of sending partial responses - all connections are closed gracefully with proper `Connection: close` header set.

Vlix:
@mkleczek 

The global app counter is also a way to provide metrics and more insight into what's going on in the server. Adding it like this will make it possible to extend the API later to give users the possibility to keep track of both the amount of current connections, AND the amount of current handlers (i.e. how many requests are currently being handled)

Doing this first also shows where there might be places to improve the overall code.

I appreciate your enthousiasm to push #1064 forward, but seeing as this server code is widely used, we tend to err on the side of caution.
Your PR has given us a lot of things to think about, and I think we can sharpen our data types and functions to make changes like this less painful in the future.

---

@kazu-yamamoto 

I do think I see some situations where `connRecv` would not stop returning bytes (i.e. return `""`).
What are all the ways the warp server could stop?
  - If the `settingsInstallShutdownHandler` is set, and the socket is closed.
    (I see in this case why `connRecv` would stop returning bytes)
  - If `acceptLoop` gets thrown an async exception.
  - If `getConnMaker` throws an exception (other than `eCONNABORTED` and `eMFILE`)

I don't think this is a reason to not accept this PR, but I do wonder what would happen if someone uses `runSettingsSocket`, but doesn't stop the binary immediately afterward.
(`close :: Socket -> IO ()` is also only `bracket`ed in `runSettings`, not `runSettingsSocket` which is exposed to user space, so we can't ignore it)

If `gracefulShutdown` is reached e.g. because of an async exception thrown to `acceptLoop`, and if the socket or connections aren't actually closed, I don't see the forked threads stopping on their own.

kazu-yamamoto:
> There is per connection `connActiveApps` that is used to check for active applications in wrapped `recv` but there is no need for global application counter and it seems redundant - am I missing something?

You are right. The global application counter is not necessary to fix the issue of graceful shutdown. See #1068.

Removing the connection counter and switching to an application counter would eliminate redundancy for this PR, but the connection counter is already publicly available, so it cannot be removed.

kazu-yamamoto:
@Vlix 

> The global app counter is also a way to provide metrics and more insight into what's going on in the server. Adding it like this will make it possible to extend the API later to give users the possibility to keep track of both the amount of current connections, AND the amount of current handlers (i.e. how many requests are currently being handled)

Personally, I don't mind whether we use application counters or not. I respect @mkleczek's efforts, and if he dislikes application counters, I'm fine with adopting #1068.

Anyway, I should answer your question next.

kazu-yamamoto:
> If `gracefulShutdown` is reached e.g. because of an async exception thrown to `acceptLoop`, and if the socket or connections aren't actually closed, I don't see the forked threads stopping on their own.

In `fork`, the new thread is registered to the timeout manager. If the thread does not `tickle` the timeout handle, it will be eventually killed by the timeout manger.

Vlix:
> > There is per connection `connActiveApps` that is used to check for active applications in wrapped `recv` but there is no need for global application counter and it seems redundant - am I missing something?
> 
> You are right. The global application counter is not necessary to fix the issue of graceful shutdown. See #1068.
> 
> Removing the connection counter and switching to an application counter would eliminate redundancy for this PR, but the connection counter is already publicly available, so it cannot be removed.

The connection counter is NOT redundant. It is used to determine whether new connections can be opened when the file descriptor limit has been reached. (cf. `eMFILE`)

---

I still feel like we can add it to `Settings`. Using an opaque type, just like `Counter` to make the `ShutdownState` read-only and giving the user the possibility to read it (to maybe do cleanup outside of the webserver inside the same program, or wait on the `ShutdownState` to flip to `True` to start something else, before the `gracefulShutdown` has returned)

```hs
module Network.Wai.Handler.Warp.Internal.Example (
  ServerState, 
  makeSettingsAndServerState,
  getCurrentConnections,
  isShuttingDown,
 ) where

data ServerState = ServerState
    { ssConnCounter :: Counter
    , ssShutdownState :: TVar Bool
    }

getCurrentConnections :: ServerState -> IO Int
getCurrentConnections = getCount . ssConnCounter

isShuttingDown :: ServerState -> IO Bool
isShuttingDown = readTVarIO . ssShutdownState

-- This function would deprecate `makeSettingsAndCounter :: IO (Counter, Settings)`
-- and provide an extensible data type (`ServerState`) to later add more metrics to
-- for production users to keep track of what `warp` is doing.
makeSettingsAndServerState :: IO (ServerState, Settings)
makeSettingsAndServerState = do
    ssConnCounter <- newCounter
    ssShutdownState <- newTVarIO False
    let settings = defaultSettings
          { settingsConnectionCounter = Just ssConnCounter
          , settingsShutdownState = Just ssShutdownState
          }
    pure (ServerState{..}, settings)

-- Or we can make the following, where `defaultSettings` is pure, but doesn't
-- have a `ServerState`, which is fine for anyone who doesn't care about metrics.
-- And if there's no `ServerState`, we initialize it in the `runSettings...` functions,
-- like we already do with the `Counter`.
-- (though `ServerState` might have to be initialized sooner)
initializeSettings :: IO Settings
initializeSettings = do
   serverState <- initializeServerState -- TODO
   pure defaultSettings{settingsServerState = Just serverState}
```

kazu-yamamoto:
@Vlix What about 2f787001 in #1068?

Vlix:
> @Vlix What about [2f78700](https://github.com/yesodweb/wai/commit/2f787001b5ad6a1cc08666aa7f00c92cb916a9c8) in #1068?

That is a step in the direction I mean, but still has a bunch of API changes that are probably unnecessary.

---

I'm a bit annoyed at my current situation, as I'd like to propose my own PR, but I've been waiting on a new laptop since my current one broke. Unfortunately, it will still take a week or more, at least, so I hope you'll both (@mkleczek @kazu-yamamoto) have the patience for me to show using code what is a bit too cumbersome to describe in PR comments 🙏

Vlix:

PR #1068, opened 2026-03-25
Yet another graceful shutdown

kazu-yamamoto:
Alternative to #1067.

### DIFF
--- a/warp/Network/Wai/Handler/Warp.hs
+++ b/warp/Network/Wai/Handler/Warp.hs
@@ -93,8 +93,14 @@ module Network.Wai.Handler.Warp (
 
     -- ** Connection counter
     makeSettingsAndCounter,
+    setCounter,
     Counter,
     getCount,
+
+    -- ** Shutting down state
+    ShuttingDown,
+    setShuttingDown,
+    readShuttingDown,
 
     -- ** Exception handler
     defaultOnException,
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
todo

- [ ] documentation
- [ ] warp-tls

mkleczek:
> Alternative to #1067.

@kazu-yamamoto what is the difference between this PR and #1064 - it looks quite similar and code churn seems to be even bigger, especially that #1064 also covers and - as requested by yourself and @Vlix - was split into multiple commits (separating the one introducing connection counter)

There are now 5 PRs targeting graceful shutdowns: #1064 #1065 #1066 #1067 and this one - maybe it would be a good idea to consolidate efforts so that there is no confusion (closing dead ends and marking experiments as drafts).

kazu-yamamoto:
> @kazu-yamamoto what is the difference between this PR and #1064 - it looks quite similar and code churn seems to be even bigger, especially that #1064 also covers and - as requested by yourself and @Vlix - was split into multiple commits (separating the one introducing connection counter)

I don't think there's much difference now.

> There are now 5 PRs targeting graceful shutdowns: #1064 #1065 #1066 #1067 and this one - maybe it would be a good idea to consolidate efforts so that there is no confusion (closing dead ends and marking experiments as drafts).

@Vlix asked us to wait a little longer, so I'm waiting. I'll look at his ideas and then let's narrow them down to one.

Vlix:
I should be getting my new laptop tomorrow. I hope to make a PR before the end of the weekend.

Vlix:
Made a PR at #1071

- I'm trying not to break any APIs (not even those in `.Internal`)
- `warp-tls` should build with older versions, so there's a bunch of CPP
- I've simplified `Response.hs` a bit
- Might want to rename `makeRecv`, doesn't describe what it's doing well. If anyone has suggestions?
- Would have liked to make `appsInProgress` a `Counter`, but we don't export `newCounter`, so we wouldn't be able to create it in `warp-tls`
- `makeServerState` is idempotent, so as long as it is used before anything that needs to use the `Counter` or `ShuttingDown`, it will keep the same variables.

Vlix:

PR #1071, opened 2026-04-04
Vlix/graceful shutdown

Vlix:

After submitting your PR:

---

My attempt at showing what I meant with adding the `ShuttingDown` state to the `Settings`.
This breaks no APIs AFAICT, and the new `ServerState` can be used to give the end-user more insight into what the server is doing.
If at some point we'd want to track more metrics (e.g. "amount of active `Application`s"), this is a way to expose them in a read-only fashion.

It also makes some argument passing unnecessary, as you can get the `ServerState` with `makeServerState`.

It's unfortunate we can't change the `Settings` type without breaking backward compatibility, as I'd like to have the type reflect the presence of a `ServerState` or not. Though we could also remove `defaultSettings :: Settings` and only provide `initSettings :: IO Settings` so that the `settingsServerState` would not have to be a `Maybe`. (Maybe food for thought for a major version bump)

---

TL;DR

- I'm trying not to break any APIs (not even those in `.Internal`)
- `warp-tls` should build with older versions, so there's a bunch of CPP
- I've simplified `Response.hs` a bit
- Might want to rename `makeRecv`, doesn't describe what it's doing well. If anyone has suggestions?
- Would have liked to make `appsInProgress` a `Counter`, but we don't export `newCounter`, so we wouldn't be able to create it in `warp-tls`
- `makeServerState` is idempotent, so as long as it is used before anything that needs to use the `Counter` or `ShuttingDown`, it will keep the same variables.

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -59,6 +59,9 @@ module Network.Wai.Handler.WarpTLS (
 ) where
 
 import Control.Applicative ((<|>))
+#if MIN_VERSION_warp(3,4,13)
+import Control.Concurrent.STM (newTVarIO, TVar)
+#endif
 import Control.Exception (
     Exception,
     IOException,
@@ -90,9 +93,6 @@ import Network.Socket (
 #endif
     withSocketsDo,
  )
-#if MIN_VERSION_warp(3,4,13)
-import GHC.Conc (newTVarIO, TVar)
-#endif
 import Network.Socket.BufferPool
 import Network.Socket.ByteString (sendAll)
 import qualified Network.TLS as TLS
@@ -360,12 +360,10 @@ mkConn tlsset set s params = do
       Just bs -> switch bs
   where
     recvFirstBS = safeRecv s 4096 `onException` close s
-    switch firstBS =
-        case S.uncons firstBS of
-            Nothing -> close s >> throwIO ClientClosedConnectionPrematurely
-            Just (w8, _)
-                | w8 == 0x16 -> httpOverTls tlsset set s firstBS params
-                | otherwise -> plainHTTP tlsset set s firstBS
+    switch firstBS
+        | S.null firstBS = close s >> throwIO ClientClosedConnectionPrematurely
+        | S.head firstBS == 0x16 = httpOverTls tlsset set s firstBS params
+        | otherwise = plainHTTP tlsset set s firstBS
 
 ----------------------------------------------------------------
 
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -394,7 +394,7 @@ httpOverTls TLSSettings{..} set s bs0 params =
 #if MIN_VERSION_warp(3,4,13)
         appsInProgress <- newTVarIO 0
         (ss, _) <- makeServerState set
-        let recv = makeRecv s pool ss appsInProgress
+        let recv = makeGracefulRecv s pool ss appsInProgress
 #else
         let recv = receive s pool
 #endif
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -72,6 +72,7 @@ import Control.Exception (
     throwIO,
     try,
  )
+import qualified Control.Exception as E
 import Control.Monad (guard, void)
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as L
@@ -89,7 +90,9 @@ import Network.Socket (
 #endif
     withSocketsDo,
  )
-import qualified Control.Exception as E
+#if MIN_VERSION_warp(3,4,13)
+import GHC.Conc (newTVarIO, TVar)
+#endif
 import Network.Socket.BufferPool
 import Network.Socket.ByteString (sendAll)
 import qualified Network.TLS as TLS
@@ -274,10 +277,16 @@ runTLSSocket'
     -> Socket
     -> Application
     -> IO ()
-runTLSSocket' tlsset@TLSSettings{..} set credentials mgr sock =
-    runSettingsConnectionMakerSecure set get
+runTLSSocket' tlsset@TLSSettings{..} set credentials mgr sock app = do
+#if MIN_VERSION_warp(3,4,13)
+    (_, newSettings) <- makeServerState set
+    let get = getter tlsset newSettings sock params
+    runSettingsConnectionMakerSecure newSettings get app
+#else
+    let get = getter tlsset set sock params
+    runSettingsConnectionMakerSecure set get app
+#endif
   where
-    get = getter tlsset set sock params
     params =
         TLS.defaultParamsServer
             { TLS.serverWantClientCert = tlsWantClientCert
@@ -351,10 +360,12 @@ mkConn tlsset set s params = do
       Just bs -> switch bs
   where
     recvFirstBS = safeRecv s 4096 `onException` close s
-    switch firstBS
-        | S.null firstBS = close s >> throwIO ClientClosedConnectionPrematurely
-        | S.head firstBS == 0x16 = httpOverTls tlsset set s firstBS params
-        | otherwise = plainHTTP tlsset set s firstBS
+    switch firstBS =
+        case S.uncons firstBS of
+            Nothing -> close s >> throwIO ClientClosedConnectionPrematurely
+            Just (w8, _)
+                | w8 == 0x16 -> httpOverTls tlsset set s firstBS params
+                | otherwise -> plainHTTP tlsset set s firstBS
 
 ----------------------------------------------------------------
 
@@ -382,7 +393,14 @@ httpOverTls TLSSettings{..} set s bs0 params =
   where
     makeConn = do
         pool <- newBufferPool 2048 16384
-        rawRecvN <- makeRecvN bs0 $ receive s pool
+#if MIN_VERSION_warp(3,4,13)
+        appsInProgress <- newTVarIO 0
+        (ss, _) <- makeServerState set
+        let recv = makeRecv s pool ss appsInProgress
+#else
+        let recv = receive s pool
+#endif
+        rawRecvN <- makeRecvN bs0 recv
         let recvN = wrappedRecvN rawRecvN
         ctx <- TLS.contextNew (backend recvN) params
         TLS.contextHookSetLogging ctx tlsLogging
@@ -390,7 +408,11 @@ httpOverTls TLSSettings{..} set s bs0 params =
         mconn <- timeout tm $ do
             TLS.handshake ctx
             mysa <- getSocketName s
+#if MIN_VERSION_warp(3,4,13)
+            attachConn mysa ctx appsInProgress
+#else
             attachConn mysa ctx
+#endif
         case mconn of
           Nothing -> throwIO IncompleteHeaders
           Just conn -> return conn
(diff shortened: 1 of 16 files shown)


warp/Network/Wai/Handler/Warp/ShuttingDown.hs
```diff
@@ -0,0 +1,34 @@
...
+-- and to not expose 'writeShuttingDown' to the end user.
+module Network.Wai.Handler.Warp.ShuttingDown (
+    ShuttingDown,
+    newShuttingDown,
+    readShuttingDown,
+    readShuttingDownSTM,
+    writeShuttingDown,
+) where
+
+import GHC.Conc (
```
kazu-yamamoto: Please use `import Control.Concurrent.STM`.
The same applies to other files.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -148,17 +154,14 @@ sendResponse settings conn ii th req reqidxhdr src response = do
...
-                       else hs
     getdate = getDate ii
     addServerAndDate = addDate getdate rspidxhdr . addServer defServer rspidxhdr
     (isPersist, isChunked0) = infoFromRequest req reqidxhdr
     isChunked = not isHead && isChunked0
     (isKeepAlive, needsChunked) = infoFromResponse rspidxhdr (isPersist, isChunked)
     method = requestMethod req
     isHead = method == H.methodHead
-    rsp = case response of
+    !rsp = case response of
```
kazu-yamamoto: The bang is not necessary because we are using `Strict`.

kazu-yamamoto:
07cdada1b098fd4a5a23f16f5c0da922504e54ab should have the credit to Michał Kłeczek.

kazu-yamamoto:
I can support this PR.
Cc: @mkleczek

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -121,6 +135,22 @@ socketConnection _ s = do
             E.throwIO
             $ Sock.sendAll sock bs
 
+makeRecv :: Socket -> BufferPool -> ServerState -> TVar Int -> Recv
+makeRecv sock pool ss appsInProgress = do
+    sockWait <- waitReadSocketSTM sock
+    ok <- atomically $
```
mkleczek: I find it more difficult to follow than idiomatic `join . atomically`

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -121,6 +135,22 @@ socketConnection _ s = do
...
+        -- when shutting down throw
+        (checkShutdown >> pure False)
```
mkleczek: I'd use `checkShutdown $> False`

warp/Network/Wai/Handler/Warp/Response.hs
Vlix: Shall I remove the one below it as well then? Because the discrepancy was the main reason I added it.

warp/Network/Wai/Handler/Warp/Run.hs
Vlix: Sure, I think I'll change it.

Does the rest of the PR look good/acceptable?
Did I miss anything? Does this cover what you proposed?

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: It's like apples or oranges.
I like the original code.

warp/Network/Wai/Handler/Warp/Response.hs
kazu-yamamoto: The reason Bangs remains is historical. Let's remove them in a separate pull request. I think this pull request should provide minimal diffs.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -120,6 +120,14 @@ sendResponse
     -> IO Bool
     -- ^ Returing True if the connection is persistent.
 sendResponse settings conn ii th req reqidxhdr src response = do
+    isShuttingDown <-
+        case settingsServerState settings of
+            Just serverState -> currentShuttingDownState serverState
+            Nothing -> pure False
```
mkleczek, line 128: This is problematic: users will be confused when they find out they not only need to use `settingsInstallShutdownHandler` but also the new `makeSettingsAndServerState` to have graceful shutdowns.

TBH I find using `Settings` (which is a public API) to "smuggle" internal server state not the best decision that is going to cause more harm than good.

I am not sure how important compatibility of `Internal` API is (ie. whether the signatures of `runSettingsConnectionMakerSecure` and friends need to stay compatible), but IMHO it is less important than having a clean and cohesive public (ie. `Settings`) API.

warp/Network/Wai/Handler/Warp/Settings.hs
mkleczek, line 189: Shouldn't this field be deprecated if `makeSettingsAndCounter` is deprecated?

I think the interaction between `settingsServerState` `settingsConnectionCounter` and `settingsInstallShutdownHandler` is going to be very confusing for the users.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -351,10 +360,12 @@ mkConn tlsset set s params = do
       Just bs -> switch bs
   where
     recvFirstBS = safeRecv s 4096 `onException` close s
-    switch firstBS
-        | S.null firstBS = close s >> throwIO ClientClosedConnectionPrematurely
-        | S.head firstBS == 0x16 = httpOverTls tlsset set s firstBS params
-        | otherwise = plainHTTP tlsset set s firstBS
+    switch firstBS =
```
mkleczek: This seems like unrelated refactoring? Maybe it is a good idea to propose it in a separate PR?

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: Ahh... I see that you are initializing server state in settings in both WarpTLS and in Warp.

It is still problematic though: we strive to maintain `Internal` API signatures intact just to force clients of this API to remember to initialize one of the `Settings` fields, otherwise graceful shutdowns won't work.
IMHO, we should follow the principle of "least surprise", It would be better to simply break the clients of `Internal` API explicitly (by changing the signatures).

warp/Network/Wai/Handler/Warp/Run.hs
mkleczek: Well, at least the variable name should be different I guess. I find `ok` misguiding here (suggesting "not ok" means error).

warp/Network/Wai/Handler/Warp/Settings.hs
Vlix: Yes, `settingsConnectionCounter` would also start the deprecation track.

warp/Network/Wai/Handler/Warp/Response.hs
Vlix: The exposed API doesn't need to initialize anything, though?

Users of the API just get the opportunity to access (read-only) the shutting down state and connection counter if they want to.
If it is not initialized by the user/client, it will be initialized by the server before it is used.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -274,10 +277,16 @@ runTLSSocket'
     -> Socket
     -> Application
     -> IO ()
-runTLSSocket' tlsset@TLSSettings{..} set credentials mgr sock =
-    runSettingsConnectionMakerSecure set get
+runTLSSocket' tlsset@TLSSettings{..} set credentials mgr sock app = do
+#if MIN_VERSION_warp(3,4,13)
```
mkleczek, line 281: Why is it better that way instead of simply change version bounds in the cabal file? (genuinely asking).

At the end of the day both `warp-tls` and `warp` are released together anyway. I am doubtful anyone wants to use newer `warp-tls` with older `warp` version - what would be the point?

Vlix: If a newer version of `warp-tls` fixes a security issue or bug, and the user (for legacy reasons, or corporate reasons, or whatever) needs to use an older `warp` because they're stuck on an older GHC, or can't use certain library versions of dependencies.

I'd just like to make the transition smoother. We can get rid of the CPP once we need to bump the dependency constraints for another reason, or if we're a few more versions in, so that there's less chance of someone being in such a position.

Vlix:
Made some revisions according to PR comments, and added more documentation.
Especially, I've added a header to the `Internal.hs` module, so that users from now on will have a better expectation of why that module can change.

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: > The exposed API doesn't need to initialize anything, though?
> 
> Users of the API just get the opportunity to access (read-only) the shutting down state and connection counter if they want to. If it is not initialized by the user/client, it will be initialized by the server before it is used.

If someone is using **only** `warp-tls` or `warp` (which both initialize it) then `Internal` API compatibility does not matter anyway.
`Internal` API compatibility matters **only** for clients that do not use neither `warp-tls` nor `warp` - but in this case they need to remember to initialize server state first (ie. the changes are only syntactically compatible but semantically they are breaking anyway).

warp-tls/Network/Wai/Handler/WarpTLS.hs
mkleczek: > If a newer version of `warp-tls` fixes a security issue or bug, and the user (for legacy reasons, or corporate reasons, or whatever) needs to use an older `warp` because they're stuck on an older GHC, 

But we do not release them using different GHC versions (ie. if the user upgrades `warp-tls` it is safe to upgrade `warp`)

[...]
> 
> I'd just like to make the transition smoother. We can get rid of the CPP once we need to bump the dependency constraints for another reason, or if we're a few more versions in, so that there's less chance of someone being in such a position.

IMHO it is not worth the effort and maintenance burden, but it is your call.

Vlix:
Credited `Michał Kłeczek <michal@kleczek.org>` in the `GracefulShutdownSpec` commit.

Also want to add both `Michał Kłeczek` and `Kazu` to the merge commit when we're done with this.

warp/Network/Wai/Handler/Warp/Response.hs
Vlix: I don't see the issue.

Anything exported from `Internal.hs` at the moment either doesn't need a `ServerState` (anything other than the `socketConnection` and `runSettings*` functions) or they initialize their own if the `Settings` don't have any.

kazu-yamamoto, line 128: I guess that `Nothing` is never reached.
If so, let's put "`-- Never reached`".

kazu-yamamoto:
@Vlix The current code looks good to me. When @mkleczek is convinced, please merge this PR.

mkleczek:
> @Vlix The current code looks good to me. When @mkleczek is convinced, please merge this PR.

@Vlix @kazu-yamamoto - please, treat my comments as non-blocking.

warp/Network/Wai/Handler/Warp/Response.hs
mkleczek: > I don't see the issue.
> 
> Anything exported from `Internal.hs` at the moment either doesn't need a `ServerState` (anything other than the `socketConnection` and `runSettings*` functions) or they initialize their own if the `Settings` don't have any.

The issue is that this design violates "make invalid states unrepresentable" rule: well illustrated by comment about adding `-- Never reached`.

Anyway - I'm fine with whatever decisions you guys make :)

kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
@Vlix Why don't you merge this PR? Or, should I do it by myself?

Vlix:
I hadn't come around to it, unfortunately. But I have some time right now, so I'll merge and publish :+1:

Vlix:
Published `warp-3.4.13` and `warp-tls-3.4.14`

wolfgangwalther:
> Published `warp-3.4.13` and `warp-tls-3.4.14`

I'm missing tags for these in this repo - it seems they were forgotten.

Vlix:
Oh wow, not sure why. I could have sworn I pushed them.

Fixed it.

PR #1072, opened 2026-04-16
Add since annotation to connAppsInProgress

Bodigrim:


kazu-yamamoto (approved):
Nice catch!

kazu-yamamoto:
Merged.
Thanks!

PR #1078, opened 2026-05-25
Remove typeable deriving

Vlix:
Since we don't build on any GHC < 8.6 (because of the `base >= 4.12` constraint) all the `Typeable` deriving is automatic, so I've removed the explicit deriving.


kazu-yamamoto (approved):
LGTM

PR #1083, opened 2026-07-01
fix: Warp crashes with "thread blocked indefinitely in an STM transaction" when receiving eMFILE from accept

### LINKED ISSUE #1082: Warp crashes with "thread blocked indefinitely in an STM transaction" when receiving eMFILE from accept
See https://github.com/PostgREST/postgrest/issues/5012

It looks like the issue is this line: https://github.com/yesodweb/wai/blob/8573a706d051648811753fb48c59507dcf1c4232/warp/Network/Wai/Handler/Warp/Run.hs#L355

`warp` assumes there are request handling threads running and tries to wait for one of them to finish before re-entering accept loop. This fails when `eMFILE` is returned when no request handling threads are running (for example due to another `warp` instance running in the same process).

I am not sure if `warp` should employ such a heuristic at all (ie. that it is itself that is the cause of file descriptor exhaustion), but even if it does,  the waiting logic should be "wait for counter decreased or zero".

mkleczek:
Fixes #1082

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Counter.hs
+++ b/warp/Network/Wai/Handler/Warp/Counter.hs
@@ -6,6 +6,7 @@ module Network.Wai.Handler.Warp.Counter (
     waitForZero,
     increase,
     decrease,
+    waitForCounter,
     waitForDecreased,
     getCount,
     getCountSTM,
@@ -26,11 +27,14 @@ waitForZero (Counter var) = atomically $ do
     when (x > 0) retry
 
 waitForDecreased :: Counter -> IO ()
-waitForDecreased (Counter var) = do
-    n0 <- atomically $ readTVar var
+waitForDecreased = waitForCounter (>)
+
+waitForCounter :: (Int -> Int -> Bool) -> Counter -> IO ()
+waitForCounter condition (Counter var) = do
+    n0 <- readTVarIO var
     atomically $ do
         n <- readTVar var
-        check (n < n0)
+        check (condition n0 n)
 
 increase :: Counter -> IO ()
 increase (Counter var) = atomically $ modifyTVar' var $ \x -> x + 1
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -352,7 +352,8 @@ acceptConnection set getConnMaker app counter ii = do
                 if | isErrno eCONNABORTED -> acceptNewConnection
                    | isErrno eMFILE -> do
                        settingsOnException set Nothing $ E.toException e
-                       waitForDecreased counter
+                       -- if there are requests in flight wait for (at least) one to finish
+                       waitForCounter (\prev next -> prev <= 0 || prev > next) counter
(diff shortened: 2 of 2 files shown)


Vlix:
Does this fix the problem? Or just cause `warp` to go into a failing loop?

Does the connection that triggers the `eMFILE` being thrown get aborted immediately? Or does it also wait indefinitely until the connection is properly accepted?

Vlix:
Current code:
https://github.com/yesodweb/wai/blob/8573a706d051648811753fb48c59507dcf1c4232/warp/Network/Wai/Handler/Warp/Run.hs#L345-L359

Ok, I've been trying things out locally to understand how the `accept` function works and how it interacts with file descriptors. And what kind of implications that behavior would have when ignored/retried/etc.

What I've gathered (at least in Linux) is that before Warp accepts a new connection, it first allocates a file descriptor. So when at `getConnMaker` there are no more file descriptors available, `accept` will immediately throw `eMFILE`, even without an outside connection attempt.

This should make it `safe` to just continue with `waitForDecreased` if the counter is already at 0.

The only issue I see is that this will be spamming `settingsOnException` into infinity. So we'd want / need to have a flag that gets unset once it jumps out of the `eMFILE` branch. Something like this?

```hs
data FdIssues = NoFdIssue | FdExhausted
  deriving (Show, Eq)

...
fdExcRef <- newIORef NoFdIssue
let resetFdExcRef = writeIORef fdExcRef NoFdIssue
acceptNewConnection = do
    ...
    case ex of
      Right x -> resetFdExcRef >> pure (Just x)
      Left e -> do
         if | eCONNABORTED ->  resetFdExcRef >> acceptNewConnection
            | eMFILE -> do
                 fdIssue <- readIORef fdExcRef
                 when (fdIssue == NoFdIssue) $
                     settingsOnException set Nothing $ E.toException e
                 wasAtZero <- waitForDecreased counter
                 when wasAtZero $ writeIORef fdExcRef FdExhausted
                 acceptNewConnection
            | otherwise -> resetFdExcRef >> ...
```

@kazu-yamamoto @mkleczek Do you have a strong opinion here? How would you like `settingsOnException` to be called? And how often?

kazu-yamamoto:
Sorry but I'm busy this week.
I can return to this issue next week.

mkleczek:
When working on https://github.com/PostgREST/postgrest/pull/5064 I realized that maybe the "wait for counter decreased" heuristic should be moved to `defaultAccept` letting clients decide how to handle `EMFILE` instead of hardcoding it here?

Vlix:
Users being able to handle it is fine. Maybe even preferable to it being handled internally. But the fact that Warp can basically hard-lock on a situation which is not impossible, is something that should be fixed.

So if you can fix it in your situation which is optimal for your situation, then I would definitely recommend it 👍 but we're going to have to think of a way to also fix it in the default Warp situation so this doesn't bite others.

kazu-yamamoto:
I'm back.
I think this PR should be taken over by #1084.

kazu-yamamoto:
Closing in favor of #1084.

PR #1084, opened 2026-07-03
Fix wait for decreased at zero connections

Vlix:

After submitting your PR:

---

Trying my fix of #1082

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -21,7 +21,7 @@ import Control.Concurrent.STM (
 import qualified Control.Exception as E
 import qualified Data.ByteString as S
 import Data.Functor (($>))
-import Data.IORef (newIORef, readIORef)
+import Data.IORef (newIORef, readIORef, IORef, writeIORef)
 import Data.Streaming.Network (bindPortTCP)
 import Foreign.C.Error (Errno (..), eCONNABORTED, eMFILE)
 import GHC.Conc.Sync (labelThread, myThreadId)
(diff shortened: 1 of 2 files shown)


steve-chavez:
Hey @Vlix,

JFYI, I can confirm your fix solves https://github.com/PostgREST/postgrest/issues/5012. Tried your branch on https://github.com/PostgREST/postgrest/pull/5078.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -266,8 +266,9 @@ runSettingsConnectionMakerSecure
 runSettingsConnectionMakerSecure oldSettings getConnMaker app = do
     settingsBeforeMainLoop oldSettings
     (ServerState{serverConnectionCounter}, newSettings) <- makeServerState oldSettings
-    withII newSettings $
-        acceptConnection newSettings getConnMaker app serverConnectionCounter
+    withII newSettings $ \ii ->
+        initFdExhaustionRef >>=
```
kazu-yamamoto, line 270: @Vlix Can't you push `IORef FdExhaustion` into `InternalInfo`?

Vlix: I think I can, but `InternalInfo` is passed done also to `http1` and `http2` and further. I'd prefer not to "leak" this concept deep into the HTTP handling if it's only on the connection level.

Do you think it's better/cleaner to add it to `InternalInfo`?

kazu-yamamoto: It is not a strong opinion, and if this form was decided upon after careful consideration, then that is fine.

kazu-yamamoto (approved):
I understand that the bug regarding the file descriptor limit was caused by a misassumption on my part. I had mistakenly believed that only the `Warp` thread would increase or decrease file descriptors within the process, and I failed to consider the possibility that other threads might also change them. I am approving this PR because its semantics is clear and it has been verified to resolve the issue.

kazu-yamamoto:
@Vlix Thank you for the nice fix. Please feel free to release a new version by yourself.

Vlix:
I'll see if I can do it this weekend.
I'm also looking at another bug I'd like to fix. #956 I'll probably also do that this weekend.

PR #1085, opened 2026-07-09
Add requestSendEarlyHints and wire it through Warp's HTTP/2 handler

thomasjm:
Follow-on from
* https://github.com/kazu-yamamoto/http-semantics/pull/18
* https://github.com/kazu-yamamoto/http2/pull/168

After submitting your PR:

### DIFF
--- a/wai/Network/Wai.hs
+++ b/wai/Network/Wai.hs
@@ -61,6 +61,7 @@ module Network.Wai (
     requestHeaderRange,
     requestHeaderReferer,
     requestHeaderUserAgent,
+    requestSendEarlyHints,
     -- $streamingRequestBodies
     strictRequestBody,
     consumeRequestBodyStrict,
@@ -322,6 +323,7 @@ defaultRequest =
         , requestHeaderRange = Nothing
         , requestHeaderReferer = Nothing
         , requestHeaderUserAgent = Nothing
+        , requestSendEarlyHints = \_ -> return ()
         }
 
 -- | A @Middleware@ is a component that sits between the server and application.
--- a/wai/Network/Wai/Internal.hs
+++ b/wai/Network/Wai/Internal.hs
@@ -91,6 +91,16 @@ data Request = Request
     -- ^ The value of the User-Agent header in a HTTP request.
     --
     -- @since 3.2.0
+    , requestSendEarlyHints :: H.ResponseHeaders -> IO ()
+    -- ^ Send a @103 Early Hints@ informational response carrying the given
+    -- headers, ahead of the final response. This lets a client (typically over
+    -- HTTP\/2) start fetching resources named in @Link@ headers while the final
+    -- response is still being produced.
+    --
+    -- A handler (e.g. Warp over HTTP\/2) that supports early hints installs an
+    -- action here; otherwise it is a no-op.
+    --
+    -- @since 3.2.5
     }
 
 -- | Get the next chunk of the body. Returns 'B.empty' when the
--- a/warp/test/EarlyHintsSpec.hs
+++ b/warp/test/EarlyHintsSpec.hs
@@ -1,0 +1,72 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module EarlyHintsSpec (spec) where
+
+import Test.Hspec
+
+#define HAS_EARLY_HINTS_SUPPORT (MIN_VERSION_http_semantics(0,4,1) && MIN_VERSION_http2(5,4,2))
+
+#if HAS_EARLY_HINTS_SUPPORT
+import Control.Exception (bracket)
+import Data.ByteString (ByteString)
+import Data.IORef
+import Network.HPACK (TokenHeaderTable, getFieldValue)
+import Network.HPACK.Token (toToken)
+import Network.HTTP.Types (Status, methodGet, ok200, status200, status404)
+import qualified Network.HTTP2.Client as C
+import Network.Socket
+import Network.Wai
+import Network.Wai.Handler.Warp (Port, testWithApplication)
+
+spec :: Spec
+spec = describe "HTTP/2 Early Hints" $
+    it "delivers a WAI app's 103 Early Hints to the client before the final response (h2c)" $
+        testWithApplication (pure app) $ \port -> do
+            hintsRef <- newIORef []
+            earlyHintsClient port hintsRef `shouldReturn` Just ok200
+            hints <- readIORef hintsRef
+            map (getFieldValue (toToken "link") . snd) hints
+                `shouldBe` (Just <$> earlyResponses)
+
+-- | The @Link@ header values delivered as Early Hints, in order.
+earlyResponses :: [ByteString]
+earlyResponses =
+    [ "</style.css>; rel=preload; as=style"
+    , "</app.js>; rel=preload; as=script"
+    ]
+
+-- | A WAI app that emits two Early Hints sections, then the final response.
+app :: Application
+app req respond
+    | pathInfo req == ["early"] = do
+        mapM_ (\link -> requestSendEarlyHints req [("link", link)]) earlyResponses
+        respond $ responseLBS status200 [("content-type", "text/plain")] "Hello"
+    | otherwise = respond $ responseLBS status404 [] ""
+
+-- | Drive Warp over h2c with the HTTP/2 client, recording each 103 Early Hints
+--   section via the client's informational handler, and return the final status.
+earlyHintsClient :: Port -> IORef [TokenHeaderTable] -> IO (Maybe Status)
+earlyHintsClient port hintsRef = withTCP "127.0.0.1" port $ \sock ->
+    bracket (C.allocSimpleConfig sock 4096) C.freeSimpleConfig $ \conf ->
+        C.run cliconf (conf{C.confOnInformational = onInformational}) $ \sendRequest _aux ->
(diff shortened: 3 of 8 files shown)


wai/Network/Wai.hs
```diff
@@ -207,6 +212,36 @@ responseRaw = ResponseRaw
...
+-- hints (for example HTTP\/1.0, or a handler that does not populate
+-- 'earlyHintsKey'), so it is always safe to call. It may be called several
+-- times to emit multiple Early Hints sections before the final response.
+--
+-- @since 3.2.5
+sendEarlyHints :: Request -> [H.Header] -> IO ()
+sendEarlyHints req hdrs =
+    case Vault.lookup earlyHintsKey (vault req) of
+        Just send -> send hdrs
+        Nothing -> return ()
```
Vlix: It feels really weird having the function to send a response inside the vault of a request.

Vlix (commented):
Can you explain why doing early hints can't be implemented with `responseStream`?

```hs
withEarlyHints :: H.Headers -> IO Response -> Response
withEarlyHints hdrs ioResponse =
    responseStream status103 hdrs $ \send flush ->
        -- insert handling the actual `ioResponse`
```

thomasjm:
> Can you explain why doing early hints can't be implemented with `responseStream`?

I'm not totally sure that it can't, but see my response [here](https://github.com/kazu-yamamoto/http-semantics/pull/18#issuecomment-4941113268).

wai/Network/Wai.hs
thomasjm: Yes, another implementation approach I considered was to put a new field on the `Request`, something like this:

```haskell
requestSendEarlyHints :: [H.Header] -> IO ()
```

I was hoping to discuss the options here in review.

FWIW, there is some precedent of using the vault for similar things. Like `pauseTimeout`, which reaches into the connection to pause its slowloris timeout.

thomasjm:
Hi @kazu-yamamoto -- thanks for your attention to https://github.com/kazu-yamamoto/http-semantics/pull/18 and https://github.com/kazu-yamamoto/http2/pull/168!

I thought I'd ping you on this one also in case you missed it, would love to know what you think about this part.

kazu-yamamoto:
@ thomasjm We ran out of time yesterday.

First, this PR conflicts with `master` because other PRs were merged into it. Please rebase onto `master` and resolve the conflicts; the resolution is straightforward. The next version is 3.4.15.

I didn't understand the intention behind including `sendEarlyHints` and `earlyHintsKey` in `wai`. If it is acceptable to include them in `warp` instead, that would limit the scope of the impact.

thomasjm:
Just rebased!

> I didn't understand the intention behind including sendEarlyHints and earlyHintsKey in wai. If it is acceptable to include them in warp instead, that would limit the scope of the impact.

One of my goals is to open a PR against [http-reverse-proxy](https://hackage.haskell.org/package/http-reverse-proxy) to add 103 early hints support there. This package provides a reverse proxy in the form of a WAI `Application`. Under the hood it uses `http-client`, which already has a callback called when early hint headers are received. My goal is to wire that up to the WAI application, which in the current design would look like this:

https://github.com/fpco/http-reverse-proxy/commit/39a8e109b4702f0c783c01b1494d32a5c4680393

`http-reverse-proxy` is not tied to `warp`; I'd like it to continue to work with any WAI server.

wai/Network/Wai.hs
Vlix: It's still something that isn't WAI specific. It's more of an implementation detail (just like `pauseTimeout`)

So I agree with Kazu that this would be better suited to be in `warp`.

warp/warp.cabal
Vlix, line 256: This should also be bumped to the version that supports `auxSendInformational`.
Though I'd rather we also depend on `http-semantics` (instead of importing Aux/Server/etc. via `http2`) and add CPP to only use it when it's actually exposed in the dependency. Especially if `http2` will use CPP too.

Vlix (changes requested):
The change to WAI is unnecessary and `wai` package should stay focused and small.

I also don't like new versions of packages requiring the newest version of dependencies, since this breaks upgrade paths.

warp/warp.cabal
thomasjm: Okay, I've revamped the approach to something I'm hoping you'll like better. Let me explain in a new comment below.

wai/Network/Wai.hs
thomasjm: Okay, I've revamped the approach to the `requestSendEarlyHints` version, let me explain below why I think it's still needed.

thomasjm:
Okay, after the feedback I've switched it to the `requestSendEarlyHints` approach I mentioned, with a focus on minimizing the impact to `wai`.

The only change to `wai` is adding the `requestSendEarlyHints` field to the `Request` object, which is necessary for downstream libraries like `http-reverse-proxy` to make use of this in a server-agnostic way. I'd argue that this is in the spirit of WAI: 103 early hints are part of the "interface" of HTTP (they're part of the spec after all). So it makes sense that WAI would include a hook in the interface of `Request`, while leaving all implementation to servers like `warp`.

A few debatable calls I made:
* I bumped `wai` to `3.3.0`. I don't think a minor version bump is strictly necessary, since the `Request` constructor is not exposed (aside from via `Internal` module), so this could change.
* I bumped the `warp` bound on `wai` to `3.3.0`. Assuming the decision above stands, I think this makes sense because `wai` and `warp` live in the same repo and can be published together.
* In `warp`, I added an explicit (but loose) bound on `http-semantics` and left the `http2` bound as it was, preferring to gate the new feature with CPP for these ones. This means `warp` will continue to compile against older versions.

Would you mind taking another look @Vlix? Thanks!

kazu-yamamoto (commented):
This PR now makes sense to me.
Thank you for your good work!
Let's wait for @Vlix's review.

wai/Network/Wai/Internal.hs
```diff
@@ -91,6 +91,16 @@ data Request = Request
     -- ^ The value of the User-Agent header in a HTTP request.
     --
     -- @since 3.2.0
+    , requestSendEarlyHints :: [H.Header] -> IO ()
```
Vlix: Please use the `ResponseHeaders` type synonym here.
It has better backwards compatibility.

Vlix (changes requested):
I got back from vacation today, and this looks better so far, but I haven't looked at the new tests yet. I'll do that tomorrow.

In the meanwhile, please bump the upper bounds for _any_ `wai` dependency in the following cabal files:
```yaml
- wai-app-static
- wai-conduit
- wai-extra
- wai-frontend-monadcgi
- wai-websockets
- warp-tls
```
This way, the CI will actually run.

wai/Network/Wai/Internal.hs
thomasjm: Done

thomasjm:
Okay, done! CI is green now except for one apparent flake.

warp/test/EarlyHintsSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+import Network.Socket
+import Network.Wai
+import Network.Wai.Handler.Warp (Port, testWithApplication)
+
+spec :: Spec
+spec = describe "HTTP/2 Early Hints" $
+    it "delivers a WAI app's 103 Early Hints to the client before the final response (h2c)" $
+        testWithApplication (pure app) $ \port -> do
+            hintsRef <- newIORef []
+            earlyHintsClient port hintsRef >>= (`shouldBe` Just ok200)
```
Vlix:
```hs
earlyHintsClient port hintsRef `shouldReturn` Just ok200
```

warp/test/EarlyHintsSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+            hints <- readIORef hintsRef
+            map (getFieldValue (toToken "link") . snd) hints
+                `shouldBe` [ Just "</style.css>; rel=preload; as=style"
```
Vlix: I would like to see a constant list of the results, so that it's obvious it should be the same.

```hs
earlyResponses :: [ByteString]
earlyResponses =
  [ "</style.css>; rel=preload; as=style"
  , "</app.js>; rel=preload; as=script"
  ]

-- ...
    `shouldBe` (Just <$> earlyResponses)

-- ...
    traverse (requestSendEarlyHints req) $ (,) "link" <$> earlyResponses
```

or something of the sort

Vlix (changes requested):
Test looks good. I think this is workable.

After the changes, this can be merged and I'll rebase the other open PRs that will go with the `3.4.15` release.

EDIT: Also also, maybe change the version bump for `wai` to `3.2.5`, since we're using PVP, not SemVer; `3.3.0` would be a new major version bump.

warp/test/EarlyHintsSpec.hs
thomasjm: Done

warp/test/EarlyHintsSpec.hs
thomasjm: Done

Vlix (approved):

kazu-yamamoto:
Merged.

thomasjm:
Thanks for the reviews all!

Will there be a new release of `wai`/`warp` soon?

Vlix:
Probably in about a week, as I'll be gone this week and I will be cleaning up the current open PRs and publish the new versions when I get back.

Issue #620, opened 2017-05-17
Question: 'connection: close' header is ignored?


NCrashed:
Hi!

Warp from `lts-8.13` doesn't close connection when `Connection: close` header is passed.

Code: 
``` haskell
#!/usr/bin/env stack
{- stack --install-ghc runghc
--package warp
-}

{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE OverloadedStrings #-}
import Network.Wai
import Network.HTTP.Types
import Network.Wai.Handler.Warp (run)

app :: Application
app _ respond = do
    putStrLn "I've done some IO here"
    respond $ responseLBS
        status200
        [("Content-Type", "text/plain")]
        "Hello, Web!"

main :: IO ()
main = do
    putStrLn $ "http://localhost:8090/"
    run 8090 app 

```

Curl request:
``` bash
[ncrashed@localhost roc-cluster]$ curl "http://localhost:8089" -k -v -H "Connection: close"
* Rebuilt URL to: http://localhost:8089/
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8089 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* connect to 127.0.0.1 port 8089 failed: Connection refused
* Failed to connect to localhost port 8089: Connection refused
* Closing connection 0
curl: (7) Failed to connect to localhost port 8089: Connection refused
[ncrashed@localhost roc-cluster]$ curl "http://localhost:8090" -k -v -H "Connection: close"
* Rebuilt URL to: http://localhost:8090/
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8090 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8090 (#0)
> GET / HTTP/1.1
> Host: localhost:8090
> User-Agent: curl/7.51.0
> Accept: */*
> Connection: close
> 
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< Date: Wed, 17 May 2017 12:54:01 GMT
< Server: Warp/3.2.12
< Content-Type: text/plain
< 
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
Hello, Web!
```

[...]

PR #1086, opened 2026-07-13
Revise keepalive and count streaming bodies (and more small stuff)

Vlix:

---

Wanted to tackle #956 and ended up encountering more small instances that were either outdated or (technically) wrong.

So from largest changes to smallest:
- Revised keep alive / persistence logic
- Replaced sanitization of header values (no more multiline folding, as per RFC 9110)
- Count message bodies better for `settingsLogger` purposes
- Only add headers if not already set, AND handle "Transfer-Encoding" better
- Documentation/comments

### DIFF
--- a/wai/Network/Wai.hs
+++ b/wai/Network/Wai.hs
@@ -120,6 +120,14 @@ import System.IO.Unsafe (unsafeInterleaveIO)
 
 -- | Creating 'Response' from a file.
 --
+-- Server implementations like @warp@ might disregard the t'Status' when
+-- using 'responseFile', since the server might rework the response based on
+-- headers or presence/absence of the file.
+-- @warp@ does not do do any extra processing when sending file parts, though,
+-- so be mindful of how each server implementation handles file responses.
+--
+-- /The above was written when @warp-3.4.14@ was the newest version/
+--
 -- @since 2.0.0
 responseFile
     :: H.Status -> H.ResponseHeaders -> FilePath -> Maybe FilePart -> Response
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -15,12 +15,8 @@
 * Also counts bytes for streaming responses for `settingsLogger`'s size argument.
   And fixed size for builder responses, since those included the status and header
   lines. Now the `Maybe Integer` argument in `settingsLogger` is consistently
-  just the message body's bytes.
+  just the amount of bytes of the sent message body.
   [#1086](https://github.com/yesodweb/wai/pull/1086)
-* Add phantom type parameter to `IndexedHeader` to catch request/response
-  mismatches at compile time.
-  [#1088](https://github.com/yesodweb/wai/pull/1088)
-
 * Reworked internal indexed headers to records for performance and to remove
   dependencies on `array`.
   [#1092](https://github.com/yesodweb/wai/pull/1092) [#1093](https://github.com/yesodweb/wai/pull/1093)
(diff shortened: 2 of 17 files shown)


Vlix:
I do want to talk about `ResponseRaw` and the "Connection" header in responses.

1) Do we want to ALWAYS close connections when `ResponseRaw` is used? Because it wasn't that way until this PR. The "fallback" was used to determine whether to use `isPersist` or `False`. If `hasBody {fallbackStatus}`, then it would be `ret == False`, but if the fallback has status 1XX/204/304, then it would persist. (if the request "Connection" header would dictate so)
2) I've now added that if the user logic adds `"Connection: close"`, then we honor this and close the connection. To me that makes sense, but might there have been a reason it was not in the logic?

wai/Network/Wai.hs
```diff
@@ -119,6 +119,14 @@ import System.IO.Unsafe (unsafeInterleaveIO)
 
 -- | Creating 'Response' from a file.
 --
+-- Server implementations like @warp@ might disregard the t'Status' when
```
kazu-yamamoto, line 123: I don't understand why "t" exists before 'Status'.

kazu-yamamoto:
@Vlix `ResponseRaw` was originally introduced by Michael and I cannot speak on his behalf. I support making changes as you think correct.

wai/Network/Wai.hs
Vlix: Ah, it is haddock specific syntax to indicate this should refer to the `type` constructor `Status`, not the `data` constructor `Status`.

Haddock will emit warnings if the type and data constructor are the same and the documentation doesn't specify which is meant.

warp/ChangeLog.md
```diff
@@ -1,5 +1,17 @@
...
+## 3.4.15
+
+* Rework keep alive logic for HTTP/1.X so connections won't be automatically
+  closed on HEAD requests anymore. Should conform more to spec in general.
+* Replace multiline header value support with sanitizing newlines and NUL bytes
+  with spaces. (as per RFC 9110 section 5.5)
+* Also counts bytes for streaming responses for `settingsLogger`'s size argument.
+  And fixed size for builder responses, since those included the status and header
+  lines. Now the `Maybe Integer` argument in `settingsLogger` is consistently
+  just the message body's bytes.
```
Vlix: Note to self:

`is consistently just the amount of bytes of the sent message body`

kazu-yamamoto:
This PR conflicts with the current `master`.
@Vlix You would be better suited than I am to handle the task of resolving the conflict.

Vlix:
I've found that `toBufIOWith` might not be counting the length correctly?

The `Chunk` steps might not be counted at all, if I understand the `ByteString.Builder` source code correctly. Any `insertChunk` doesn't touch the buffer range (which is used to calculate the amount of bytes written) and just the `ByteString` is returned. This would mean the `toBufferIO` step would give an empty `ByteString` to the `io` action, and the `io bs` in the `Chunk` section doesn't get added to the `totalBytesSent`.

Vlix:
Finally done with everything.
This was a bit of a back and forth, but the counting of sent response body bytes, updating sanitation of response header values, and cleaning up the keep alive logic is now all in place.
This PR can be regarded as fixing #956, and maybe #620?

Vlix:
It's a bit late now, but note to self for later:
  - [x] publish `wai-3.2.5`
  - [x] publish `warp-3.4.15`

PR #1087, opened 2026-07-13
upper bound for 'http-types'

Vlix:
Because I'm preparing for a major-major rework of headers (in the far future), we should add upper bounds in future releases ASAP. The packages that do not use headers from 'http-types' do not need an upper bound. ('wai' uses type synonyms only, which is forwards-compatible)

I have not bumped any versions, since this can just be updated when a new release is made.
We're going to have to revise a bunch of older versions anyway if an `http-types-1.x` is released.

### DIFF
--- a/wai-app-static/wai-app-static.cabal
+++ b/wai-app-static/wai-app-static.cabal
@@ -27,7 +27,7 @@ library
     build-depends:   base                      >= 4.12     && < 5
                    , wai                       >= 3.0      && < 3.3
                    , bytestring                >= 0.10.4
-                   , http-types                >= 0.7
+                   , http-types                >= 0.7      && < 1
                    , transformers              >= 0.2.2
                    , unix-compat               >= 0.2
                    , directory                 >= 1.0.1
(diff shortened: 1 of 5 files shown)


wai-app-static/wai-app-static.cabal
```diff
@@ -27,7 +27,7 @@ library
     build-depends:   base                      >= 4.12     && < 5
                    , wai                       >= 3.0      && < 3.3
                    , bytestring                >= 0.10.4
-                   , http-types                >= 0.7
+                   , http-types                >= 0.7      && < 1
```
kazu-yamamoto, line 30: `warp` requires `> 0.12`. Why don't you require the same here?

Vlix: The lowest acceptable `warp` version (`3.0.11`) has `http-types >= 0.8.5`, so it's not even too far off. But keeping those the same is unnecessary. If this package can build with `0.7`, that's enough for version resolution. 

It's also technically more correct from a holistic point of view, since a future `warp` version _COULD_ (theoretically) allow `http-types >= 0.1`, and then suddenly the restriction here would be "wrong" (since it _CAN_ build with lower than `0.12`)

kazu-yamamoto (approved):
Now LGTM.

kazu-yamamoto:
Rebased and merged.

PR #1088, opened 2026-07-13
Safer IndexedHeader 

Vlix:

---

Unfortunately, the `IndexedHeader` type is exposed in the `Internal` module, but this is a safer version without any extra overhead.
And this `RequestHeaderIndex` wasn't exposed anyway, the `IndexedHeader` from `recvRequest` can't really be inspected reliably, so the types seem to only be exposed to allow users to use `recvRequest` into `sendResponse`.
This is why I feel this is a safe enough change to make, which _might_ break something, but I don't foresee a large amount of packages using `recvRequest/sendResponse`.

If you know any packages that use that API, do tell.

### DIFF
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -1,5 +1,12 @@
 # ChangeLog for warp
 
+## 3.4.15
+
+* Add phantom type parameter to `IndexedHeader` to catch request/response
+  mismatches at compile time. This does change the API of
+  `Network.Wai.Handler.Warp.Internal`, so this gets a minor version bump.
+  [#1088](https://github.com/yesodweb/wai/pull/1088)
+
 ## 3.4.14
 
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
Is my understanding correct? In this PR, phantom types are used to distinguish between the index types for requests and responses, thereby enhancing safety.

> If you know any packages that use that API, do tell.

I don't know such packages.

Vlix:
That is exactly what's happening here, yes.

kazu-yamamoto (approved):
LGTM.

Vlix:
I'll be away for a week on vacation. Afterwards, I'll combine #1086 with this (and #1087) before publishing a new version, I think.

kazu-yamamoto:
Have a nice vacation.
I will do what I can, but I'll leave the release up to you.

PR #1090, opened 2026-07-19
warp: hot-path performance optimizations (+101% keep-alive throughput, -51% allocation)

seanparsons:
## Summary

A series of hot-path optimizations for warp (plus small changes to `recv` and `time-manager`), found by profiling and syscall-tracing a keep-alive hello-world server and verified end to end. Cumulative effect on the same hardware/workload (wrk -t4 -c100, server `-N4`, GHC 9.10.3):

|  | stock warp 3.4.15 | this branch |
|---|---|---|
| hello-world throughput | 35.7k req/s | **71.7k req/s (+101%)** |
| allocation per request | 13.7 KB | **6.7 KB (-51%)** |
| `sendResponse` (4-header builder response) | 1.58 us | **290 ns (-82%)** |
| syscalls per request | recv + send + epoll_ctl + ~28 futex ops | **recv + send** |

Also validated against a real production Servant app (full middleware stack, DB pool, logging): +57-80% throughput on light endpoints, +27% on a 124KB HTML page, ~40% lower latency, byte-identical response bodies.

## The two big ones

**Optimistic recv (last commit).** `makeGracefulRecv` (introduced with graceful shutdown, on the hot path for every `runSettings` server) registers with the IO manager and parks in STM on *every* receive, even though under load the next request is nearly always already in the kernel buffer. The branch tries a direct non-blocking `recv(2)` first and only falls back to the STM/shutdown machinery on would-block. A syscall census went from ~1 `epoll_ctl` + ~28 `futex` per request to 186 and 725 *total* across 198k requests.

**Timer traffic (commit 5).** Every request performed 3+ operations against GHC's global TimerManager (pause/resume/tickle; streaming did resume+pause per fragment). `time-manager` now caches the TimerManager in the Handle and rate-limits `tickle` to one real `updateTimeout` per min(1s, timeout/4) using a monotonic-clock check; streaming keeps the timeout armed with a rate-limited tickle per fragment. This commit alone was +45% throughput.

The remaining commits: single-pass response-header handling, composing the header directly into the connection write buffer (saves a full memcpy + pinned allocation per response), strict-record header indexes on both request and response side, a cached `ForeignPtr` in `WriteBuffer`, and a lazily-built request vault.

## Correctness

- Full warp spec suite passes at every commit boundary I built (and on the final tree).
- Wire bytes diffed against master for content-length, chunked, and many-header responses: byte-identical including chunked framing.
- Timeout behavior after the timer changes verified manually: idle, keep-alive-then-idle, and stalled-streaming connections all closed at ~3.01s with `setTimeout 3`.
- The vault laziness was proven with a temporary trace probe (plus a control run).

## Behavior changes maintainers should weigh

1. **Timeout precision** (timer commit): timeouts can fire up to min(1s, timeout/4) early relative to the last suppressed tickle.
2. **Streaming timeout semantics** (timer commit): the timeout is now armed while user code runs between fragments, so a `responseStream` that emits nothing for a full timeout period is killed (previously it could stall forever between fragments). Streams that write at least once per timeout period, or send heartbeats, are unaffected. This closes a slowloris-shaped gap but is observable for e.g. SSE without heartbeats.
3. **Graceful shutdown edge** (recv commit): a request already in the kernel buffer when shutdown begins is now served (pre-graceful-shutdown behavior for bytes the client already sent).
4. **API changes**: `Connection.connAppsInProgress` and `makeGracefulRecv` change `TVar Int` -> `IORef Int` (exposed via `Internal`); `WriteBuffer` gains a `bufFPtr` field; `Header.hs` drops the now-unused array machinery. The STM->IORef commit measured neutral on its own, so I'm happy to drop it if the API churn isn't worth it.
5. `receiveNoWait` (recv) uses a direct FFI `recv(2)` because `Network.Socket.recvBufNoWait` isn't exported from `network`'s public API; on Windows it returns `Nothing` (unchanged behavior). If exporting `recvBufNoWait` upstream is preferred I can switch to it.

## Benchmarks included

The first commit adds `warp:bench:response`, an end-to-end criterion benchmark of `sendResponse` with a sink `Connection` — the response side previously had no benchmark coverage at all. Each commit message carries its measured effect.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -11,3 +11,5 @@ packages:
     wai-websockets
     wai-conduit
     time-manager
+
+benchmarks: True
--- a/recv/Network/Socket/BufferPool.hs
+++ b/recv/Network/Socket/BufferPool.hs
@@ -24,9 +24,11 @@ module Network.Socket.BufferPool (
     -- * Recv
     Recv,
     receive,
+    receiveNoWait,
     BufferPool,
     newBufferPool,
     withBufferPool,
+    tryWithBufferPool,
 
     -- * RecvN
     RecvN,
--- a/recv/Network/Socket/BufferPool/Buffer.hs
+++ b/recv/Network/Socket/BufferPool/Buffer.hs
@@ -1,6 +1,7 @@
 module Network.Socket.BufferPool.Buffer (
     newBufferPool,
     withBufferPool,
+    tryWithBufferPool,
     mallocBS,
     copy,
 ) where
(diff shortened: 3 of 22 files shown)


seanparsons:
Splitting this into independently reviewable PRs as offered, each build- and spec-verified standalone:

- #1091: response-path benchmark (the measurement rig the others quote)
- #1092: single-pass response header handling
- #1093: request-header index as a strict record
- #1094: compose header directly into the write buffer
- #1095: cached ForeignPtr in WriteBuffer
- #1096: lazy request vault
- #1097: time-manager, cached TimerManager + rate-limited tickle
- #1098: warp, fewer timer ops around streaming bodies (pairs with #1097)
- #1099: optimistic non-blocking recv (recv + warp, the biggest single win)
- ~~#1100: in-flight counter STM → IORef~~ (withdrawn: measured neutral, not worth the API churn)

This branch remains as the combined reference (+101% keep-alive throughput, −51% allocation when everything is applied together).

PR #1091, opened 2026-07-19
warp: add end-to-end response-path benchmark

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs.

The existing bench suite covers only request-line parsing and header splitting; the response side had no benchmark at all. This adds `warp:bench:response`, a criterion benchmark driving the real `sendResponse` end to end with a sink `Connection`, a real time-manager `Handle`, and a realistic indexed request header, plus isolated benches for `composeHeader` and the header indexers. Compiled `-threaded` with the library's `Strict`/`StrictData` to match production semantics. Also sets `benchmarks: True` in `cabal.project`.

Baseline it establishes (GHC 9.10.3): builder response with 4 headers 1.58µs, 20 headers 2.80µs, 204-no-body 727ns. The other PRs in the series quote measurements from this benchmark.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -11,3 +11,5 @@ packages:
     wai-websockets
     wai-conduit
     time-manager
+
+benchmarks: True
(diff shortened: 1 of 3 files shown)


Vlix:
I'm not opposed to having more benchmarks.
Don't see anything off with the implementation, so this is fine as a first "response benchmark" attempt.

We'll merge this after `3.4.15` is published, I think.

PR #1092, opened 2026-07-19
warp: stop rebuilding clean response headers, index in one pass

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs; independent of the others (touches only the response side of `Header.hs`, `Response.hs`, `File.hs`).

`sendResponse` rebuilt the response header list even when nothing needed sanitizing (the overwhelmingly common case) and byte-scanned every value with `S.any`; `indexResponseHeader` built a boxed `Data.Array` although only four slots are ever consulted. Now: a memchr-based (`S.elemIndex`) dirty check returns the input list untouched when clean, and the response index is a four-field strict record filled in one traversal. Dirty-value sanitization output is unchanged.

Measured (bench from #1091): `sendResponse` 4 headers 1.58µs → 1.39µs (−12%), 20 headers 2.80µs → 2.08µs (−26%); ≈ −425 B allocated per request under keep-alive load. Spec suite passes.

Rebase note: this and the request-side index PR both touch `Header.hs`; whichever merges second has a trivial conflict (they modify disjoint halves; together they make the shared array machinery fully dead, and the second PR deletes it).

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Header.hs
+++ b/warp/Network/Wai/Handler/Warp/Header.hs
@@ -4,13 +4,12 @@
 module Network.Wai.Handler.Warp.Header (
     IndexedHeader,
     IndexedRequestHeader,
-    IndexedResponseHeader,
+    IndexedResponseHeader (..),
     (!),
     RequestHeaderIndex (..),
     indexRequestHeader,
     requestMaxIndex,
     defaultIndexRequestHeader,
-    ResponseHeaderIndex (..),
     indexResponseHeader,
 ) where
 
@@ -29,7 +28,6 @@ import Network.Wai.Handler.Warp.Types
 newtype IndexedHeader a = IxHeader (Array Int (Maybe HeaderValue))
 
 type IndexedRequestHeader = IndexedHeader RequestHeaderIndex
-type IndexedResponseHeader = IndexedHeader ResponseHeaderIndex
 
 -- | Safer way to lookup 'IndexedHeader' values
 (!) :: Enum a => IndexedHeader a -> a -> Maybe HeaderValue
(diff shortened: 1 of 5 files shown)


warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -180,16 +180,20 @@ sendResponse settings conn ii th req reqidxhdr src response = do
 
 ----------------------------------------------------------------
 
+-- Values without CR/LF (the overwhelmingly common case) leave the
+-- header list untouched; only a dirty value triggers a rebuild.
 sanitizeHeaders :: H.ResponseHeaders -> H.ResponseHeaders
-sanitizeHeaders = map (sanitize <$>)
+sanitizeHeaders hs
```
Vlix, line 187: This function can also be added to the benchmark.

I think the following will be just as quick, if not quicker:
```hs
-- This will be rewritten to `memchr || memchr` via the `anyByte` rule from `bytestring`
containsNewlines v = S.any (_lf ==) v || S.any (_cr ==) v
```

warp/Network/Wai/Handler/Warp/Header.hs
```diff
@@ -104,29 +102,40 @@ defaultIndexRequestHeader =
...
+-- | Index for the response headers Warp itself consults.
+--   Only these four headers are ever looked up on the response side,
+--   so a flat record built in a single traversal beats a boxed array.
+--   The fields are strict via StrictData.
+data IndexedResponseHeader = IndexedResponseHeader
+    { resContentLength :: Maybe HeaderValue
+    , resServer :: Maybe HeaderValue
+    , resDate :: Maybe HeaderValue
+    , resLastModified :: Maybe HeaderValue
+    }
```
Vlix: The values aren't ever checked, just the presence, so this can be a collection of `Bool`s.

```hs
data ResponseHeaderPresence = ResponseHeaderPresence
    { hasContentLength :: Bool
    , hasServer :: Bool
    , hasDate :: Bool
    , hasLastModified :: Bool
    }
```

Vlix (changes requested):
Looks ok.

kazu-yamamoto (approved):
LGTM

Vlix:
@kazu-yamamoto I would like to keep this (and all the PRs up to #1099) out of master for the time being, as together with #1091 - #1099 this does touch a lot of code, and I'd like to take the time to also run the benchmarks to see if these changes are worth it. (first #1091 and then the rest one by one)

So please don't merge these (#1091 - #1099) PRs for a while. I'll be away next week as well, but will have a lot of time afterwards.

kazu-yamamoto:
@Vlix I support your opinion.

Vlix:
I'm merging this and #1093 into master before rebasing my PR because these touch a bit of the same code, and having these changes in would also "invalidate" PR #1088. And since that is exposed in the `Internal.hs` module, I'd rather only affect it once.

PR #1093, opened 2026-07-19
warp: replace boxed-array request-header index with a strict record

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs; independent of the others (request side of `Header.hs` + call sites in `Request.hs`, `Response.hs`, `File.hs`).

`indexRequestHeader` built a boxed `Array Int (Maybe ByteString)` via `runSTArray` per request, with bounds-checked reads + `fromEnum` at each use. It becomes a 13-field strict record filled in one `foldl'` (later duplicates win, matching the old `writeArray` behavior); `requestKeyIndex` keeps its length-keyed dispatch but returns `Maybe` instead of a −1 sentinel; call sites use plain selectors.

Measured: `indexRequestHeader` 60ns → 53ns (−11%), ≈ −400 B allocated per request. `Internal` still exports `IndexedRequestHeader`/`defaultIndexRequestHeader`. Spec suite passes.

Rebase note: shares `Header.hs` with the response-side header PR; second-to-merge resolves a trivial conflict and deletes the then-dead array machinery.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Header.hs
+++ b/warp/Network/Wai/Handler/Warp/Header.hs
@@ -3,18 +3,16 @@
 
 module Network.Wai.Handler.Warp.Header (
     IndexedHeader,
-    IndexedRequestHeader,
+    IndexedRequestHeader (..),
     IndexedResponseHeader,
     (!),
-    RequestHeaderIndex (..),
     indexRequestHeader,
-    requestMaxIndex,
     defaultIndexRequestHeader,
     ResponseHeaderIndex (..),
     indexResponseHeader,
 ) where
 
-import Data.Array (Array, array)
+import Data.Array (Array)
 import qualified Data.Array as A ((!))
 import Data.Array.ST
 import qualified Data.ByteString as BS
@@ -28,7 +26,6 @@ import Network.Wai.Handler.Warp.Types
 -- | Array for a set of HTTP headers.
 newtype IndexedHeader a = IxHeader (Array Int (Maybe HeaderValue))
 
-type IndexedRequestHeader = IndexedHeader RequestHeaderIndex
 type IndexedResponseHeader = IndexedHeader ResponseHeaderIndex
 
 -- | Safer way to lookup 'IndexedHeader' values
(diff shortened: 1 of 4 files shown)


warp/Network/Wai/Handler/Warp/Header.hs
```diff
@@ -54,53 +66,66 @@ data RequestHeaderIndex
...
--- - \"If-Match\"
--- - \"If-None-Match\"
-requestMaxIndex :: Int
-requestMaxIndex = fromEnum (maxBound :: RequestHeaderIndex)
-
-requestKeyIndex :: HeaderName -> Int
+indexRequestHeader :: RequestHeaders -> IndexedRequestHeader
+indexRequestHeader = foldl' insert defaultIndexRequestHeader
+  where
+    insert ix (key, val) = case requestKeyIndex key of
```
Vlix: You can just inline `requestKeyIndex` here.
No need to keep using `RequestHeaderIndex`.

(I can tell #1092 and #1093 were generated. A human would have just done the same thing twice.)

Vlix (changes requested):
@kazu-yamamoto Do you think it's worth the "speedup" to move to using a record instead of an array?

I feel like adding/removing a `RequestHeaderIndex` or adding/removing a field from the record is about the same amount of work, so I don't feel too strongly about whichever is used.

If a value isn't used (the `If(None)Match` ones) the type can also just be a `Bool`, even.

kazu-yamamoto:
@Vlix I suppose we could say it becomes safer in the sense that we don't have to worry about index overruns with this change.
However, I don't have any strong opinion on the matter.

warp/Network/Wai/Handler/Warp/Header.hs
seanparsons: Inlined it in https://github.com/yesodweb/wai/pull/1093/commits/c4324623f0af36a97ebb75b33a92388ddbdc8ed8

Vlix:
Rebased and merged via #1101

PR #1094, opened 2026-07-19
warp: compose response header directly into the write buffer

seanparsons:

For `RspBuilder` responses (responseBuilder/responseLBS, the most common type) the header block was copied twice: composed into a fresh pinned ByteString, then copied again into the connection write buffer (headers are nearly always < 4096 bytes, so bytestring's builder copies rather than inserts). Now the header is composed directly into the write buffer (`composeHeaderPtr`/`composeHeaderLength` in ResponseHeader, sharing the existing copy loops) and the body builder continues after it via `unsafeToBufIOWithOffset` (`toBufIOWith = unsafeToBufIOWithOffset 0`, no duplicated loop). Oversized headers fall back to the old path. The length reported to the logger is unchanged.

### Correctness

Wire bytes verified byte-identical against current master for content-length, chunked, and 20-header responses, including chunked framing (`Transfer-Encoding: chunked` and the terminating `0\r\n` chunk). Re-verified against master @ 79b0e6b7 rather than the original base. Spec suite: 121 examples, 0 failures.

### Measurements

Against master @ 79b0e6b7, `warp:bench:response`, GHC 9.10.3, 7 rounds alternating master/branch from prebuilt binaries, pinned with `taskset -c 3`, medians across rounds.

Allocation, bytes per iteration, via `--regress allocated:iters +RTS -T` (R² = 1.000, so this is exact rather than sampled):

| bench | master | branch | |
|---|---|---|---|
| builder 4 headers content-length | 2385 | 1842 | -22.8% |
| builder 3 headers chunked | 2514 | 1986 | -21.0% |
| builder 20 headers content-length | 4539 | 3434 | -24.4% |
| no body 204 (untouched path) | 1009 | 1009 | 0% |
| headers/* (untouched) | unchanged | unchanged | 0% |

That is roughly 540 B less allocated per response, and the paths this PR does not touch allocate byte-identically.

Times, same runs: -4.8% (4 headers), -5.2% (chunked), -3.7% (20 headers).

**Treat the timings as weak evidence.** In the same sweep `indexResponseHeader`, which this PR does not touch, moved +7.7% between the two binaries purely from code layout, so a few percent on these sub-microsecond benchmarks is noise. The allocation figures are the claim worth reviewing. (An earlier revision of this description quoted -6% and -11%; those did not reproduce under interleaved runs against current master and have been dropped.)

Two notes for anyone re-running these:

- The `sendResponse` benchmarks report severely inflated variance. That is not GC (`+RTS -s` shows 99.7% productivity). `strace -c` counts ~313k `write` syscalls for a single `sendResponse` benchmark, about one per iteration, from `T.tickle th` reaching `wakeManager`. The timer wake-up dominates both the jitter and a good part of the absolute number, which is why the allocation regression is the more useful signal here. That cost is what #1097 targets.
- Comparing distributions rather than single runs caught a real regression in this branch: `composeHeaderPtr` was not being inlined, leaving `composeHeader` with an out-of-line call plus a `minusPtr` it discards (~4% slower than master). Fixed in the latest commit with an `INLINE` pragma, which also inlines it into `sendRsp`.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/IO.hs
+++ b/warp/Network/Wai/Handler/Warp/IO.hs
@@ -4,23 +4,38 @@ import Control.Exception (mask_)
 import Data.ByteString.Builder (Builder)
 import Data.ByteString.Builder.Extra (Next (Chunk, Done, More), runBuilder)
 import Data.IORef (IORef, readIORef, writeIORef)
+import Foreign.Ptr (plusPtr)
 import Network.Wai.Handler.Warp.Buffer
 import Network.Wai.Handler.Warp.Imports
 import Network.Wai.Handler.Warp.Types
 
 toBufIOWith
     :: Int -> IORef WriteBuffer -> (ByteString -> IO ()) -> Builder -> IO Integer
-toBufIOWith maxRspBufSize writeBufferRef io builder = do
+toBufIOWith = toBufIOWithOffset 0
+
+-- | Like 'toBufIOWith' but the first @offset@ bytes of the write buffer
+-- are assumed to be already filled (e.g. with a response header composed
+-- directly into the buffer). They are flushed together with the first
+-- batch of builder output and included in the returned total.
+-- @offset@ must not exceed the current buffer size.
+toBufIOWithOffset
+    :: Int
+    -> Int
+    -> IORef WriteBuffer
+    -> (ByteString -> IO ())
+    -> Builder
+    -> IO Integer
+toBufIOWithOffset offset0 maxRspBufSize writeBufferRef io builder = do
     writeBuffer <- readIORef writeBufferRef
-    loop writeBuffer firstWriter 0
+    loop writeBuffer offset0 firstWriter 0
   where
     firstWriter = runBuilder builder
-    loop writeBuffer writer bytesSent = do
+    loop writeBuffer offset writer bytesSent = do
         let buf = bufBuffer writeBuffer
             size = bufSize writeBuffer
-        (len, signal) <- writer buf size
-        bufferIO buf len io
-        let totalBytesSent = toInteger len + bytesSent
+        (len, signal) <- writer (buf `plusPtr` offset) (size - offset)
+        bufferIO buf (offset + len) io
+        let totalBytesSent = toInteger (offset + len) + bytesSent
         case signal of
             Done -> return totalBytesSent
             More minSize next
@@ -43,8 +58,8 @@ toBufIOWith maxRspBufSize writeBufferRef io builder = do
                         biggerWriteBuffer <- createWriteBuffer minSize
                         writeIORef writeBufferRef biggerWriteBuffer
                         return biggerWriteBuffer
-                    loop biggerWriteBuffer next totalBytesSent
-                | otherwise -> loop writeBuffer next totalBytesSent
+                    loop biggerWriteBuffer 0 next totalBytesSent
+                | otherwise -> loop writeBuffer 0 next totalBytesSent
             Chunk bs next -> do
                 io bs
-                loop writeBuffer next totalBytesSent
+                loop writeBuffer 0 next totalBytesSent
--- a/warp/Network/Wai/Handler/Warp/IO.hs
+++ b/warp/Network/Wai/Handler/Warp/IO.hs
@@ -11,21 +11,30 @@ import Network.Wai.Handler.Warp.Types
 
 toBufIOWith
     :: Int -> IORef WriteBuffer -> (ByteString -> IO ()) -> Builder -> IO Integer
-toBufIOWith = toBufIOWithOffset 0
+toBufIOWith = unsafeToBufIOWithOffset 0
 
 -- | Like 'toBufIOWith' but the first @offset@ bytes of the write buffer
 -- are assumed to be already filled (e.g. with a response header composed
 -- directly into the buffer). They are flushed together with the first
 -- batch of builder output and included in the returned total.
--- @offset@ must not exceed the current buffer size.
-toBufIOWithOffset
+--
+-- === WARNING: @offset@ MUST NOT exceed the current buffer size!!!
+--
+-- This function performs NO bounds checking on @offset@. The builder is
+-- handed the pointer @buffer + offset@ with @bufSize - offset@ bytes of
+-- claimed free space, so an oversized @offset@ points past the end of the
+-- allocation and advertises negative capacity, i.e. out-of-bounds writes
+-- and memory corruption. Every caller MUST verify
+-- @offset < bufSize@ of the current write buffer first (see the
+-- @hdrLen@ check in 'Network.Wai.Handler.Warp.Response.sendRsp').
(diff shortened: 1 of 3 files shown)


warp/Network/Wai/Handler/Warp/IO.hs
```diff
@@ -4,23 +4,38 @@ import Control.Exception (mask_)
...
+-- batch of builder output and included in the returned total.
+-- @offset@ must not exceed the current buffer size.
+toBufIOWithOffset
+    :: Int
+    -> Int
+    -> IORef WriteBuffer
+    -> (ByteString -> IO ())
+    -> Builder
+    -> IO Integer
+toBufIOWithOffset offset0 maxRspBufSize writeBufferRef io builder = do
```
Vlix: I'd like the `@offset@ must not exceed the current buffer size.` section to be a lot more visible, with more exclamation marks even. This is a very important invariant that should NOT be broken, as this could lead to memory issues.

It might even warrant the addition of `unsafe` to the title of the function.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -237,21 +238,32 @@ sendRsp conn _ _ ver s hs _ _ _ RspNoBody = do
...
-            | otherwise = header <> body
-        writeBufferRef = connWriteBuffer conn
+    writeBuffer <- readIORef writeBufferRef
     len <-
-        toBufIOWith
-            maxRspBufSize
-            writeBufferRef
-            (\bs -> connSendAll conn bs >> T.tickle th)
-            hdrBdy
+        if hdrLen < bufSize writeBuffer
```
Vlix, line 303: This check is AMAZINGLY important, so please add a comment above it explaining why it's important that the `hdrLen` should be less than the buffer size when using `composeHeaderPtr`.

Vlix (changes requested):
The introduction of `toBufIOWithOffset` does introduce ways that future code changes could result in bad behaviour/errors/security issues. But since it's pretty local and not widely used, it should be fine.

---

Off-topic:

I also think we've overlooked the `Chunk` section, because the length of the `bs` there isn't counted towards the `totalBytesSent`... I think? I've dug through the `ByteString.Builder` source code to find how it works, and from what I can tell, the amount of bytes written would be 0, and the `ByteString` is just given... in which case the `totalByteSent` value would add `len == 0`, and then the `Chunk` is sent... without counting it???

A test would help us get this right.

I think I'll make a new issue for this.

warp/Network/Wai/Handler/Warp/IO.hs
seanparsons: Bigger warning and `unsafe` added to the name in https://github.com/yesodweb/wai/pull/1094/commits/151f69b401a954dff703bd5205b7d63daf4fa68b

warp/Network/Wai/Handler/Warp/Response.hs
seanparsons: Comment added above the condition in https://github.com/yesodweb/wai/pull/1094/commits/151f69b401a954dff703bd5205b7d63daf4fa68b

Vlix:
Will continue with this PR and the others when I have time to also run the benchmarks.

This PR and the others might have to be rebased since my PR might have changed how some things are handled.

Vlix:
I've rebased this branch locally and compared it to master, and the increase to me (on a ThinkPad with minimal other programs running) is fairly minimal.

@seanparsons How did you run the benchmarks? (And or, please try them again against current master)
The `sendResponse` benchmarks are also VERY inflated (near 100%) so it's difficult to discern the actual results.

seanparsons:
@Vlix Re-ran everything against current master (`79b0e6b7`), and you are right on both counts: the timings are marginal, and the `sendResponse` benches are badly inflated. Details, including the cause of the inflation:

### How I ran them

```
cabal build warp:bench:response          # GHC 9.10.3, cabal's default -O1 for benches
taskset -c 3 ./response --csv out.csv --regress allocated:iters +RTS -T
```

7 rounds alternating master/branch (master, branch, master, branch, ...) from prebuilt
binaries, one core, medians across rounds. Interleaving matters: consecutive runs of the
same binary drift together, so a plain "run A then run B" comparison picks up machine
state as if it were signal.

No rebase was needed, the branch was already based on the current master head.

### Why the sendResponse benches are inflated

It is not GC. `+RTS -s` reports 99.7% productivity and 0.014s of GC over a 5s run.

It is the timeout manager. `strace -c` on a single `sendResponse` bench counts
**313,420 `write` syscalls**, about one per iteration, plus futex traffic. That is
`T.tickle th` reaching `wakeManager`, so every iteration pays a syscall and a
cross-thread wake-up. That is what the outlier variance is measuring, and it is also
a large part of the absolute number, which is why a header-copy saving is hard to see
in it. (Aside: this is the cost #1097 targets.)

The workaround is criterion's allocation regression, `--regress allocated:iters +RTS -T`,
which reports **R² = 1.000** here. It is deterministic and immune to the timer jitter.

### Numbers vs current master

Allocation, bytes/iteration, exact:

| bench | master | branch | |
|---|---|---|---|
| builder 4 headers content-length | 2385 | 1842 | **−22.8%** |
| builder 3 headers chunked | 2514 | 1986 | **−21.0%** |
| builder 20 headers content-length | 4539 | 3434 | **−24.4%** |
| no body 204 | 1009 | 1009 | 0% |
| headers/* | unchanged | unchanged | 0% |

≈540 B less per response. The untouched paths (204, the `headers` group) allocate
byte-identically, which is the control I would want to see.

Times, medians of 7: −4.8% (4 headers), −5.2% (chunked), −3.7% (20 headers).

**But treat those time deltas with suspicion.** `indexResponseHeader`, which this PR does
not touch at all, moved +7.7% between the two binaries in the same sweep. That is code
layout, and it sets the honest noise floor for these sub-100ns benches. The −6%/−11% in
the PR description does not reproduce at this rigour and I would not defend it. The
allocation drop is the claim I would stand behind.

### The re-run caught a regression

Comparing distributions rather than single runs showed standalone `composeHeader` was
consistently ~4% *slower* on the branch (master 125-133ns vs branch 130-138ns, barely
overlapping). Cause: `composeHeaderPtr` was not being inlined, so `composeHeader` paid an
out-of-line call plus a `minusPtr` it discards. `{-# INLINE composeHeaderPtr #-}` fixes it
(+3.8% → +0.2%) and, because it also inlines into `sendRsp`, it is what moved the
`sendResponse` times from −2.5%/−0.1%/−3.5% to −4.8%/−5.2%/−3.7%.

Pushed as e2da1fb. Thanks for pushing back on this, the regression would have shipped
otherwise. warp spec suite: 121 examples, 0 failures.

I'd also suggest I correct the PR description, since its -6%/-11% is the number that
sent you looking in the first place.

Vlix (approved):
The code seems fine and I think it's ok to add this as a slight optimisation to throughput and allocation.

PR #1095, opened 2026-07-19
warp: cache a ForeignPtr in WriteBuffer, drop per-flush allocation

seanparsons:

`bufferIO` called `newForeignPtr_` + `PS` on every write-buffer flush (at least once per response) and `toBuilderBuffer` once per streaming response, allocating a fresh wrapper for a pointer that is stable for the `WriteBuffer`'s lifetime. `WriteBuffer` now carries a `bufFPtr` minted once in `createWriteBuffer` (finalizer-free; `bufFree` remains the only free path). The one site that replaces a `WriteBuffer` (growth in `toBufIOWith`) goes through `createWriteBuffer`, so pointer and ForeignPtr cannot drift.

Following review, `sendFile`, `readSendFile` and `packHeader` now take the `WriteBuffer` itself instead of a `Buffer` plus a `BufSize`, so the file-sending paths reach the cached ForeignPtr too and `rawBufferIO` is gone entirely. Both call sites (`Run.socketConnection` and warp-tls `attachConn`) already had the `WriteBuffer` in hand and took it apart, so they get shorter.

API note: this changes the API of `Network.Wai.Handler.Warp.Internal` (the new `WriteBuffer` field, and the `sendFile`/`readSendFile` signatures), so there is a ChangeLog entry under the unreleased 3.4.15 section. `createWriteBuffer` remains the sanctioned constructor. warp-tls is updated in the same commit. Spec suite passes (including large-response buffer growth).

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -485,8 +485,7 @@ attachConn mysa ctx = do
         sendfile fid offset len hook headers = do
             writeBuffer <- I.readIORef writeBufferRef
             readSendFile
-                (bufBuffer writeBuffer)
-                (bufSize writeBuffer)
+                writeBuffer
                 sendall
                 fid
                 offset
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -27,6 +27,11 @@
 * Rework internal indexed headers to records for performance and to remove
   dependencies on `array`.
   [#1092](https://github.com/yesodweb/wai/pull/1092) [#1093](https://github.com/yesodweb/wai/pull/1093)
+* `WriteBuffer` now carries a cached `ForeignPtr` of its buffer, so flushing a
+  response no longer allocates a fresh wrapper. `sendFile` and `readSendFile`
+  now take the `WriteBuffer` instead of a `Buffer` and a `BufSize`, which
+  changes the API of `Network.Wai.Handler.Warp.Internal`.
+  [#1095](https://github.com/yesodweb/wai/pull/1095)
 
 ## 3.4.14
 
--- a/warp/Network/Wai/Handler/Warp/Buffer.hs
+++ b/warp/Network/Wai/Handler/Warp/Buffer.hs
@@ -4,6 +4,7 @@ module Network.Wai.Handler.Warp.Buffer (
     freeBuffer,
     toBuilderBuffer,
     bufferIO,
+    rawBufferIO,
 ) where
 
 import Data.IORef (IORef, readIORef)
@@ -23,9 +24,11 @@ import Network.Wai.Handler.Warp.Types
 createWriteBuffer :: BufSize -> IO WriteBuffer
 createWriteBuffer size = do
     bytes <- allocateBuffer size
+    fptr <- newForeignPtr_ bytes
     return
         WriteBuffer
             { bufBuffer = bytes
+            , bufFPtr = fptr
             , bufSize = size
             , bufFree = freeBuffer bytes
             }
@@ -48,12 +51,19 @@ freeBuffer = free
 toBuilderBuffer :: IORef WriteBuffer -> IO B.Buffer
 toBuilderBuffer writeBufferRef = do
     writeBuffer <- readIORef writeBufferRef
-    let ptr = bufBuffer writeBuffer
+    let fptr = bufFPtr writeBuffer
+        ptr = bufBuffer writeBuffer
         size = bufSize writeBuffer
-    fptr <- newForeignPtr_ ptr
     return $ B.Buffer fptr ptr ptr (ptr `plusPtr` size)
 
-bufferIO :: Buffer -> Int -> (ByteString -> IO ()) -> IO ()
-bufferIO ptr siz io = do
+-- | Slice the given number of bytes out of a 'WriteBuffer' using its
+-- cached 'ForeignPtr', without allocating a fresh wrapper.
+bufferIO :: WriteBuffer -> Int -> (ByteString -> IO ()) -> IO ()
+bufferIO writeBuffer siz io = io $ PS (bufFPtr writeBuffer) 0 siz
+
+-- | Like 'bufferIO' for callers that only have a raw pointer.
+-- This allocates a fresh 'ForeignPtr' wrapper on every call.
+rawBufferIO :: Buffer -> Int -> (ByteString -> IO ()) -> IO ()
+rawBufferIO ptr siz io = do
     fptr <- newForeignPtr_ ptr
     io $ PS fptr 0 siz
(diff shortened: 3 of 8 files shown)


Vlix (commented):
@kazu-yamamoto Do you expect anything going wrong if the `ForeignPtr` is included in the `WriteBuffer`?

If we accept this change, I think we can also refactor the `(read)sendFile` functions to accept a `WriteBuffer` instead of a `Buffer`, since the `WriteBuffer` is also exposed in `Internal` AND the functions basically already need everything from the `WriteBuffer` (except the `bufFree`). Though it might be unnecessary API churn for very minimal gain...

kazu-yamamoto:
If @seanparsons could show us how much faster this change would make things, I think we could discuss it further.

seanparsons:
> If @seanparsons could show us how much faster this change would make things, I think we could discuss it further.

This one on its own makes very little speed difference if any but on responses with headers makes about a 1% reduction in allocations.

warp/Network/Wai/Handler/Warp/Buffer.hs
```diff
@@ -48,12 +51,19 @@ freeBuffer = free
...
-bufferIO :: Buffer -> Int -> (ByteString -> IO ()) -> IO ()
-bufferIO ptr siz io = do
+-- | Slice the given number of bytes out of a 'WriteBuffer' using its
+-- cached 'ForeignPtr', without allocating a fresh wrapper.
+bufferIO :: WriteBuffer -> Int -> (ByteString -> IO ()) -> IO ()
+bufferIO writeBuffer siz io = io $ PS (bufFPtr writeBuffer) 0 siz
+
+-- | Like 'bufferIO' for callers that only have a raw pointer.
+-- This allocates a fresh 'ForeignPtr' wrapper on every call.
+rawBufferIO :: Buffer -> Int -> (ByteString -> IO ()) -> IO ()
```
Vlix: This would be unnecessary if we'd have `sendFile` get the `WriteBuffer` instead of only the `Buffer`. Though that would change the API in the `Internal` module.

I only see `warp-tls` make use of the `readSendFile`, but it also takes apart the `WriteBuffer` there. So I'm in favor of changing the API to get rid of `rawBufferIO`.

Vlix (changes requested):
I'm ok with this change, if we get rid of `rawBufferIO` by changing the `Buffer -> BufSize` arguments of `(read)sendFile` to `WriteBuffer`.

(Also, this needs a rebase on master)

warp/Network/Wai/Handler/Warp/Buffer.hs
seanparsons: Agreed, done in 0edecac2. `sendFile`, `readSendFile` and `packHeader` now take the `WriteBuffer` instead of a `Buffer` plus a `BufSize`, so every flush path goes through `bufferIO` and `rawBufferIO` is gone.

Both call sites already read the `WriteBuffer` out of the `IORef` and took it apart, so `Run.socketConnection` and warp-tls `attachConn` both get a little shorter. The `WriteBuffer` carries the size as well, so nothing is lost by dropping the separate `BufSize` argument.

This does change the API of `Network.Wai.Handler.Warp.Internal`, so I added a bullet to the existing unreleased 3.4.15 section of the ChangeLog next to the `IndexedHeader` one from #1088. Say the word if you would rather that be its own version section.

warp and warp-tls both build clean, and the spec suite passes (120 examples, 0 failures), including the `packHeader` and `readSendFile` tests which now build a real `WriteBuffer` via `createWriteBuffer`. The Windows branch of `readSendFile` is edited but not compiled here; it loses its local `newForeignPtr_` and uses the cached one instead.

kazu-yamamoto:
No objections from me.

warp/ChangeLog.md
```diff
@@ -27,6 +27,11 @@
 * Rework internal indexed headers to records for performance and to remove
   dependencies on `array`.
   [#1092](https://github.com/yesodweb/wai/pull/1092) [#1093](https://github.com/yesodweb/wai/pull/1093)
+* `WriteBuffer` now carries a cached `ForeignPtr` of its buffer, so flushing a
```
Vlix, line 30: This will be added to the next `3.5.0` release.

warp-tls/Network/Wai/Handler/WarpTLS.hs
```diff
@@ -485,8 +485,7 @@ attachConn mysa ctx = do
         sendfile fid offset len hook headers = do
             writeBuffer <- I.readIORef writeBufferRef
             readSendFile
-                (bufBuffer writeBuffer)
-                (bufSize writeBuffer)
+                writeBuffer
```
Vlix, line 488: This needs to go behind CPP so it doesn't lose the lower bound for `warp`.

`warp-3.5.0` will get this new API, so it should split on that version.

Vlix (changes requested):

Vlix:
@seanparsons We're ready to prep for a new major version.

Please move the ChangeLog entry to a new `warp-3.5.0` section. And add CPP to the `warp-tls` diff so that `warp-tls` will work for both `warp < 3.5` and `warp >= 3.5`.
Please bump the patch version of `warp-tls` (i.e. `3.4.14.1`), bump the upper bounds of the `warp` dependency(s) to `< 3.6`, and add an entry to the ChangeLog "Build with `warp-3.5.0`".

warp/ChangeLog.md
seanparsons: Moved into the 3.5.0 section in 0f9eacaf.

warp-tls/Network/Wai/Handler/WarpTLS.hs
seanparsons: Now split on `MIN_VERSION_warp(3,5,0)` in 0f9eacaf. I compiled both sides: against the in-tree 3.5.0 and against warp-3.4.16 from Hackage with `warp < 3.5` constrained.

seanparsons:
Done in 0f9eacaf, on top of a fresh merge of master.

- The ChangeLog entry now sits in the `warp-3.5.0` section.
- `warp-tls` picks the `readSendFile` call shape with `MIN_VERSION_warp(3,5,0)`, so it builds against `warp < 3.5` and `warp >= 3.5`. I compiled both sides: against the in-tree 3.5.0 (warp and warp-tls specs green) and against `warp-3.4.16` from Hackage with `warp < 3.5` constrained.
- `warp-tls` is bumped to 3.4.14.1 with a "Build with `warp-3.5.0`" ChangeLog entry.
- The `< 3.6` bounds were already raised on master in 82b2b8ae, so the branch picks those up through the merge and the commit doesn't touch them. Shout if you'd rather see them in this PR's diff anyway.

Vlix:
Everything looks good, I think.
I'll be away this weekend, but I'll give it another look-over (and check that I'm not missing anything) and merge this in next week.

PR #1096, opened 2026-07-19
warp: build the request vault lazily

seanparsons:

Every HTTP/1 request eagerly built a Vault with 2-3 inserts (5-6 for HTTP/2), each allocating map spine + a Locker, though most applications never read the vault. warp is compiled with `-XStrict`, so a lazy pattern (`~`) on the vault binding keeps the insert chain a thunk that only materializes if a handler forces `vault req` (the field is lazy in wai's `Request`).

Laziness proven with a temporary trace probe: handlers ignoring the vault never build it; a handler forcing it builds it exactly once; a control run with the `~` removed fired on every request. Contents when read are identical. ≈ −200 B/request. Spec suite passes.


Vlix (approved):
Seems fine.

@kazu-yamamoto Do you expect anything to go wrong with a lazy vault?

kazu-yamamoto:
@Vlix The reason for using `Strict` and `StrictData` is that they make the behavior more intuitively understandable. While `Strict` and `StrictData` are used by default, there is absolutely no issue with selectively introducing laziness where necessary.

Vlix (approved):
Seems fine to me 👍

PR #1097, opened 2026-07-19
time-manager: cache the TimerManager and rate-limit tickle

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs. Standalone: benefits current warp releases as-is (warp already tickles on every request); compounds with the companion warp streaming PR.

Every tickle/pause/resume/cancel re-fetched the global TimerManager (a global IORef read), and every tickle did a real `updateTimeout` against the process-wide timer queue (warp tickles per request and per recv/send chunk). The `Handle` now caches the TimerManager at registration (stable for the process lifetime; per-Handle caching also keeps `TimeoutKey`s paired with their issuing manager), records the monotonic time of the last registration/update, and rate-limits tickle to one real update per min(1s, timeout/4). The common-case tickle is one VDSO clock read + one IORef read + compare, zero timer-queue traffic. The timeout/4 floor caps the precision loss at 25% of short timeouts.

Semantics: timeouts can fire up to min(1s, timeout/4) early relative to the last suppressed tickle (documented on `tickle`). Verified manually with `setTimeout 3`: idle, keep-alive-then-idle, and stalled-streaming connections all closed at ~3.01s. time-manager and warp test suites pass.

Measured together with the warp streaming PR on a keep-alive hello-world server: +45% throughput (35.7k → 51.9k req/s), `sendResponse` 1.58µs → 436ns, −3.5KB allocation/request. The largest single win in the series.

### DIFF
--- a/time-manager/System/TimeManager.hs
+++ b/time-manager/System/TimeManager.hs
@@ -48,8 +48,10 @@ module System.TimeManager (
 
 import Control.Concurrent (forkIO, mkWeakThreadId, myThreadId)
 import qualified Control.Exception as E
-import Control.Monad (void)
+import Control.Monad (void, when)
 import qualified Data.IORef as I
+import Data.Word (Word64)
+import GHC.Clock (getMonotonicTimeNSec)
 import System.Mem.Weak (deRefWeak)
 import System.TimeManager.Internal
 
@@ -73,8 +75,11 @@ emptyHandle =
     Handle
         { handleTimeout = 0
         , handleAction = pure ()
+        , handleTimerManager = error "time-manager: Handle.handleTimerManager not set"
         , handleKeyRef = error "time-manager: Handle.handleKeyRef not set"
         , handleState = error "time-manager: Handle.handleState not set"
+        , handleLastRenewed = error "time-manager: Handle.handleLastRenewed not set"
+        , handleMinRenewGap = 0
         }
 
 ----------------------------------------------------------------
@@ -126,39 +131,68 @@ register :: Manager -> TimeoutAction -> IO Handle
 register mgr@(Manager timeout) onTimeout
     | isNoManager mgr = pure emptyHandle
     | otherwise = do
+        -- The system timer manager is stable for the lifetime of the
+        -- process (and even if it were replaced, e.g. around a fork,
+        -- the key registered below would only be meaningful to the
+        -- manager it was registered with). So fetch it once here and
+        -- cache it in the 'Handle' instead of re-reading the global
+        -- IORef on every tickle/pause/resume.
         sysmgr <- getTimerManager
         key <- EV.registerTimeout sysmgr timeout onTimeout
         keyref <- I.newIORef key
         state <- I.newIORef Active
+        now <- getMonotonicTimeNSec
+        lastRenewed <- I.newIORef now
         let h =
                 Handle
                     { handleTimeout = timeout
                     , handleAction = onTimeout
+                    , handleTimerManager = sysmgr
                     , handleKeyRef = keyref
                     , handleState = state
+                    , handleLastRenewed = lastRenewed
+                    , handleMinRenewGap = minRenewGap timeout
                     }
         pure h
+
+-- | How long 'tickle' waits before actually renewing the timeout:
+--   a quarter of the timeout, capped at one second. Skipping a renewal
+--   inside this window only shortens the effective timeout by up to
+--   this gap, but turns hot 'tickle' loops (one per chunk sent or
+--   received) into a clock read and a comparison.
+minRenewGap :: Int -> Word64
+minRenewGap timeout = min oneSecond (microToNano timeout `div` 4)
+  where
+    oneSecond = 1000000000
+    microToNano = (* 1000) . fromIntegral
 
 -- | Unregistering the timeout.
 cancel :: Handle -> IO ()
 cancel h@Handle{..} = withNonEmptyHandle h $ do
-    mgr <- getTimerManager
     key <- I.readIORef handleKeyRef
-    EV.unregisterTimeout mgr key
+    EV.unregisterTimeout handleTimerManager key
     I.atomicWriteIORef handleState Stopped
 
 -- | Extending the timeout.
 --
+-- To keep frequent callers cheap, the renewal is rate-limited: it is
+-- skipped unless at least a quarter of the timeout (capped at one
+-- second) has passed since the timeout was last registered or updated.
+--
 -- Careful: this does NOT reactivate an already paused 'Handle'!
 tickle :: Handle -> IO ()
 tickle h@Handle{..} = withNonEmptyHandle h $ do
-    mgr <- getTimerManager
-    key <- I.readIORef handleKeyRef
+    now <- getMonotonicTimeNSec
+    lastRenewed <- I.readIORef handleLastRenewed
+    when (now - lastRenewed >= handleMinRenewGap) $ do
+        key <- I.readIORef handleKeyRef
 #if defined(mingw32_HOST_OS)
-    EV.updateTimeout mgr key $ fromIntegral (handleTimeout `div` 1000000)
+        EV.updateTimeout handleTimerManager key $
+            fromIntegral (handleTimeout `div` 1000000)
 #else
-    EV.updateTimeout mgr key handleTimeout
+        EV.updateTimeout handleTimerManager key handleTimeout
 #endif
+        I.writeIORef handleLastRenewed now
 
 -- | This is identical to 'cancel'.
 --   To resume timeout with the same 'Handle', 'resume' MUST be called.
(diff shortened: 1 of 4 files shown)


time-manager/System/TimeManager.hs
```diff
@@ -126,39 +131,68 @@ register :: Manager -> TimeoutAction -> IO Handle
...
                     }
         pure h
 
+-- | How long 'tickle' waits before actually renewing the timeout:
+--   a quarter of the timeout, capped at one second. Skipping a renewal
+--   inside this window only shortens the effective timeout by up to
+--   this gap, but turns hot 'tickle' loops (one per chunk sent or
+--   received) into a clock read and a comparison.
+minRenewGap :: Int -> Word64
+minRenewGap timeout = min oneSecond (microToNano timeout `div` 4)
```
Vlix:
```hs
`div` 4 === `shiftR` 2
```

Vlix (changes requested):
Looks good and sounds like this might be a great improvement.

This PR does need to bump the `time-manager` minor version and add to the CHANGELOG to be accepted.

kazu-yamamoto (approved):
Sorry for the delay.
I was on a business trip.
This looks great to me.

kazu-yamamoto:
@Vlix We have a lot of PRs including yours at this moment.
I think it would be less stressful if you handled the merging yourself—what do you think?
I prefer a linear history, so if possible, I would like you to rebase onto `master` before merging.

Vlix:
> @Vlix We have a lot of PRs including yours at this moment. I think it would be less stressful if you handled the merging yourself—what do you think? I prefer a linear history, so if possible, I would like you to rebase onto `master` before merging.

My thoughts exactly.
I'll be away next week but will have a lot of time afterwards, so I'll start with my PR to get a `warp-3.4.15` release out first. Then I'll go through the #1091 - #1099 PRs and double check the benchmarks to see how the changes affect performance (and to see if I can replicate the speedups mentioned by @seanparsons, just to be sure)

kazu-yamamoto:
@seanparsons So, it would take time to merge your PRs. Please be patient.

seanparsons:
> @seanparsons So, it would take time to merge your PRs. Please be patient.

No worries, totally fine by me!

time-manager/System/TimeManager.hs
seanparsons: Made those changes in https://github.com/yesodweb/wai/pull/1097/commits/4aaca9c77a1de1fdc1b1edd7289d6d31f62b07c1

time-manager/System/TimeManager.hs
```diff
@@ -126,39 +132,68 @@ register :: Manager -> TimeoutAction -> IO Handle
...
+        -- process (and even if it were replaced, e.g. around a fork,
+        -- the key registered below would only be meaningful to the
+        -- manager it was registered with). So fetch it once here and
+        -- cache it in the 'Handle' instead of re-reading the global
+        -- IORef on every tickle/pause/resume.
         sysmgr <- getTimerManager
         key <- EV.registerTimeout sysmgr timeout onTimeout
         keyref <- I.newIORef key
         state <- I.newIORef Active
+        now <- getMonotonicTimeNSec
```
Vlix: Technically, this line should be right after `registerTimeout`.

time-manager/System/TimeManager.hs
```diff
@@ -126,39 +132,68 @@ register :: Manager -> TimeoutAction -> IO Handle
...
 
+-- | How long 'tickle' waits before actually renewing the timeout:
+--   a quarter of the timeout, capped at one second. Skipping a renewal
+--   inside this window only shortens the effective timeout by up to
+--   this gap, but turns hot 'tickle' loops (one per chunk sent or
+--   received) into a clock read and a comparison.
+minRenewGap :: Int -> Word64
+minRenewGap timeout = min oneSecond (microToNano timeout `shiftR` 2)
+  where
+    oneSecond = 1000000000
```
Vlix: Please use `NumericUnderscores` for readability

Vlix (changes requested):
Looks like a good change. 👍 
This can be merged and published after the requested changes are integrated.

time-manager/System/TimeManager.hs
seanparsons: Done in be4b626. `getMonotonicTimeNSec` now runs directly after `registerTimeout`, with the `IORef` allocations after it, so `handleLastRenewed` holds the registration time itself.

time-manager/System/TimeManager.hs
seanparsons: Done in be4b626. Added the pragma and applied it to `1_000_000_000` and `1_000` here, plus the `1_000_000` in `tickle` for consistency.

Vlix: Thank you, appreciated :+1:

Vlix (approved):

Vlix (changes requested):
We're going to attempt to improve the `pause` functionality with @seanparsons. Because if we can make pausing also lightweight, we can keep `resume/pause` semantics without the performance hit.

cf. #1098

Vlix:
Ok, I've written up some things about how it would look like if we wouldn't cancel on a pause, but instead just change the state and potentially let the timeout go off early.

- `register` creates a `Handle` (`registerTimeout`) in an `Active` state
- `tickle` delays the timeout (`updateTimeout`) if `Active`
- `pause` marks the `Handle` as `Paused`, but does not unregister it
- `resume` tickles the `Handle` if not `Stopped`, and marks it as `Active` if `Paused`
- `cancel` unregisters the timeout action (`unregisterTimeout`) and sets it to `Stopped`

Changing the `pause` behaviour can give some performance gains, but it also introduces race conditions. Because now, the registered action needs to check the state of the `Handle` and potentially reregister an action, which means also changing the state.

### No `pause` scenario

If `pause` wasn't used and the registered action runs, that means the timeout has elapsed, so if the registered action reads:

- a `Stopped` state, it can just discard the action, since a `cancel` has beat it in the race to reading/changing the state
- an `Active` state, it can just run the action, since the timeout has elapsed. (otherwise the action wouldn't have run)
- a `Paused` state, it should handle this as a "pause" scenario, since a `pause` has beat it in the race to reading/changing the state.

### `pause` scenario

If a `pause` _has_ been used, then this complicates things a bit. In this scenario:

- the registered action will run after `handleTimeout` amount of microseconds (delta `handleMinRenewGap`) from the last `pause`
- `tickle` doesn't do anything to a paused action
- a `cancel` winning the race with the `registered` action is either still on time, or it's too late, and then the registered action will act as a cancel since the state is still `Paused` anyway.
- this means the only race condition we can get is with a `resume`.

A `resume` winning the race means the registered action might read an `Active` state that was _just_ renewed, so it should not run it, but also not discard it. Since running it would be too soon, and discarding it would act as a `cancel`, even though the `Handle` was resumed.
This means that the cases where race conditions can occur are when `resume` reads a `Paused` or `Stopped` state, since `resume` changes the state to `Active` in those cases; and when the registered action reads a `Paused` or `Active` state, since it either changes the state to `Stopped` (`Paused` case) or the state might have been changed by `resume` (`Active` case).

---

## Side note

Currently, `resume` will restart a `cancel`led action.
We could also change this to be more "consistent", so that `resume` will restart a `paused` action, but that a `cancel` will stop the action from ever running, meaning you need to `register` a new one after a `cancel`. (This could be introduced as a `Cancelled` state)

This will give a more paired quality to the `resume/pause` combination and the `register/cancel` combination.

What do you think?

kazu-yamamoto:
I would like to divide this PR into two.
That is, a new PR , which introduces a cache of `TimeManager` in `Handle`, should be open and be merged first.

kazu-yamamoto:
@Vlix I like the side note approach. 
If I understand correctly, the `time-manager` becomes more lightweight, and we should simply focus on checking that `resume`/`pause` and `register`/`cancel` are used in matching pairs.

Vlix:
@kazu-yamamoto 
I have a branch with these changes waiting, and I'll make a new PR when it's done.

The goal is to make the `time-manager` not fire in hot loops, so that it doesn't reach out to the IO manager as much. And I've noticed that it would make the logic more straight-forward if we use `resume/pause` and `register/cancel` in pairs, since having `cancel` be resumable gives more chances for race conditions to keep in mind.

So if you think that would be an OK change (probably a major version bump, since the behaviour is changing) then I'd like to make that more of a rule in the code.

In short:
- `register` is used to start a timeout
- `cancel` will completely stop the timeout (cannot be `resume`d)
- `pause` will stop the timeout, but leave it `resumeable`
- `resume` will restart a timeout if it has been `paused` (or it will `tickle` if it's still active)
- `tickle` resets the timer of the timeout

kazu-yamamoto:
@Vlix Sounds awesome!

Vlix:
Merged via #1109

PR #1098, opened 2026-07-19
warp: fewer timer operations around streaming response bodies

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs. Works standalone (halves per-fragment timer ops); compounds with the time-manager tickle rate-limiting PR, which makes the per-fragment tickle nearly free.

`sendFragment` wrapped every streamed fragment in `T.resume`/`T.pause`: two timer-queue operations per fragment. Streaming now resumes once before the body, tickles per fragment, and pauses once after.

Deliberate semantics change for maintainers to weigh: The timeout is armed while user code computes between fragments, so a `responseStream` that produces nothing for a full timeout period is killed, previously it could stall forever between fragments, since the timeout was only armed during warp's own sends. Streams writing at least once per timeout period (or sending heartbeats) are unaffected, and the sends themselves are now covered, closing a slowloris-shaped gap. Observable for e.g. SSE without heartbeats. Verified manually that a stalled stream closes at the configured timeout.

### DIFF
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -257,6 +257,12 @@ sendRsp conn _ th ver s hs _ maxRspBufSize _ (RspBuilder body needsChunked) = do
 
 sendRsp conn _ th ver s hs _ _ _ (RspStream streamingBody needsChunked) = do
     header <- composeHeaderBuilder ver s hs needsChunked
+    -- Make sure the timeout is armed for the whole streaming section,
+    -- so that the per-fragment 'T.tickle' below has effect. Normally
+    -- the handle is already active (the respond callback resumed it),
+    -- making this a cheap no-op; it matters for streaming responses
+    -- sent from an exception handler, where the handle is still paused.
+    T.resume th
     (recv, finish) <-
         newByteStringBuilderRecv $
             reuseBufferStrategy $
@@ -408,14 +414,18 @@ sendRspFile404 conn ii th ver hs0 rspidxhdr maxRspBufSize method =
 -- | Use 'connSendAll' to send this data while respecting timeout rules.
 sendFragment :: Connection -> T.Handle -> ByteString -> IO ()
 sendFragment Connection{connSendAll = send} th bs = do
-    T.resume th
+    T.tickle th
     send bs
-    T.pause th
 
--- We pause timeouts before passing control back to user code. This ensures
--- that a timeout will only ever be executed when Warp is in control. We
--- also make sure to resume the timeout after the completion of user code
--- so that we can kill idle connections.
+-- The timeout stays armed during the whole streaming section (see the
+-- 'T.resume' in the RspStream case of 'sendRsp'); each fragment merely
+-- tickles it, which the rate limiter in time-manager makes nearly free.
+-- Compared to the previous resume/send/pause per fragment this keeps
+-- slowloris protection active while the fragment is being sent. The
+-- flip side is that user code producing the next fragment now runs with
+-- the timeout armed, so a streaming body that emits nothing for a full
+-- timeout period is killed, whereas it previously had the timeout
+-- paused between fragments.
 
 ----------------------------------------------------------------
 
--- a/warp/Network/Wai/Handler/Warp/Response.hs
+++ b/warp/Network/Wai/Handler/Warp/Response.hs
@@ -310,6 +310,12 @@ sendRsp conn _ th ver s hs rspidxhdr maxRspBufSize _ (RspBuilder body needsChunk
 
 sendRsp conn _ th ver s hs rspidxhdr _ _ (RspStream streamingBody needsChunked) = do
     (header, hdrLen) <- composeHeaderBuilder ver s hs rspidxhdr needsChunked
+    -- Make sure the timeout is armed for the whole streaming section,
+    -- so that the per-fragment 'T.tickle' below has effect. Normally
+    -- the handle is already active (the respond callback resumed it),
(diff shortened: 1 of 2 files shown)


Vlix (commented):
I don't like the change in behaviour.
We shouldn't act like a stream is a DOS/slow-loris if the user code is taking a long time.

Could this maybe work with a sort of "debounced" resume/pause function?

kazu-yamamoto:
> and pauses once after.

@seanparsons I don't understand which code this "pauses" refers to.
Could you tell me its filename and line number?

seanparsons:
> > and pauses once after.
> 
> @seanparsons I don't understand which code this "pauses" refers to. Could you tell me its filename and line number?

Ah yes, it's referring to this `pause` call higher up the call chain: https://github.com/yesodweb/wai/blob/82084792637783df0b53043b8089af8c6b2a6ce3/warp/Network/Wai/Handler/Warp/HTTP1.hs#L189

kazu-yamamoto:
Thanks. If we were to accept your changes, I think placing `resume` and `pause` close to each other would improve readability.

warp/Network/Wai/Handler/Warp/Response.hs
```diff
@@ -408,14 +414,18 @@ sendRspFile404 conn ii th ver hs0 rspidxhdr maxRspBufSize method =
 -- | Use 'connSendAll' to send this data while respecting timeout rules.
 sendFragment :: Connection -> T.Handle -> ByteString -> IO ()
 sendFragment Connection{connSendAll = send} th bs = do
-    T.resume th
+    T.tickle th
     send bs
-    T.pause th
```
Vlix, line 481: `resume == tickle` if the `th` hasn't been paused, so if we can find a way to _NOT_ pause on every send, but only if we're out of `sendFragment` for long enough (let's say half of the timeout), then we'll keep the behaviour we want in a `StreamingBody` response AND we'll not call `registerTimeout` so often. 🤔

Vlix (changes requested):
The problem, I think, is the `pause`. We don't need to pause immediately after being done sending a fragment, but if we debounce the pause, it will also have to get "cancelled" by the `resume/tickle`, which will complicate things.

If I can find the time I can try to make a rough sketch of what it could look like, but it will probably involve some STM and I'm not sure it will be worth it 🤔

seanparsons:
Rebased onto master, the conflict was just the `composeHeaderBuilder` signature change. I also added a streaming case to the response benchmark, since the per-fragment timer work was unmeasured until now. 64 fragments through `sendResponse` (one write plus flush each, the SSE shape), criterion medians of 3 alternating runs, GHC 9.10.3 -N4, stock time-manager:

```
resume/send/pause per fragment  118 us
tickle per fragment              76 us  (-36%)
```

@Vlix I have replied in your thread: I think you are right about the `pause`, and that the fix belongs in time-manager rather than in warp, in which case this PR should close rather than be reworked.

@kazu-yamamoto happy to move the `pause` next to the `resume` if we keep this shape, I have held off on that until the above is settled.

The four red CI jobs are not from this change: two are ghcup failing to install GHC with a 429, and two are the `time-manager` spec ("resume works as expected" and "resume also works as tickle") being killed by the timeout manager on loaded macOS and Windows runners. The warp suite passed on those same jobs, and this PR touches only `warp`.

warp/Network/Wai/Handler/Warp/Response.hs
seanparsons: I think you are right, and I think it means this PR is aimed at the wrong layer.

`pause` is `cancel`, so it drops the `GHC.Event` registration, which is why the next `resume` has to `registerTimeout` again. Both of those go through `editTimeouts` plus `wakeManager`. If `pause` instead only marked the handle as paused and left the registration in place, nothing in warp would need to change:

- `pause`: write `Paused` to `handleState`, no timer operation at all.
- `resume`: write `Active`, then `tickle`, which is `updateTimeout` on the key that is still registered. With the rate limiting in #1097 that is usually just a clock read.
- the action fires while the handle is `Paused`: do not run the timeout action, set the state to `Stopped` and let the registration go. The next `resume` sees `Stopped` and registers a fresh one.

That is the debounce you were after, except the thing being debounced is the deregistration rather than the pause itself, so nothing has to be cancelled by a later `resume`. It keeps the current semantics exactly, so the timeout still cannot fire while user code is producing the next fragment and a slow `StreamingBody` is not treated as a slowloris.

It also removes the timer operations from the normal request path, not just from streaming: `processRequest` pauses before the application runs and resumes in the respond callback (`HTTP1.hs` L189 and L196), which is an unregister plus a register on every single request.

The part that needs care is the race between a `resume` and a firing that is already in flight. If `resume` wins the state write, the in flight fire would see `Active` and kill a connection that has just come back. The `handleLastRenewed` timestamp from #1097 covers it: the fire path can compare it against the deadline and re-arm instead of firing when the handle was renewed recently. As a side effect that also makes the timeout precise again, rather than firing up to `min(1s, timeout/4)` early, which is the tradeoff #1097 currently asks you to accept.

Is that something you would want? It is roughly 25 lines in time-manager and it would fold naturally into #1097. If so then this PR should simply close, because warp would not need to change at all.

Vlix:
If we can make this robust in `time-manager`, so that you can declaratively say "pause/resume", and it still being as light-weight as possible, then that would be the dream.

The most important part of the `pause` update to the code would be that we handle race conditions correctly. This will need explicit tests and thinking through what the edge cases are.

So I'm in favor of keeping this PR "on hold" to see if we can move this fix over to `time-manager`.

Vlix:
Continuing this trail of thought in #1109

PR #1099, opened 2026-07-19
warp/recv: optimistic non-blocking recv before parking the thread

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs; self-contained (recv + warp). The single largest win of the series.

`makeGracefulRecv` (on the hot path of every `runSettings` server, since `ServerState` is always set up) registers with the IO manager and parks in STM on **every** receive, `waitReadSocketSTM` creates an event-manager registration up front, the composed `atomically` blocks until readable, and only then is `recv(2)` called. Under load the next keep-alive request is nearly always already in the kernel buffer, so every request paid one `epoll_ctl` plus a futex sleep/wake pair for nothing. Syscall census before: ~1 `epoll_ctl` and ~28 `futex` calls per request, with recv essentially never returning EAGAIN.

Changes: `recv` gains `receiveNoWait` (direct non-blocking `recv(2)` on the already-SOCK_NONBLOCK fd; `tryWithBufferPool` keeps the pool intact on decline; EAGAIN and real errors both defer to the blocking path so exceptions keep their errno; `Just ""` is EOF; Windows always returns `Nothing` = unchanged behavior). `makeGracefulRecv` tries it first; the slow path is byte-for-byte the previous implementation. One edge: a request already in the kernel buffer is served even if shutdown began meanwhile restoring pre-graceful-shutdown behavior for bytes the client already sent. The `Conc.yield` after each response turns out to be load-bearing with this fast path (it lets the next request arrive; removing it costs ~17%), comment updated.

After: 186 `epoll_ctl` and 725 `futex` calls **total** across 198k requests; the per-request syscall profile is one recvfrom + one sendto. Measured: +36% on top of the rest of the series (52.8k → 71.7k req/s; stock baseline 35.7k), −N1 unpinned 67.6k → 91.5k. Also validated on a production Servant app: +57-80% on light endpoints, byte-identical responses. Spec suite passes.

Note: `receiveNoWait` uses its own FFI `recv(2)` because `Network.Socket.recvBufNoWait` isn't exported from `network`'s public API; happy to switch if exporting it upstream is preferred.

### DIFF
--- a/recv/Network/Socket/BufferPool/Buffer.hs
+++ b/recv/Network/Socket/BufferPool/Buffer.hs
@@ -1,6 +1,7 @@
 module Network.Socket.BufferPool.Buffer (
     newBufferPool,
     withBufferPool,
+    tryWithBufferPool,
     mallocBS,
     copy,
 ) where
@@ -43,6 +44,26 @@ withBufferPool (BufferPool l h ref) f = do
     consumed <- withForeignBuffer buf f
     writeIORef ref $ unsafeDrop consumed buf
     return $ unsafeTake consumed buf
+
+-- | Like 'withBufferPool' for fillers that can decline to fill:
+--   a negative return value from the filler leaves the pool untouched
+--   and produces 'Nothing'.
+tryWithBufferPool
+    :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO (Maybe ByteString)
+tryWithBufferPool (BufferPool l h ref) f = do
+    buf0 <- readIORef ref
+    buf <-
+        if BS.length buf0 >= l
+            then return buf0
+            else mallocBS h
+    consumed <- withForeignBuffer buf f
+    if consumed < 0
+        then do
+            writeIORef ref buf
+            return Nothing
+        else do
+            writeIORef ref $ unsafeDrop consumed buf
+            return $ Just $ unsafeTake consumed buf
 
 withForeignBuffer :: ByteString -> (Buffer -> BufSize -> IO Int) -> IO Int
 withForeignBuffer (PS ps s l) f = withForeignPtr ps $ \p -> f (castPtr p `plusPtr` s) l
--- a/recv/Network/Socket/BufferPool/Buffer.hs
+++ b/recv/Network/Socket/BufferPool/Buffer.hs
@@ -10,6 +10,7 @@ import qualified Data.ByteString as BS
 import Data.ByteString.Internal (ByteString (..))
 import Data.ByteString.Unsafe (unsafeDrop, unsafeTake)
 import Data.IORef (newIORef, readIORef, writeIORef)
+import Data.Maybe (fromMaybe)
 import Foreign.ForeignPtr
 import Foreign.Marshal.Alloc (finalizerFree, mallocBytes)
 import Foreign.Marshal.Utils (copyBytes)
@@ -35,19 +36,11 @@ newBufferPool l h = BufferPool l h <$> newIORef BS.empty
 --   how many bytes are filled in the buffer.
 --   The buffer in the buffer pool is automatically managed.
 withBufferPool :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO ByteString
-withBufferPool (BufferPool l h ref) f = do
-    buf0 <- readIORef ref
-    buf <-
-        if BS.length buf0 >= l
-            then return buf0
-            else mallocBS h
-    consumed <- withForeignBuffer buf f
-    writeIORef ref $ unsafeDrop consumed buf
-    return $ unsafeTake consumed buf
+withBufferPool pool f = fromMaybe BS.empty <$> tryWithBufferPool pool f
 
--- | Like 'withBufferPool' for fillers that can decline to fill:
---   a negative return value from the filler leaves the pool untouched
---   and produces 'Nothing'.
+-- | Like 'withBufferPool', but for fillers that can decline to fill the
+--   buffer: a negative return value from the filler leaves the buffer pool
+--   untouched and produces 'Nothing'.
 tryWithBufferPool
     :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO (Maybe ByteString)
 tryWithBufferPool (BufferPool l h ref) f = do
--- a/recv/Network/Socket/BufferPool/Recv.hs
+++ b/recv/Network/Socket/BufferPool/Recv.hs
@@ -1,7 +1,9 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 module Network.Socket.BufferPool.Recv (
     receive,
+    receiveNoWait,
     makeRecvN,
 ) where
 
@@ -9,6 +11,12 @@ import qualified Data.ByteString as BS
 import Data.ByteString.Internal (ByteString (..), unsafeCreate)
 import Data.IORef
 import Network.Socket (Socket, recvBuf)
+#ifndef mingw32_HOST_OS
+import Foreign.C.Types (CInt (..), CSize (..))
+import Foreign.Ptr (Ptr, castPtr)
+import Network.Socket (withFdSocket)
+import System.Posix.Types (CSsize (..))
+#endif
 
 import Network.Socket.BufferPool.Buffer
 import Network.Socket.BufferPool.Types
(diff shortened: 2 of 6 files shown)


warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -144,22 +144,33 @@ socketConnection set s = do
...
+    -- Fast path: under load the next request has usually already arrived,
+    -- so read it without registering with the IO manager or entering STM.
+    -- One epoll_ctl + one futex sleep/wake pair saved per request.
+    -- A request already in the kernel buffer is served even if shutdown
+    -- began meanwhile, which merely restores pre-graceful-shutdown behavior
+    -- for bytes the client already sent.
+    mbs <- receiveNoWait sock pool
+    case mbs of
+        Just bs -> return bs
+        Nothing -> slowPath
```
Vlix: Doesn't this mean that we might never check the shutting down state if the client is prompt with their bytes and we happen to only `recv` when there's bytes in the kernel buffer?

recv/Network/Socket/BufferPool/Buffer.hs
```diff
@@ -44,6 +45,26 @@ withBufferPool (BufferPool l h ref) f = do
     writeIORef ref $ unsafeDrop consumed buf
     return $ unsafeTake consumed buf
 
+-- | Like 'withBufferPool' for fillers that can decline to fill:
+--   a negative return value from the filler leaves the pool untouched
+--   and produces 'Nothing'.
+tryWithBufferPool
+    :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO (Maybe ByteString)
+tryWithBufferPool (BufferPool l h ref) f = do
```
Vlix, line 46: I don't like the duplication of code here, though we might have to.

This will also need tests like 'withBufferPool', though, if it IS duplicated.
And it will need tests either way, even if we refactor `withBufferPool` for this new functionality.

recv/Network/Socket/BufferPool/Recv.hs
```diff
@@ -20,6 +28,27 @@ import Network.Socket.BufferPool.Types
...
+    withFdSocket sock $ \fd -> do
+        -- The socket is non-blocking, so an unsafe foreign call is fine:
+        -- recv(2) returns immediately either way. Both EAGAIN and real
+        -- errors map to a negative result, deferring to the blocking path
+        -- so errors surface there with their errno intact.
+        n <- c_recv fd (castPtr ptr) (fromIntegral size) 0
+        return $ fromIntegral n
+
+foreign import ccall unsafe "recv"
+    c_recv :: CInt -> Ptr CSsize -> CSize -> CInt -> IO CSsize
```
Vlix, line 47: We basically want to use `recvBufNoWait` from `Network.Socket.Buffer`, but it's not exported.

kazu-yamamoto (approved):
I support this change because it restores an optimization that was lost when transitioning to the *event-poll* model. The `recv` function in the `network` library is actually implemented this way: it first attempts a non-blocking `c_recv` and only invokes the I/O manager if it encounters `EAGAIN`.

kazu-yamamoto:
@Vlix Compared to other PRs, I think this one is clear and easy to merge.

mkleczek:
> One edge: a request already in the kernel buffer is served even if shutdown began meanwhile restoring pre-graceful-shutdown behavior for bytes the client already sent.

So under load `warp` won't stop serving incoming requests? It looks to me it breaks graceful shutdowns - that was exactly the problem described in #853

Vlix (changes requested):
This might break the graceful shutdown. We need to know the extent of this change before we can accept it.

recv/Network/Socket/BufferPool/Recv.hs
kazu-yamamoto: If necessary, I can export it from `network`.

Vlix: I think that in itself is a good idea. It's nice for the user to be able to decide to block or not block on a `recv`.

Vlix:
We can't accept this PR if it breaks graceful shutdown, so the `makeGracefulRecv` should have a `checkShutdown` before it continues reading / tries to read from the request.

kazu-yamamoto:
I don't think we should give up on the optimization that uses the fast path.

A new fast path: how about checking the graceful shutdown status, and attempting a non-blocking `recv` if a graceful shutdown is not in progress? We might also want to check whether the pool is empty.

warp/Network/Wai/Handler/Warp/HTTP1.hs
```diff
@@ -214,13 +214,10 @@ processRequest settings ii conn app th istatus src req mremainingRef idxhdr next
...
-    -- it is likely to fail and cause the IO manager to do some work.
-    -- It is very costly, so we yield to another Haskell
-    -- thread hoping that the next Request will arrive
-    -- when this Haskell thread will be re-scheduled.
-    -- This improves performance at least when
-    -- the number of cores is small.
+    -- receive a Request again. Yield to other Haskell threads so the
+    -- next Request has a chance to arrive before we call recv: the
+    -- non-blocking fast path in 'makeGracefulRecv' then reads it without
+    -- involving the IO manager at all.
```
mkleczek: I am not sure it is a good idea to change only comment - there are no changes to the code.

I don't find the new comment any better than the old one, nevertheless - if it is about improving comments, it should be a different PR, shouldn't it?

recv/Network/Socket/BufferPool/Recv.hs
Vlix: @kazu-yamamoto 
Can you make it so `recvBufNoWait` is exposed from the `network` package? I think it's a good idea in general.

recv/Network/Socket/BufferPool/Buffer.hs
seanparsons: Deduplicated in [537e2e8](https://github.com/yesodweb/wai/pull/1099/commits/537e2e8e5cbbd0d35d139552e842c120ccab7929), `withBufferPool` is now `fromMaybe BS.empty <$> tryWithBufferPool pool f`. Tests for `tryWithBufferPool` in the same commit, including one I mutation-checked (dropping the buffer on decline fails it).

warp/Network/Wai/Handler/Warp/HTTP1.hs
seanparsons: Reverted in [537e2e8](https://github.com/yesodweb/wai/pull/1099/commits/537e2e8e5cbbd0d35d139552e842c120ccab7929), HTTP1.hs is out of this PR now.

recv/Network/Socket/BufferPool/Recv.hs
seanparsons: Worth more than deleting the stub here: `recvBufNoWait` in network-3.2.8.0 already has a Windows implementation, and returns -1 for EAGAIN and -2 otherwise, which is the convention `tryWithBufferPool` already expects. Drop-in, and it'd turn the fast path on for Windows too, where `receiveNoWait` currently just returns `Nothing`. Happy to swap once it's released.

warp/Network/Wai/Handler/Warp/Run.hs
seanparsons: Yes, that was a real hole. Fixed in [537e2e8](https://github.com/yesodweb/wai/pull/1099/commits/537e2e8e5cbbd0d35d139552e842c120ccab7929), taking @kazu-yamamoto's suggestion: the shutdown flag is read before the fast path.

Reading just the flag is enough. `checkShutdown` is `flag && apps <= 0`, so it can't fire while the flag is unset, which means the old code would have waited and read anyway. Flag set hands over to the slow path, which still waits for in-flight apps. Same ordering as master, same window.

On the extent: `sendResponse` already does `shouldPersist = not isShuttingDown && ...` (Response.hs:122-129), so once a shutdown is visible at response time warp sends `Connection: close` and the loop never calls recv again. Worst case before was one extra request per connection, in the gap between that check and the recv. Not unbounded, but still wrong. Now it's none.

Worth knowing if anyone writes this differently: `atomically $ (checkShutdown $> True) <|> pure False` costs about 8%. `readTVarIO` costs nothing.

Interleaved A/B, hello world, `wrk -t4 -c100`, medians at -N1/-N4: master 79.8k/36.5k, this PR 112.0k/38.8k. Guard off vs on at -N1 is 113.8k vs 111.3k, sign flips between rounds. The +36% in the description was on top of the rest of the series; on its own off master it's +40% at -N1 and +6% at -N4, because the timer manager still dominates at -N4 (that's #1097/#1098).

No new test for the window itself: it needs the flag to flip between two steps of the request loop and there's no hook for that. `GracefulShutdownSpec` passes with and without the fix, so it proves nothing here.

Vlix:
We're continuing the goal of this PR in #1107 

@seanparsons If you feel you should be credited as Co-Author, please say so, otherwise we'll leave you out of the commits as we don't want to include someone without their acknowledgement.

seanparsons:
> @seanparsons If you feel you should be credited as Co-Author, please say so, otherwise we'll leave you out of the commits as we don't want to include someone without their acknowledgement.

Happy to, but don't go out of your way to do so.

PR #1100, opened 2026-07-19
warp: track in-flight apps with an IORef instead of STM (optional)

seanparsons:
Part of a series splitting #1090 into independently reviewable PRs. **Offered as optional** — measured performance-neutral on its own; take it or leave it.

Every request pays two STM transactions to bump the graceful-shutdown in-flight counter, even when no shutdown ever happens. This makes `connAppsInProgress` an `IORef Int` (atomicModifyIORef', exact); the shutdown-side wait polls at 20ms (sleeping inside `timeout` around the socket-readiness STM wait, no busy loop) while still feeding in-progress streams. Not-shutting-down path unchanged; warp-tls updated in lockstep.

Honest numbers: no measurable throughput or allocation change on a keep-alive hello-world load test — the value is a simpler hot path, at the cost of an `Internal` API change (`TVar Int` → `IORef Int` on `Connection.connAppsInProgress` and `makeGracefulRecv`). If the API churn isn't worth it, closing this PR affects nothing else in the series.

Interaction note: if the benchmark PR merges first, this needs a one-line rebase in the bench's `Connection` construction. GracefulShutdown spec passes; full spec suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011HBrMGfWRxTpYeEB4UT8Kf

### DIFF
--- a/warp-tls/Network/Wai/Handler/WarpTLS.hs
+++ b/warp-tls/Network/Wai/Handler/WarpTLS.hs
@@ -58,9 +58,6 @@ module Network.Wai.Handler.WarpTLS (
 ) where
 
 import Control.Applicative ((<|>))
-#if MIN_VERSION_warp(3,4,13)
-import Control.Concurrent.STM (newTVarIO, TVar)
-#endif
 import Control.Exception (
     Exception,
     IOException,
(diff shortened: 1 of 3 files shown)


seanparsons:
Withdrawing this one: it was the measured-neutral optional change in the series (no throughput, latency, or allocation difference at any tested concurrency), and dropping it keeps the graceful-shutdown wait on STM's blocking retry instead of a polling loop. The remaining PRs #1091-#1099 are unaffected.

PR #1103, opened 2026-08-06
Use atomicWriteIORef when clearing value in auto-update

### LINKED ISSUE #1102: Possible stale values returned by auto-update due to lack of memory barrier
We have some intermittent and hard to pinpoint issues when using auto-update with `getCurrentTime`. It looks like the action sometimes returns values far in the past.

I've analyzed auto-update code and have a suspicion it _might_ be caused by this line:
https://github.com/yesodweb/wai/blob/4903427b48adf0244b6957e745b1f8053fa02a06/auto-update/Control/AutoUpdate/Thread.hs#L111

`writeIORef` should be replaced with `atomicWriteIORef` here because the lack of memory barrier means some threads might always get `Right a` from `readIORef`.
That would explain the issue we see - some threads see stale value.

mkleczek:
Fixes #1102

### DIFF
--- a/auto-update/Control/AutoUpdate/Thread.hs
+++ b/auto-update/Control/AutoUpdate/Thread.hs
@@ -20,7 +20,7 @@ import Control.Exception (
     try,
  )
 import Control.Monad (void)
-import Data.IORef (newIORef, readIORef, writeIORef)
+import Data.IORef (newIORef, readIORef, writeIORef, atomicWriteIORef)
 import Data.Maybe (fromMaybe)
 import GHC.Conc.Sync (labelThread)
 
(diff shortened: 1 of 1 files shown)


auto-update/Control/AutoUpdate/Thread.hs
Vlix: Let's use `atomicWriteIORef` here as well, to keep it symmetric.

Vlix (changes requested):

auto-update/Control/AutoUpdate/Thread.hs
mkleczek: > Let's use `atomicWriteIORef` here as well, to keep it symmetric.

Good catch, of course! Done.

Vlix:
Looks good, will merge after GitHub actions start cooperating on the big PR I'm working through.

Vlix:
published as [`auto-update-0.2.7`](https://hackage.haskell.org/package/auto-update)

Issue #979, opened 2024-03-08
Improve Documentation of Graceful Shutdown


andrewthad:
I spent a few hours recently figuring out what the how graceful shutdown behaves. I believe that I now mostly understand what is happening, and I am writing this for these purposes:

* to help other users who are trying to figure out how graceful shutdown works
* to suggest improvements to the documentation of graceful shutdown
* to suggest improvements to the behavior of graceful shutdown

# Closing Sockets

The documentation of warp suggests closing (i.e. `Network.Socket.close`) the socket to initiate graceful shutdown. On UNIX systems, a cross-thread `close()` on a socket like this is dangerous because the kernel could hand out the same file descriptor as a result to an `open()` or `socket()` call. But with the `network` library, this race condition cannot happen because the implementation atomically swaps out the file descriptor with negative one (-1) before the `close()` syscall. Which means that subsequent use of the socket is not merely likely to return EBADF, it is certain to return EBADF (on Linux at least, -1 is never a valid file descriptor). To someone who is accustomed to UNIX, this improved guarantee is not obvious. It would be nice to mention in the warp docs that network's implementation of `close()` doesn't have the dangerous race condition that UNIX sockets have. Even the documentation of `Network.Socket.close` itself is deficient here:

> Close the socket. This function does not throw exceptions even if the underlying system call returns errors.
>
> If multiple threads use the same socket and one uses unsafeFdSocket and the other use close, unexpected behavior may happen. For more information, please refer to the documentation of unsafeFdSocket.

This could be improved by mentioning that any calls to `send`, `receive`, `accept`, etc. will certainly throw exceptions after the socket is closed.

A question I have: Does the user have to `close` the listening socket, or it is sufficient to call `shutdown` on it instead? Warp's source code leaves me with the impression that `shutdown` should work, but I'm not certain. If calling shutdown is supported, it would be good to document this.

# Ending HTTP Sessions

## My Specific Scenario

In the application that I am working on, a load balancer (HAProxy) sits in front of the application and runs in HTTP mode. For anyone unfamiliar with HTTP-mode load balancing, this means that the proxy opens a small number (possibly just 1) of TCP sessions to the backend and leaves these TCP sessions open forever. Any request that a client (on the frontend) makes to the proxy is handled by forwarding (basically copying) this request to one of these existing backend TCP sessions. Running a load balancer like this (instead of in TCP mode) is considered good practice and is pretty common for HTTP traffic.

## Graceful Shutdown Behavior

[...]

PR #1104, opened 2026-08-16
warp: count a connection when it is accepted, not when its thread starts

NorfairKing:
Warp returns from `runSettingsSocket` while a connection it has accepted has neither been served nor closed. Graceful shutdown is meant to wait for connections and does not wait for this one, because it is not in the counter yet. This raises the counter in the accept loop rather than in the connection's own thread, which closes the window.

This is one of two shapes of the same fix. The other is #1105, which leaves `getCount` alone and gives graceful shutdown its own counter. This one is the smaller diff but shifts slightly what `getCount` and `currentOpenConnections` report, described under "Behaviour that shifts with this change" below. **Only one of the two should be merged**, and if you have no preference I would take the other one.

## Minimal repro

```haskell
{-# LANGUAGE RankNTypes #-}

import Control.Concurrent (forkIOWithUnmask, threadDelay)
import Control.Exception (bracket)
import Control.Monad (void)
import Data.IORef
import Network.Socket

main :: IO ()
main = do
    accepted <- newIORef (0 :: Int)
    closed <- newIORef (0 :: Int)

    -- The gap between accepting a connection and the thread serving it being
    -- scheduled.  Delayed here to make it a certainty rather than a race; in a
    -- server it is however long the runtime takes to get to the new thread.
    let slowFork :: ((forall a. IO a -> IO a) -> IO ()) -> IO ()
        slowFork act = void $ forkIOWithUnmask $ \unmask ->
            threadDelay 200_000 >> act unmask

        -- Take one connection and then stop accepting.  That is what closing
        -- the listening socket does to the accept loop, without closing a
        -- descriptor the accept loop is parked on.
        acceptOnce sock = do
            taken <- atomicModifyIORef' accepted $ \n -> (n + 1, n)
            if taken == 0
                then accept sock
                else ioError (userError "the listening socket is closed")

        settings =
            setFork slowFork $
                setAccept acceptOnce $
                    setOnClose (\_ -> atomicModifyIORef' closed $ \n -> (n + 1, ())) $
                            setOnException (\_ _ -> pure ()) defaultSettings

        app _ respond = respond $ responseLBS status200 [("Content-Length", "0")] ""

    bracket openFreePort (close . snd) $ \(port, sock) -> do
        bracket (socket AF_INET Stream defaultProtocol) close $ \client ->
            connect client $
                SockAddrInet (fromIntegral port) (tupleToHostAddress (127, 0, 0, 1))

        runSettingsSocket settings sock app

        -- runSettingsSocket has returned.  Warp has stopped its timeout
        -- manager and its caches, and a server's next step is to exit, so
        -- whatever is unfinished here stays unfinished.
        readIORef accepted >>= \n -> putStrLn ("accepted:       " ++ show (min 1 n))
        readIORef closed >>= \n -> putStrLn ("closed by warp: " ++ show n)
```

On master:

```
accepted:       1
closed by warp: 0
```

With this change:

```
accepted:       1
closed by warp: 1
```

Warp accepted a connection and then returned without finishing with it, having stopped the timeout manager and the caches that serving it would need. Whatever the caller does next, that connection is over.

## Why

`fork` hands the connection to `settingsFork` and the accept loop goes straight back to `accept`, but the counter is raised by `onOpen`, which runs in the forked thread. From `accept` returning until that thread is scheduled the connection exists and is not counted. `gracefulShutdown` then runs `waitForZero` on that counter: a connection accepted in that gap reads as zero, so warp stops.

The repro widens the gap on purpose. Left alone it is however long the runtime takes to schedule a thread, which is short, but it is reached constantly where the listening socket outlives the process, as under systemd socket activation: connections keep being accepted right up to the close, and the process exits immediately afterwards. A service of mine dropped six requests across twenty restarts when probed with sixteen concurrent clients, showing up as `curl: (52) Empty reply from server` and one connection reset. With this change, the same probe over the same number of restarts drops none.

## The change

`increase` moves into `fork` before `settingsFork`, so a connection is counted by the accept loop the moment `accept` hands it over.

`decrease` moves into a `finally` around the whole thread body rather than `onClose`, so every counted connection is uncounted exactly once however it ends. `onClose` would miss any connection whose `mkConn` throws before it, which under warp-tls is a failed handshake.

The thread body moves into a `where` binding unchanged, so the diff there is indentation only. `git diff -w` on `Run.hs` shows the change at its real size.

## Cost on the accept loop

This puts one STM transaction on the accept loop that used to happen in the connection's own thread, so it is fair to ask what the loop pays for it. I could not measure a difference: 4000 short-lived connections, one accept each, twelve interleaved runs of each build.

|  | best | median |
|---|---|---|
| master | 0.154s | 0.180s |
| this change | 0.151s | 0.180s |

That is around 20-26k accepts/s either way, with the two distributions fully overlapping. An uncontended `modifyTVar'` against an `accept()` syscall and a `forkIO` is not where the time goes.

## What this does and does not fix

It stops warp abandoning a connection it accepted. It does not by itself guarantee that connection's request is answered, and I would rather say so than let the repro imply otherwise.

Once shutdown has begun, `makeGracefulRecv` gives a synthetic EOF to any connection with no `Application` in progress, which is the 3.4.13 behaviour of not starting new work on idle keep-alive connections. A connection accepted moments before the close has no `Application` in progress yet, so if its thread only reaches `connRecv` after `setShuttingDown`, it is closed rather than served, even with this change. In practice the thread gets there first, which is why the probe above goes to zero drops; the repro loses that race on purpose with its 200ms delay, which is why it asserts the connection was closed rather than answered.

Whether a connection carrying an unread request should count as idle looks like a separate question, and I have not touched it. It belongs with #979, which asked for the behaviour that became `makeGracefulRecv` and lists "some HTTP requests may get a connection closed instead of a response" among the things still unsatisfying about shutdown. This PR removes one cause of that; the idle test is another.

## Behaviour that shifts with this change

`waitForDecreased`, the `EMFILE` backoff from #1084, no longer reports `NoConnections` while warp holds a connection it has accepted but not begun.

`getCount` and `currentOpenConnections` count from `accept` rather than from the connection thread starting, so a connection appears in them slightly earlier, and one whose `mkConn` throws appears at all. That looks closer to what "open connections" means, since warp is holding the socket either way, but it is a visible change to a public API: say if you would rather the exposed counter kept its old meaning and graceful shutdown waited on a separate one.

## Test

`GracefulShutdownSpec` gets the repro as a test case.

Master's suite is 121 examples and 0 failures; with this branch it is 122 and 0. Reverting just the change to `Run.hs` and keeping the new test gives 122 examples and exactly 1 failure, the new one:

```
waits for a connection accepted just before it stopped accepting [✘]
  expected: 1
   but got: 0
```

## Checklist

Before submitting:

- [x] Bumped the version number, to 3.4.15.1 as a bug fix

After submitting:

- [x] Update the Changelog.md file with a link to this PR
- [x] Check that CI passes

## P.S. two things I decided rather than assumed

A `setFork` that never runs the action it is handed would now leave the count raised, and graceful shutdown would wait out its timeout instead of returning at once. Before this change such a fork went uncounted entirely. Both are broken in their own way and I do not think a fork that drops the connection is worth designing around, but it is a difference.

`increase` sits outside any bracket. If `settingsFork` itself threw before forking, the count would stay raised, and I first wrote an `onException` to cover that. It cannot matter: `acceptConnection` is `E.mask_ acceptLoop` and only then `gracefulShutdown`, so an exception out of `fork` skips the only code that reads the counter. I dropped the guard rather than carry something that protects nothing, but I would rather say I thought about it than have it look overlooked.

## Mergeable with #1106

#1106 changes how `accept()` failures are handled, and the test here has to stop the accept loop somehow, so the two interact. The test now does it with `errnoToIOError "accept" eBADF`, which is what a closed listening socket actually gives and what #1106 lets through quietly. Merging both branches and running the suite gives 125 examples and 0 failures.

The only conflict between them is the `## 3.4.15.1` heading each adds to the ChangeLog, which is a one-line resolution.

### DIFF
--- a/warp/test/GracefulShutdownSpec.hs
+++ b/warp/test/GracefulShutdownSpec.hs
@@ -10,6 +10,7 @@ import Control.Concurrent.Async
 import Control.Exception (bracket)
 import Control.Monad (void)
 import Data.IORef
+import Foreign.C.Error (eBADF, errnoToIOError)
 import Network.HTTP.Client
 import Network.HTTP.Types (ok200, status200)
 import Network.Socket
@@ -33,14 +34,15 @@ spec = describe "graceful shutdown" $ do
                 threadDelay 200_000
                 act unmask
 
-            -- Take one connection, then stop accepting, which is what
-            -- closing the listening socket does to the accept loop without
-            -- closing a descriptor it is parked on.
+            -- Take one connection, then stop accepting, with the error a
+            -- closed listening socket actually gives.  Closing it for real
+            -- would mean closing a descriptor the accept loop is parked on,
+            -- which the IO manager does not survive cleanly.
             acceptOnlyOne sock = do
                 taken <- atomicModifyIORef' accepted $ \n -> (n + 1, n)
                 if taken == 0
                     then accept sock
-                    else ioError (userError "the listening socket is closed")
+                    else ioError (errnoToIOError "accept" eBADF Nothing Nothing)
 
             settings =
                 setFork slowFork $
--- a/warp/test/GracefulShutdownSpec.hs
+++ b/warp/test/GracefulShutdownSpec.hs
@@ -54,8 +54,14 @@ spec = describe "graceful shutdown" $ do
             app _ respond = respond $ responseLBS status200 [("Content-Length", "0")] ""
 
         bracket openFreePort (close . snd) $ \(testPort, sock) -> do
-            -- Connect before the server starts, so the accept loop has a
-            -- connection waiting for it whatever else the machine is doing.
+            -- Connect before the server exists. openFreePort has already put
+            -- the socket in listen state, so this lands in its accept queue
+            -- in the kernel and stays there: closing the client end sends a
+            -- FIN but does not take it off the queue, and accept() still
+            -- hands it over. Queueing it up front is what makes the accept
+            -- loop's first accept() return immediately, rather than racing a
+            -- client connecting alongside it, which on a loaded machine it
+            -- can lose.
             bracket (openConnection testPort) close $ \_ -> pure ()
 
             withAsync (runSettingsSocket settings sock app) $ \server -> do
--- a/warp/test/GracefulShutdownSpec.hs
+++ b/warp/test/GracefulShutdownSpec.hs
@@ -10,7 +10,6 @@ import Control.Concurrent.Async
 import Control.Exception (bracket)
 import Control.Monad (void)
 import Data.IORef
-import Foreign.C.Error (eBADF, errnoToIOError)
 import Network.HTTP.Client
 import Network.HTTP.Types (ok200, status200)
 import Network.Socket
@@ -34,15 +33,15 @@ spec = describe "graceful shutdown" $ do
                 threadDelay 200_000
                 act unmask
 
-            -- Take one connection, then stop accepting, with the error a
-            -- closed listening socket actually gives.  Closing it for real
-            -- would mean closing a descriptor the accept loop is parked on,
-            -- which the IO manager does not survive cleanly.
+            -- Take one connection, then stop accepting by closing the
+            -- listening socket, which is what a graceful shutdown does. The
+            -- close happens here rather than from another thread so that it
+            -- cannot land while the accept loop is parked inside accept().
             acceptOnlyOne sock = do
                 taken <- atomicModifyIORef' accepted $ \n -> (n + 1, n)
                 if taken == 0
                     then accept sock
-                    else ioError (errnoToIOError "accept" eBADF Nothing Nothing)
+                    else close sock >> accept sock
 
             settings =
                 setFork slowFork $
--- a/warp/test/GracefulShutdownSpec.hs
+++ b/warp/test/GracefulShutdownSpec.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 
 module GracefulShutdownSpec (spec) where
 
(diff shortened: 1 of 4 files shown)


NorfairKing:
CI here needs a maintainer to approve the run, so I ran your workflow unchanged in my fork against the same commit (3ab36712): all 18 cells green, 6 resolvers from lts-20 through nightly on ubuntu, macos and windows. https://github.com/NorfairKing/wai/actions/runs/31946791122

Vlix:
This seems like a decent way to solve this issue. I'll take a look at #1105 now.

EDIT: Yes, I prefer this solution. `currentOpenConnections` is meant to show all accepted connections, so this doesn't change that. We could maybe improve the documentation or name of the function reading the counter, but in essence its role is unchanged.

kazu-yamamoto:
@NorfairKing Thank you for this report. Nice catch!

@Vlix If you will improve the documentation or name of the function, I support your decision.

warp/test/GracefulShutdownSpec.hs
```diff
@@ -1,23 +1,71 @@
...
+                        setOnClose (\_ -> atomicModifyIORef' closed $ \n -> (n + 1, ())) $
+                            setGracefulShutdownTimeout (Just 5) $
+                                setOnException (\_ _ -> pure ()) defaultSettings
+
+            app _ respond = respond $ responseLBS status200 [("Content-Length", "0")] ""
+
+        bracket openFreePort (close . snd) $ \(testPort, sock) -> do
+            -- Connect before the server starts, so the accept loop has a
+            -- connection waiting for it whatever else the machine is doing.
+            bracket (openConnection testPort) close $ \_ -> pure ()
```
Vlix, line 64: @NorfairKing 
Does this do anything? It seems to open a connection, but then immediately closes it? And the async only happens after the close?

NorfairKing: Yes. Take that line out and the test hangs instead of failing, because `accept()` blocks with nothing queued.

`openFreePort` is `bindRandomPortTCP`, so the socket is already in listen state before any of this runs. The `connect` completes against that socket's accept queue in the kernel, and the connection sits there whether or not warp exists yet. Closing the client end sends a FIN but does not take it off the queue, and `accept()` still hands it over. That is what the test wants: it is about what warp does with a connection it has accepted, not about what the client does with it.

Doing it before the server starts is for determinism. The accept loop's first `accept()` then returns immediately with something waiting, instead of racing a client connecting alongside it. On a loaded machine that race can go the other way, the loop reaches the `EBADF` first, and the test stops exercising the window it exists for.

You are the second person to read that line and stop, so I have expanded the comment rather than leave it at "connect before the server starts":

```haskell
-- Connect before the server exists. openFreePort has already put
-- the socket in listen state, so this lands in its accept queue
-- in the kernel and stays there: closing the client end sends a
-- FIN but does not take it off the queue, and accept() still
-- hands it over. Queueing it up front is what makes the accept
-- loop's first accept() return immediately, rather than racing a
-- client connecting alongside it, which on a loaded machine it
-- can lose.
```

warp/test/GracefulShutdownSpec.hs
```diff
@@ -1,23 +1,71 @@
...
+
+            -- Take one connection, then stop accepting, with the error a
+            -- closed listening socket actually gives.  Closing it for real
+            -- would mean closing a descriptor the accept loop is parked on,
+            -- which the IO manager does not survive cleanly.
+            acceptOnlyOne sock = do
+                taken <- atomicModifyIORef' accepted $ \n -> (n + 1, n)
+                if taken == 0
+                    then accept sock
+                    else ioError (errnoToIOError "accept" eBADF Nothing Nothing)
```
Vlix: This is emulating accepting a closed socket, right?
Can't we close the socket to get an `eBADF` naturally? That would also show a failing test on Windows, which would help in fixing the exception issue once we get better exceptions on Windows.

NorfairKing: Yes, great idea!

Vlix (approved):
NOTE TO SELF: adjust documentation of `currentOpenConnections` before releasing `warp-3.4.16`

NorfairKing:
@Vlix is there anything else you'd like me to do to get this merged?

Vlix:
Nope, this looks good and is in the roadmap for the next minor release.

We can treat this as __approved__, though I have a TODO in this thread that I'll add right after I merge it.

Issue #873, opened 2022-03-15
warp: Use custom exception on listen socket closure


shlevy:
Currently, my custom uncaught exception logger needs:

```haskell
onException Nothing e
  | Just (ioeGetErrorType -> et) <- fromException e
  , et == ResourceVanished || et == InvalidArgument = pure ()
```

in order to avoid logging the `invalid argument` error that results from the listen socket being closed (to start graceful shutdown). It would be nice to have a custom exception type passed to the uncaught exception handler in this case, as it's an internal implementation detail if there could be other cases of `ResourceVanished` or `InvalidArgument` thrown prior to the construction of the `Request` which we actually do want to log.

snoyberg:
Making such a change would end up silently breaking lots of existing code that's doing exactly what you've done here.

shlevy:
Hmm so options here:

1. Do nothing
2. Accept the breakage
3. Have some new interface generalizing `setOnException` with a richer context

Probably 1 is the right answer for the library but I wouldn't say no to 3.

snoyberg:
An option (4) that I think would work here: provide predicate functions that test for the common cases people care about, e.g. `isListenSocketClosedException :: SomeException -> Bool`

Issue #603, opened 2017-01-03
warp terminates silently


agrafix:
On a system I've observed that a web server application written using `warp` sometimes terminates with `exit code 0` for no reason. After some debugging, I noticed that it ran out of file descriptors. The problem lies here: https://github.com/yesodweb/wai/blob/c6a542cc4bc51106b207432b1ef81e8dd675a323/warp/Network/Wai/Handler/Warp/Run.hs#L224-L253 . With the default configuration, the exception is NOT logged and the runLoop simply terminates. I think we should always log the expection in default configuration if the server terminates.

snoyberg:
Reading through that code, shouldn't `settingsOnException` be able to print the exception?

agrafix:
Yes, but I think the default `defaultOnException` with `defaultShouldDisplayException` misses the out of file descriptors exceptions and thus does not print it

snoyberg:
Here's the relevant code:

https://www.stackage.org/haddock/lts-7.14/warp-3.2.9/src/Network.Wai.Handler.Warp.Settings.html#defaultShouldDisplayException

Which of those clauses is catching the out of file descriptors exception?

Vlix:
I think this issue has been solved by catching the FD exhaustion (which has been in the code for a while now) and either waiting for the amount of connections to go down, or to just try waiting for another connection (after #1084)

PR #1106, opened 2026-08-16
warp: tell the caller when the accept loop loses its listener

NorfairKing:
When `accept()` fails for a reason nobody asked for, warp ends the accept loop and hands the caller a plain `()`. That is exactly what a graceful shutdown returns, so a server that can no longer accept is indistinguishable from one that was asked to stop. Whatever supervises it never learns it is dead, and it sits there holding a socket it will never accept on again. This rethrows instead, while still ending quietly when the listening socket is closed on purpose.

## Minimal repro

`accept()` fails with `ENFILE`, which is the system-wide descriptor limit rather than this process's own, so warp's `EMFILE` retry does not cover it:

```haskell

import Control.Concurrent
import Control.Exception
import Foreign.C.Error (eNFILE, errnoToIOError)
import Network.Socket

main :: IO ()
main = do
    let failingAccept _ = ioError (errnoToIOError "accept" eNFILE Nothing Nothing)
        app _ respond = respond (responseLBS status200 [] "ok")
        settings = setAccept failingAccept $ setOnException (\_ _ -> pure ()) defaultSettings

    bracket openFreePort (close . snd) $ \(_, sock) -> do
        outcome <- try (runSettingsSocket settings sock app)
        putStrLn $ case outcome :: Either SomeException () of
            Right () -> "returned () -- indistinguishable from a clean shutdown"
            Left e -> "threw: " ++ show e
```

On master:

```
returned () -- indistinguishable from a clean shutdown
```

With this change:

```
threw: accept: resource exhausted (Too many open files in system)
```

## When does `accept()` actually fail this way?

The branch this changes is everything that is not `ECONNABORTED` and not `EMFILE`. On master all of it ends the accept loop. From `accept(2)`, the ones a long-running server can realistically meet:

| errno | cause | today |
|---|---|---|
| `ENFILE` | system-wide open-file limit reached | ends the loop silently |
| `ENOBUFS`, `ENOMEM` | no free memory, often the socket buffer limit rather than system memory | ends the loop silently |
| `ENETDOWN`, `EHOSTUNREACH`, ... | a network error already pending on one queued connection | ends the loop silently |
| `EPERM` | firewall rules forbid the connection (Linux) | ends the loop silently |
| `EMFILE` | this process's own descriptor limit | retried, since #1084 |
| `EBADF` | the listening socket was closed | ends the loop, which is the graceful shutdown path |
| `EINVAL` | the socket is not listening | ends the loop silently, as if a shutdown had been asked for |

`ENFILE` is the one I would lead with: it is the same situation as `EMFILE`, descriptor exhaustion, but counted system-wide. A box running several processes can hit it while warp sits well under its own limit, and warp handles the one and not the other. That asymmetry is the repro above.

`ENOBUFS`/`ENOMEM` are transient under memory pressure, and a server that stops for good on a transient condition is the same shape of problem.

## The second commit: some of those should be retried, not thrown

Rethrowing everything that is not a deliberate shutdown is too broad, and `accept(2)` says why:

> For reliable operation the application should detect the network errors defined for the protocol after `accept()` and treat them like `EAGAIN` by retrying. In the case of TCP/IP, these are `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`, `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH`, `EOPNOTSUPP`, and `ENETUNREACH`.

Linux reports the new socket's already-pending network errors through `accept()`. They describe one queued connection, not the listening socket, and today one unreachable client stops the whole server: master ends the accept loop for every one of them. So the loop now retries them, exactly as it already does for `ECONNABORTED`. The connection they refer to has left the queue by the time they are reported, so the retry blocks for a new connection rather than spinning on the same failure.

`EOPNOTSUPP` is on that list and deliberately not on the retried one. It is also what a listening socket that is not `SOCK_STREAM` answers; the two cannot be told apart here, and retrying the second would spin forever. It is thrown, which at least reaches whoever passed the socket. There is a test that it terminates rather than spins.

That makes the accept loop's failures three kinds rather than two:

| | |
|---|---|
| retry | it was about one queued connection |
| end quietly | the listening socket was closed on purpose |
| throw | the socket will keep giving this |

This is a second commit so it can be dropped if you would rather have the two-way version, though I think the three-way split is the one that is actually right.

## Why this is the wrong return value

`runSettings` returning `()` has one meaning to a caller: the server stopped because it was told to. That is the value a graceful shutdown produces, and callers act on it by exiting cleanly, or by letting a supervisor treat the exit as intended.

There is no way for the caller to tell that case apart from this one, in which `accept()` broke on its own and the server is alive, listening in name only, and will never answer anything again. `settingsOnException` is called, so it is not silent in the logs, but a log line is not a return value: nothing the caller can branch on says the server is finished.

Reported before as #603, which was closed by the `EMFILE` retry branch. That handles running out of descriptors in this process. The silent stop on every other errno stayed.

## The change

It cannot simply rethrow everything, because closing the listening socket is the documented way to end the loop. That arrives as `EBADF`, and exactly `EBADF`: `close` swaps the descriptor for -1 before the syscall, so an `accept()` racing it has nothing else to find. So the loop ends quietly on that one errno, retries what belongs to a queued connection, and throws the rest.

## Test

`AcceptFailureSpec` pins every direction, because a fix that turned deliberate shutdowns into exceptions would be worse than the bug:

- closing the listening socket on purpose still ends the loop quietly, with no exception;
- `accept()` failing with `ENFILE` reaches the caller;
- `accept()` failing with `ENETDOWN` is retried, and the next request is still served;
- `accept()` failing with `EOPNOTSUPP` terminates rather than spinning.

Master's suite passes, and so does this branch. Reverting just the changes to `Run.hs` while keeping the new tests fails one per behaviour this pins, so each stands on its own:

```
1) AcceptFailure, a listener that goes away, rethrows when accept fails for a reason nobody asked for
     runSettingsSocket returned normally after accept() failed, so a lost
     listener looks just like a clean shutdown
```

## Versioning

3.5.0. `runSettings` throws where it used to return, which is the "changing behavior of existing public API" case in CONTRIBUTING, so it is a major version under the PVP. I first bumped this to 3.4.15.1 on the grounds that anyone relying on the old behaviour was relying on not being told their server was dead; @Vlix called it as a major, which is the right reading of the rule, and it is now 3.5.0.

## Checklist

Before submitting:

- [x] Bumped the version number, to 3.5.0

After submitting:

- [x] Update the Changelog.md file with a link to this PR
- [ ] Check that CI passes

## Relation to #873

#873 is a request to stop leaning on `InvalidArgument` to mean "the listener went away", because it is an implementation detail that could in principle have other causes, and the reporter is doing that test by hand in their own `setOnException`. This PR does not lean on it: whether the loop ends quietly or the exception reaches the caller is decided on `EBADF` alone.

Two things from that thread bear on this PR. @snoyberg ruled out a custom exception type as silently breaking existing code, and suggested instead exposing a predicate, `isListenSocketClosedException :: SomeException -> Bool`. The check added here is that predicate, written inline and unexported. If you want it, extracting and exporting it would serve this PR and close most of #873 at once; I have kept it inline so this stays a bug fix with no new API, but say the word and I will do it here or as a follow-up.

The other is from #979: `network`'s `close` atomically swaps the descriptor for -1 before the syscall, so `accept()` on a closed listening socket is *certain* to give `EBADF`, not merely likely. The deliberate-shutdown path is therefore reliable, which is what makes rethrowing everything else safe.

## P.S. one thing I decided rather than assumed

On the rethrow path `settingsOnException` is still called before the exception propagates, so a caller with a custom handler logs it and then also receives it from `runSettings`. I kept the existing reporting rather than making the new path quiet, on the grounds that a caller might have no logging of its own, but it does mean one failure reported twice. Also happy to change.

### DIFF
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -6,13 +6,17 @@
   for a reason the listening socket will keep giving, such as `ENFILE` or
   `ENOMEM`. Previously the accept loop ended and the caller was handed a `()`,
   which is what a graceful shutdown returns, so a server that could no longer
-  accept was indistinguishable from one that had been asked to stop.
+  accept was indistinguishable from one that had been asked to stop. POSIX
+  only: on Windows `network` reports a socket error without an errno, so warp
+  cannot tell one `accept()` failure from another there.
   [#1106](https://github.com/yesodweb/wai/pull/1106)
 * The accept loop now retries, rather than stopping, on the errors `accept()`
   reports for a single queued connection: `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`,
   `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH` and `ENETUNREACH`. `accept(2)` asks for
   these to be treated like `EAGAIN`, which is how `ECONNABORTED` was already
-  handled. One unreachable client no longer stops the server.
+  handled. One unreachable client no longer stops the server. POSIX only: on
+  Windows `network` reports a socket error without an errno for warp to match
+  on.
   [#1106](https://github.com/yesodweb/wai/pull/1106)
 
 ## 3.4.15
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -8,7 +8,8 @@
   which is what a graceful shutdown returns, so a server that could no longer
   accept was indistinguishable from one that had been asked to stop. POSIX
   only: on Windows `network` reports a socket error without an errno, so warp
-  cannot tell one `accept()` failure from another there.
+  cannot tell one `accept()` failure from another and behaviour there is
+  unchanged.
   [#1106](https://github.com/yesodweb/wai/pull/1106)
 * The accept loop now retries, rather than stopping, on the errors `accept()`
   reports for a single queued connection: `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`,
@@ -16,7 +17,7 @@
   these to be treated like `EAGAIN`, which is how `ECONNABORTED` was already
   handled. One unreachable client no longer stops the server. POSIX only: on
   Windows `network` reports a socket error without an errno for warp to match
-  on.
+  on, so behaviour there is unchanged.
   [#1106](https://github.com/yesodweb/wai/pull/1106)
 
 ## 3.4.15
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -1,5 +1,13 @@
 # ChangeLog for warp
 
+## 3.4.15.1
+
+* `runSettings` and friends now rethrow when `accept()` fails for a reason
+  other than the listening socket being closed on purpose. Previously the
+  accept loop ended and the caller was handed a `()`, which is what a graceful
+  shutdown returns, so a server that could no longer accept was
+  indistinguishable from one that had been asked to stop.
+
 ## 3.4.15
 
 * Support `103 Early Hints` over HTTP/2: the HTTP/2 handler installs
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -3,10 +3,16 @@
 ## 3.5.0
 
 * Breaking change: `runSettings` and friends now rethrow when `accept()` fails
-  for a reason other than the listening socket being closed on purpose.
-  Previously the accept loop ended and the caller was handed a `()`, which is
-  what a graceful shutdown returns, so a server that could no longer accept was
-  indistinguishable from one that had been asked to stop.
+  for a reason the listening socket will keep giving, such as `ENFILE` or
+  `ENOMEM`. Previously the accept loop ended and the caller was handed a `()`,
+  which is what a graceful shutdown returns, so a server that could no longer
+  accept was indistinguishable from one that had been asked to stop.
+  [#1106](https://github.com/yesodweb/wai/pull/1106)
+* The accept loop now retries, rather than stopping, on the errors `accept()`
+  reports for a single queued connection: `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`,
+  `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH` and `ENETUNREACH`. `accept(2)` asks for
+  these to be treated like `EAGAIN`, which is how `ECONNABORTED` was already
+  handled. One unreachable client no longer stops the server.
   [#1106](https://github.com/yesodweb/wai/pull/1106)
 
 ## 3.4.15
--- a/warp/ChangeLog.md
+++ b/warp/ChangeLog.md
@@ -1,12 +1,13 @@
 # ChangeLog for warp
 
-## 3.4.15.1
+## 3.5.0
 
-* `runSettings` and friends now rethrow when `accept()` fails for a reason
-  other than the listening socket being closed on purpose. Previously the
-  accept loop ended and the caller was handed a `()`, which is what a graceful
-  shutdown returns, so a server that could no longer accept was
+* Breaking change: `runSettings` and friends now rethrow when `accept()` fails
+  for a reason other than the listening socket being closed on purpose.
+  Previously the accept loop ended and the caller was handed a `()`, which is
+  what a graceful shutdown returns, so a server that could no longer accept was
   indistinguishable from one that had been asked to stop.
+  [#1106](https://github.com/yesodweb/wai/pull/1106)
 
 ## 3.4.15
 
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -397,13 +397,15 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                             , eHOSTUNREACH
                             , eNETUNREACH
                             ]
+                    isFdExhaustion = isErrno eMFILE
+                    isIntentionallyClosedSocket = isErrno eBADF
                 if | isQueuedConnectionError -> do
                         -- Important to mark the exhaustion issue to be resolved
                         resetFdExhaustion fdRef
                         acceptNewConnection
                      -- Keep in mind to reset the ref when anything other
                      -- than this branch runs
-                   | isErrno eMFILE -> do
+                   | isFdExhaustion -> do
                         handleFdExhaustion e
                         acceptNewConnection
                      -- A graceful shutdown ends this loop by closing the
@@ -412,7 +414,7 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                      -- closing it, so every later accept() is handed -1 and
                      -- fails that way. Matching EBADF alone therefore cannot
                      -- miss a deliberate shutdown.
-                   | isErrno eBADF -> do
+                   | isIntentionallyClosedSocket -> do
                         resetFdExhaustion fdRef
                         settingsOnException set Nothing $ E.toException e
                         return Nothing
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -25,6 +25,7 @@ import Data.IORef (newIORef, readIORef, IORef, writeIORef)
 import Data.Streaming.Network (bindPortTCP)
 import Foreign.C.Error (
     Errno (..),
+    eBADF,
     eCONNABORTED,
     eHOSTDOWN,
     eHOSTUNREACH,
@@ -409,12 +410,12 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                         resetFdExhaustion fdRef
                         settingsOnException set Nothing $ E.toException e
                         -- Closing the listening socket is how a graceful
-                        -- shutdown ends this loop, and it arrives here as an
-                        -- InvalidArgument, which warp already reads as the
-                        -- socket going away in the normal course of running
-                        -- (see 'defaultShouldDisplayException'). Returning
-                        -- Nothing for that one ends the loop quietly and
-                        -- 'runSettings' returns ().
+                        -- shutdown ends this loop, and it gives EBADF rather
+                        -- than merely tending to: 'close' swaps the descriptor
+                        -- for -1 before the syscall, so an accept() racing it
+                        -- has nothing else to find. Returning Nothing for that
+                        -- one ends the loop quietly and 'runSettings' returns
+                        -- ().
                         --
                         -- Every errno that is left is one the listening socket
                         -- will keep giving: descriptors exhausted system-wide,
@@ -424,7 +425,7 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                         -- server that was asked to stop would be reported
                         -- identically and the caller could not tell which had
                         -- happened. Throw instead, so it can.
-                        if ioeGetErrorType e == InvalidArgument
+                        if isErrno eBADF
                             then return Nothing
                             else E.throwIO e
 
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -406,11 +406,12 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                    | isErrno eMFILE -> do
                         handleFdExhaustion e
                         acceptNewConnection
-                     -- Closing the listening socket is how a graceful shutdown
-                     -- ends this loop, and it gives EBADF rather than merely
-                     -- tending to: 'close' swaps the descriptor for -1 before
-                     -- the syscall, so an accept() racing it has nothing else
-                     -- to find.
+                     -- A graceful shutdown ends this loop by closing the
+                     -- listening socket, and that always arrives here as
+                     -- EBADF: 'close' replaces the descriptor with -1 before
+                     -- closing it, so every later accept() is handed -1 and
+                     -- fails that way. Matching EBADF alone therefore cannot
+                     -- miss a deliberate shutdown.
                    | isErrno eBADF -> do
                         resetFdExhaustion fdRef
                         settingsOnException set Nothing $ E.toException e
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -431,7 +431,17 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                         -- as resolved here, but just for completeness' sake.
                         resetFdExhaustion fdRef
                         settingsOnException set Nothing $ E.toException e
+#if WINDOWS
+                        -- None of the guards above can match on Windows, where
+                        -- network reports a socket error with no errno on it,
+                        -- so every accept() failure arrives here including the
+                        -- EBADF of a deliberate shutdown. Throwing would turn
+                        -- an ordinary shutdown into an exception, so Windows
+                        -- keeps ending the loop the way it always has.
+                        return Nothing
+#else
                         E.throwIO e
+#endif
 
     handleFdExhaustion e = do
         fdExhaustion <- readIORef fdRef
--- a/warp/Network/Wai/Handler/Warp/Run.hs
+++ b/warp/Network/Wai/Handler/Warp/Run.hs
@@ -371,13 +371,14 @@ acceptConnection set getConnMaker app counter ii fdRef = do
                 let getErrno (Errno cInt) = cInt
                     isErrno err = ioe_errno e == Just (getErrno err)
                     -- Errors about one queued connection rather than about the
-                    -- listening socket. Linux reports the new socket's
-                    -- already-pending network errors through accept(), and
-                    -- accept(2) says to treat those like EAGAIN and retry,
-                    -- which is what eCONNABORTED does here already. The
-                    -- connection they refer to has left the queue, so the
-                    -- retry blocks for a new one rather than spinning on the
-                    -- same failure.
+                    -- listening socket. eCONNABORTED is the familiar one, a
+                    -- peer that went away before it could be accepted. Linux
+                    -- also reports the new socket's already-pending network
+                    -- errors through accept(), and accept(2) asks for those to
+                    -- be treated the same way, retried like EAGAIN. Either
+                    -- way the connection has left the queue, so the retry
+                    -- blocks for a new one rather than spinning on the same
+                    -- failure.
                     --
                     -- eOPNOTSUPP is on that list in accept(2) and is left off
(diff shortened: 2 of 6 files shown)


NorfairKing:
For anyone triaging these together: this one is independent of #1104 and #1105, which are two shapes of a different graceful-shutdown fix. This can be merged on its own. All three bump warp to 3.4.15.1, so whichever lands first will leave the others needing a trivial rebase of that line.

NorfairKing:
Checked against the other two: merging this with #1104 (or #1105) and running warp's suite gives 125 examples and 0 failures. Their `GracefulShutdownSpec` test stops the accept loop with `errnoToIOError "accept" eBADF`, which this PR lets through quietly, so the change here does not disturb it. The only conflict is the `## 3.4.15.1` ChangeLog heading each adds.

NorfairKing:
CI here needs a maintainer to approve the run, so I ran your workflow unchanged in my fork against the same commit (182390b4): all 18 cells green, 6 resolvers from lts-20 through nightly on ubuntu, macos and windows. https://github.com/NorfairKing/wai/actions/runs/31946220514

warp/test/AcceptFailureSpec.hs
```diff
@@ -0,0 +1,85 @@
...
+spec :: Spec
+spec = do
+    -- Running out of descriptors does not arrive through the branch the next
+    -- group covers. eMFILE has its own, which waits for a connection to close
+    -- and then retries. That wait is the hazard: it used to block until the
+    -- connection count dropped *below* what it was on entry, so a server
+    -- holding no connections when the descriptors ran out waited for a
+    -- decrease that could never come, and stopped accepting for good without
+    -- raising anything. A server can sit like that indefinitely, logging
+    -- "Too many open files" and then going quiet.
```
Vlix: What is this comment trying to say?
What is "the next group"?
Does `eMFILE` wait in the current code? And is that bad? Should it be reverted? Or is it now fixed and this is just recounting how the situation used to be?

... Is this written by an LLM? It feels superfluous and as a comment above the first test could be much more to the point.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -372,7 +372,20 @@ acceptConnection set getConnMaker app counter ii fdRef = do
...
+                        -- (see 'defaultShouldDisplayException').
+                        --
+                        -- Anything else is an accept() that broke on its own,
+                        -- and has to reach the caller: ending the loop hands
+                        -- back a () that cannot be told apart from a clean
+                        -- stop, leaving the caller holding a server that will
+                        -- never accept again and does not know it.
+                        if ioeGetErrorType e == InvalidArgument
+                            then return Nothing
+                            else E.throwIO e
```
Vlix: I don't get why `[it will leave] the caller holding a server that will never accept again and does not know it`?
If there's an exception, the server stops; it doesn't keep going.
In what way does the caller hold on to the server?

Vlix (commented):
In a sense, it's best to fail loud, but it does change the behaviour of the `runXXX` functions. So this would require a major version bump.

@kazu-yamamoto do you have strong opinions on loudly throwing exceptions when they break the accept loop (and the server)?

kazu-yamamoto:
Major version up is fine to me.
Probably, we need to roadmap on "which PRs should be merged when" since we have a lot of open RPs now.

For this PR, I would like to check `ioe_errno`, too.

warp/test/AcceptFailureSpec.hs
NorfairKing: Hi @Vlix yes I'm using an LLM. Apologies for the verbosity. Rest assured that I've reviewed the output before asking you to review anything.

NorfairKing: To answer the questions: `EMFILE` does still wait, that wait is correct, and #1084 is what made it terminate. "The next group" stops being true the moment anyone reorders the file.

I have removed the test rather than reword the comment. It pinned `waitForDecreased` from #1084 — released, unrelated to this PR, and here only because it sits next to the code I touched.

warp/Network/Wai/Handler/Warp/Run.hs
NorfairKing: You are right, and the comment invited it: it described the behaviour being *removed* while sitting directly above the code removing it. "Holding a server" is wrong too — the server stops either way. What is left is a listening socket nobody accepts on again, and a caller that cannot tell why it stopped.

Reworded to say what each branch does now. The condition itself has also changed to `EBADF`, per @kazu-yamamoto's note about `ioe_errno`.

NorfairKing:
Thanks. Bumped to 3.5.0, ChangeLog entry marked breaking.

`ioe_errno`: done, and you were right that it wanted looking at. The quiet-exit test was `ioeGetErrorType == InvalidArgument`, which is two errnos, not one — `EBADF` for the closed listening socket this is about, and `EINVAL` for a socket that is not listening. The second is a fault worth raising and was being swallowed as if someone had asked for a shutdown. It now checks `EBADF` through the same `isErrno` helper the branches above use. That is exact rather than nearly right: `close` swaps the descriptor for -1 before the syscall, so an `accept()` racing it finds nothing else. It also drops this PR's reliance on `InvalidArgument` standing in for a lost listener, which is what #873 objects to.

On sequencing: no rush from me. This one is independent of #1104, and if it is easier to land that first I will rebase the version line.

NorfairKing:
Working out when `accept()` actually fails this way changed my mind about the first commit, so there is a second one.

`accept(2)` asks for the network errors Linux passes through `accept()` — `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`, `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH`, `ENETUNREACH` — to be retried like `EAGAIN`. They are about one queued connection, not the listening socket, and master ends the accept loop on every one: today a single unreachable client stops the server. They are now retried alongside `ECONNABORTED`. The connection has left the queue by then, so the retry blocks for a new one rather than spinning on the same failure.

`EOPNOTSUPP` is on that list and deliberately not on the retried one: it is also what a listening socket that is not `SOCK_STREAM` answers, the two are indistinguishable here, and retrying that spins forever. It throws, with a test that it terminates.

Three outcomes instead of two: retry (one queued connection), end quietly (the listener was closed on purpose), throw (the socket will keep giving this).

Tests pass. Reverting just the `Run.hs` changes while keeping the tests fails one per behaviour, so each is pinned on its own rather than resting on the others. Separate commit, so drop it if you would rather the PR kept to its title.

NorfairKing:
## This PR behaves differently on Windows, and not well

I went looking for whether `EBADF` is reliable and found that on Windows it never arrives at all. Worth laying out before this goes further, because it changes what the PR does there.

`network` builds socket errors differently per platform. `Network/Socket/Internal.hs`, under `mingw32_HOST_OS`:

```haskell
throwSocketErrorCode name rc = do
    pstr <- c_getWSError rc
    str  <- peekCString pstr
    ioError (ioeSetErrorString (mkIOError OtherError name Nothing Nothing) str)
```

with the header comment on `throwSocketErrorCode` saying "On Windows, do not use errno. Use a system error code instead." So every `accept()` failure on Windows has `ioe_errno = Nothing` and an `IOErrorType` of `OtherError`, with the WSA code flattened into a description string and discarded.

That means **no `isErrno` guard in `acceptNewConnection` can ever match on Windows**, and every `accept()` failure falls to the last branch.

Reading the table: **retry** = accept again; **wait, retry** = wait for a connection to close, then accept again; **stop** = accept loop ends and `runSettings` returns `()`; **throw** = the exception reaches the caller. Bold marks a cell this PR changes. `settingsOnException` is called in every non-retry case on both platforms, so none of this is silent in the logs; it is about what the caller can act on.

| `accept()` fails with | Linux, master | Linux, this PR | Windows, master | Windows, this PR |
|---|---|---|---|---|
| `ECONNABORTED` | retry | retry | stop | **throw** |
| `EMFILE` | wait, retry | wait, retry | stop | **throw** |
| `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`, `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH`, `ENETUNREACH` | stop | **retry** | stop | **throw** |
| `EBADF` — the listening socket was closed, i.e. a graceful shutdown | stop | stop | stop | **throw** |
| `EINVAL`, `ENFILE`, `ENOMEM`, `EPERM`, `EOPNOTSUPP`, … | stop | **throw** | stop | **throw** |

Three things fall out of that.

**This PR turns a graceful shutdown into an exception on Windows.** Bottom-right of the `EBADF` row. That is a regression against master and it is the reason I am raising this rather than just fixing the comment. It is not caused by the switch to `ioe_errno` either: the first commit tested `ioeGetErrorType == InvalidArgument`, and Windows gives `OtherError`, so the PR has behaved this way on Windows since it was opened.

**Master already has a Windows bug here, independent of this PR.** Look at the first two rows of the "Windows, master" column: because no guard matches, `ECONNABORTED` and `EMFILE` are not retried on Windows, so one aborted connection or one descriptor shortage ends the accept loop for good. #1084's retry has only ever worked on POSIX. Happy to open that separately if it is news.

**CI does not cover any of it.** `tests.yml` passes `--no-run-tests` on `windows-latest`, so those cells compile the suite and never run it. My earlier "all 18 cells green" comments on this PR and on #1104 were true about compilation and I should not have implied more than that; sorry.

### What I would like a steer on

I do not have a good way to fix Windows properly from here: the WSA code is gone by the time warp sees the exception, so there is nothing to match on. Options as I see them:

1. Guard the throw behind `#if !WINDOWS`, so Windows keeps returning `()` exactly as it does on master and the PR becomes POSIX-only. Smallest change, no regression, no improvement there either.
2. Take the regression knowingly, on the grounds that a Windows server that has lost its listener is better off throwing than returning `()` — but that also throws on ordinary shutdown, which seems clearly wrong.
3. Stop inferring intent from the error. Warp hands out the socket-closing action through `settingsInstallShutdownHandler`, so it could record that it closed the socket and check that instead of asking the errno. Works on every platform and drops the last of what #873 objects to. Bigger, and it would not cover a caller that closes the listening socket without going through that action.

I lean towards 1 for this PR and 3 as its own change, but this is your platform matrix, so tell me which you would rather have and I will do it.

Vlix:
Windows always falling through is news to me. (though it is not news that Windows isn't as well supported in the `warp` code)

I believe the reason we don't run the tests on Windows is because one of the packages fails for some reason. Apparently we should be running all the tests that will run.
(We should also cache better, I found [this comment](https://github.com/commercialhaskell/stack/issues/5754#issuecomment-1696156869) about how to cache stack in CI)

Given this new information, it might be a good idea to first remedy the Windows/Linux exception behaviour imbalance.

In short, we'll keep this PR open, and we'll want to implement something like 3. first.

@kazu-yamamoto Do you think we could get better exceptions for Windows?

NorfairKing:
@Vlix I'm not a fan of holding off on improvements because we could make an even bigger improvement. Would it be acceptable to already merge this PR with some CPP instead?

warp/test/AcceptFailureSpec.hs
Vlix, line 1: I think it'd be better to match on the exceptions we expect. (in the `Left _ -> return ()` cases)
That way these are also regression tests.

warp/ChangeLog.md
```diff
@@ -1,5 +1,20 @@
 # ChangeLog for warp
 
+## 3.5.0
```
Vlix, line 3: If we don't fix the Windows exception handling before merging this, the Changelog should also indicate that these changes do not affect `warp` when running on Windows.

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -358,7 +370,34 @@ acceptConnection set getConnMaker app counter ii fdRef = do
...
+                            [ eCONNABORTED
+                            , eNETDOWN
+                            , ePROTO
+                            , eNOPROTOOPT
+                            , eHOSTDOWN
+                            , eNONET
+                            , eHOSTUNREACH
+                            , eNETUNREACH
+                            ]
+                if | isQueuedConnectionError -> do
```
Vlix, line 402: I like the descriptive guard here. Doing the same for the other cases (file exhaustion and intended socket closing) would also improve readability and understanding.

Could we also make a test that checks the `isIntentionallyClosedSocket` (i.e. `isErrno eBADF`) catches closed sockets?

Vlix (commented):
I had some more thoughts on this PR when reading the diffs. See my comments.

warp/test/AcceptFailureSpec.hs
NorfairKing: Done. Both rethrow tests now check the errno that came back rather than taking any `Left`:

```haskell
failedWith :: Errno -> SomeException -> Bool
failedWith (Errno wanted) e = case fromException e of
    Just ioe -> ioe_errno ioe == Just wanted
    Nothing -> False
```

and they say which exception escaped when it is the wrong one. I checked that the check earns its keep: pointing the `ENFILE` test at `ENETDOWN` instead fails it with `expected the ENFILE from accept(), got: accept: resource exhausted (Too many open files in system)`.

warp/ChangeLog.md
NorfairKing: Both entries now say for themselves that they are POSIX-only, rather than a third bullet leaning on the two above it.

Rather than only document it, the last commit makes it true: the throw is behind `#if WINDOWS`, so Windows keeps returning `()` exactly as master does and this PR is a no-op there instead of a regression. That commit is on its own and drops cleanly if you would rather take the Windows change knowingly, or wait for something that can tell those failures apart there.

warp/Network/Wai/Handler/Warp/Run.hs
NorfairKing: Named both: `isFdExhaustion` and `isIntentionallyClosedSocket`, the latter your suggestion. All four guards now read as what they mean rather than as an errno the reader has to place.

On the test, I think you already have it. `"ends the accept loop quietly when the socket is closed on purpose"` uses the real `accept` and really closes the listening socket through `setInstallShutdownHandler`, so the `EBADF` there is a genuine one from the kernel rather than a synthesised `errnoToIOError`. That is the test that fails if `isIntentionallyClosedSocket` stops matching a closed socket. Say if you meant something more direct and I will add it.

Vlix:
> @Vlix I'm not a fan of holding off on improvements because we could make an even bigger improvement. Would it be acceptable to already merge this PR with some CPP instead?

We'll be holding off on this anyway (for the time being), since it will force a major version bump. But I guess now that the Changelog mentions the behaviour on Windows and the CPP keeps the original behaviour on Windows, we can add it to the release when we get to it without waiting for the Windows side to get better exceptions.

kazu-yamamoto:
> Do you think we could get better exceptions for Windows?

@Vlix Please do not ask me questions about Windows ;-)

Vlix:
@NorfairKing 
Please update the upper bounds of `warp` in the .cabal file of `wai-app-static` and `warp-tls` so that CI keeps running.

NorfairKing:
Done, both raised to `< 3.6`. Those two are the only packages in the repo that cap warp at all, the rest depend on it unbounded.

I compiled both against the warp in this branch rather than just letting the bound parse, since raising it only says the plan is allowed. No API moved in 3.5.0, so nothing needed changing beyond the bound itself.

NorfairKing:
@Vlix is there anything else you'd like me to do to get this merged?

Vlix:
I just reran the checks (last time they failed because of the issues with the hackage servers) and everything looks good.
We'll keep this PR open until we have published the next minor version, then this will be added to the next major version release.

This PR is now accepted, but I won't set it as such, so that other maintainers don't merge it accidentally.

NorfairKing:
@Vlix is there any reason this hasn't been merged yet? I don't understand the reasoning behind not releasing approved changes.

Vlix:
This change constitutes a major version increase.
We want to release a minor version first, #1108 shows what we're waiting on.

I'm currently trying to make our CI better, so we can maybe fix the failing tests we have on Windows.
I found we aren't running tests on Windows, and I'd like to know if the new changes we have lined up have introduced failing tests, or that they were always there.

I'll see if I can go forth with the `warp` minor release anyway, if I can't get the CI working today.

kazu-yamamoto:
@Vlix Just FYI on Windows.

Coming GHC 9.12.5 and 9.14.2 includes a fix of WinIO (say WinIO version 2). After they are released, I will release `network` library which make a full use of WinIO version 2. 

I don't have any experience with WinIO yet, but I think most of the special hacks for Windows will become unnecessary.
We will have to keep these hacks around for a while to support older versions of GHC, though.

Vlix:
@NorfairKing 
I've just published [`warp-3.4.16`](https://hackage.haskell.org/package/warp-3.4.16), so we can go forward with the PRs for the major release.

If you'd be so kind as to rebase on master to use the latest CI, and to resolve any conflicts that may be outstanding, we can go forward with merging this PR, I think.

NorfairKing:
I'm on it now but I won't be able to help anymore after that. I'm expecting a child this week.

NorfairKing:
Rebased onto master and pushed. Four things there met the branch, so this is more than a replay.

**Version.** 3.4.16 is out, so the entry moves above it and the bump is now 3.4.16 → 3.5.0. Nothing else about the change moved.

**ConnectionExceptionSpec, from #1114.** One example ends the accept loop by failing `accept()` with `userError "accept failed"`, to check that the legacy observer and not the peer observer hears about it. An errno-less failure like that is exactly what this PR hands to the caller, so the example threw instead of returning and the suite failed. It now fails with the errno of a closed listening socket, which still ends the loop, and asserts the same thing about the same code. It sits in the first commit, with the change that alters the contract.

**Windows tests run in CI now**, since #1117. Two `AcceptFailureSpec` examples expect an `accept()` failure to reach the caller, which is the one thing the Windows commit opts out of, so they would have gone red on `windows-latest`. Under `#if WINDOWS` they are replaced by one that expects the loop to end, in the style of the `EarlyHintsSpec` guard. It is in "warp: leave Windows exactly as it was", the commit that creates the difference, so the behaviour that commit preserves is asserted rather than skipped.

**warp-tls caps warp twice now**, in the library and again in its test suite. Both are at `< 3.6`.

On Linux: warp is 133 examples and 0 failures, and the warp-tls and wai-app-static suites pass against warp 3.5.0. Reverting only the `Run.hs` change and keeping the tests fails 3. I cannot run the Windows branch here; it typechecks with `-Wall`, and CI will say the rest.

Vlix (approved):
Looks good. Thanks for the contribution 🙏

PR #1107, opened 2026-08-17
Optimistic recv

kazu-yamamoto:
Trying to take over #1099.

I might be overlooking something, but I thought #1099 could be improved in this way. Could you please verify if this is the right direction? If so, we can fix the redundant `recv` code later.


warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -144,6 +144,22 @@ socketConnection set s = do
...
+      else do
+        makeGracefulRecvSlow sock pool ss appsInProgress
+  where
+    isShuttingdown = atomically $ do
+       s <- currentShuttingDownStateSTM ss
+       n <-readTVar appsInProgress
+       return (s && n <= 0)
+
+makeGracefulRecvSlow :: Socket -> BufferPool -> ServerState -> TVar Int -> Recv
+makeGracefulRecvSlow sock pool ss appsInProgress = do
```
Vlix, line 159: How about:

```hs
makeGracefulRecv sock pool ss appsInProgress = do
    weSkip <- skipOptimisation
    if weSkip
        then slowPath
        else receiveNoWait sock pool >>= maybe slowPath pure        
  where
    slowPath =
        makeGracefulRecvSlow sock pool ss appsInProgress
    skipOptimisation =
#if !WINDOWS
        isShuttingdown
#else
        pure True
#endif
    isShuttingdown = atomically $ do
       s <- currentShuttingDownStateSTM ss
       n <- readTVar appsInProgress
       return (s && n <= 0)
```

This way we don't even check on Windows.

Vlix (approved):
Looks ok.

We can replace everything after `withFdSocket` in the `receiveNoWait` with `recvBufNoWait` once it's exported from `network`.

And `tryWithBufferPool` section can be replaced with the following if you feel like it:

```hs
withBufferPool :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO ByteString
withBufferPool pool@(BufferPool _ _ ref) f = do
    withBufferPool' pool f $ \consumed -> do
        writeIORef ref $ unsafeDrop consumed buf
        return $ unsafeTake consumed buf

-- | Like 'withBufferPool' for fillers that can decline to fill:
--   a negative return value from the filler leaves the pool untouched
--   and produces 'Nothing'.
tryWithBufferPool
    :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO (Maybe ByteString)
tryWithBufferPool pool@(BufferPool _ _ ref) f = do
    withBufferPool' pool f $ \consumed ->
        if consumed > 0
            then do
                writeIORef ref buf
                return Nothing
            else do
                writeIORef ref $ unsafeDrop consumed buf
                return $ Just $ unsafeTake consumed buf

-- Common code for working with a buffer pool
withBufferPool' :: BufferPool -> (Buffer -> BufSize -> IO Int) -> (Int -> IO a) -> IO a
withBufferPool' (BufferPool l h ref) f cont = do
    buf0 <- readIORef ref
    buf <-
        if BS.length buf0 >= l
            then return buf0
            else mallocBS h
    withForeignBuffer buf f >>= cont
```

Vlix:
Also, this adds to the `recv` API, which needs a version bump and a publish.

And the `warp` code should have this optimisation behind CPP checking for this new `recv` version.

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: The current code is easier to understand for me.

kazu-yamamoto:
@Vlix `network` v3.2.9.0 has been released.

My last concern is `isShuttingdown` in `akeGracefulRecv`. I'm not sure whether or not we should check `appsInProgress` in this case.

recv/Network/Socket/BufferPool/Buffer.hs
```diff
@@ -34,15 +35,36 @@ newBufferPool l h = BufferPool l h <$> newIORef BS.empty
 --   how many bytes are filled in the buffer.
 --   The buffer in the buffer pool is automatically managed.
 withBufferPool :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO ByteString
-withBufferPool (BufferPool l h ref) f = do
+withBufferPool pool@(BufferPool _ _ ref) f = do
```
Vlix, line 38: Not a blocker, but I think `withBufferPool` should never get negative `Int`s, right? This isn't described in the documentation above it, and this is an exported function.

I think we should either use `drop` and `take`, or we mention that the resulting `Int` should _NOT_ be less than zero.

recv/Network/Socket/BufferPool/Recv.hs
```diff
@@ -20,6 +21,18 @@ import Network.Socket.BufferPool.Types
...
+-- | Like 'receive' but never blocks and never involves the IO manager:
+--   'Nothing' means no data was available (or an error occurred, which a
+--   subsequent blocking 'receive' will report properly). @Just \"\"@ is EOF.
+--   On Windows this always returns 'Nothing'.
+receiveNoWait :: Socket -> BufferPool -> IO (Maybe ByteString)
+receiveNoWait sock pool = tryWithBufferPool pool $ \ptr size ->
+    -- The socket is non-blocking, so an unsafe foreign call is fine:
+    -- recv(2) returns immediately either way. Both EAGAIN and real
+    -- errors map to a negative result, deferring to the blocking path
+    -- so errors surface there with their errno intact.
```
Vlix: This comment is from the code before, I take it?
There's no foreign call anymore, right? (Not in this code, at least)

```suggestion
    -- The socket is non-blocking. Both EAGAIN and real errors map to
    -- a negative result, deferring to the blocking path so errors
    -- surface there with their errno intact.
```

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -144,6 +144,22 @@ socketConnection set s = do
...
+        case mbs of
+          Just bs -> return bs
+          Nothing -> makeGracefulRecvSlow sock pool ss appsInProgress
+      else do
+        makeGracefulRecvSlow sock pool ss appsInProgress
+  where
+    isShuttingdown = atomically $ do
+       s <- currentShuttingDownStateSTM ss
+       n <-readTVar appsInProgress
+       return (s && n <= 0)
```
Vlix: I'm pretty sure we only have to check the shutting down state.
The moment we are shutting down, this then defers to the slow path which will do the complete `isShuttingDown` + `appsInProgress` check.

```suggestion
    tryFastPath <- not <$> currentShuttingDownState ss
    if tryFastPath
        then receiveNoWait sock pool >>= maybe slowPath pure
        else slowPath
  where
    slowPath = makeGracefulRecvSlow sock pool ss appsInProgress
```

Vlix (changes requested):
Yeah, looks better already.

With a few adjustments, this should be ready to be merged 👍

recv/Network/Socket/BufferPool/Buffer.hs
kazu-yamamoto: Both done.

recv/Network/Socket/BufferPool/Recv.hs
kazu-yamamoto: Applied. Thanks.

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: I did not notice that you showed this suggested code.
So, I made two commits for this.

recv/Network/Socket/BufferPool/Buffer.hs
```diff
@@ -32,17 +32,39 @@ newBufferPool l h = BufferPool l h <$> newIORef BS.empty
...
+    :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO (Maybe ByteString)
+tryWithBufferPool pool@(BufferPool _ _ ref) f = do
+    (buf, consumed) <- applyBufferPool pool f
+    if consumed < 0
+        then do
+            writeIORef ref buf
+            return Nothing
+        else do
+            writeIORef ref $ BS.drop consumed buf
+            return $ Just $ BS.take consumed buf
```
Vlix, line 56: Here, we _could_ use the `unsafeDrop/-Take` functions, since we make sure `consumed` is non-negative. But I'm fine with it, either way.

recv/Network/Socket/BufferPool/Recv.hs
```diff
@@ -20,6 +21,17 @@ import Network.Socket.BufferPool.Types
 receive :: Socket -> BufferPool -> Recv
 receive sock pool = withBufferPool pool $ \ptr size -> recvBuf sock ptr size
 
+-- | Like 'receive' but never blocks and never involves the IO manager:
+--   'Nothing' means no data was available (or an error occurred, which a
+--   subsequent blocking 'receive' will report properly). @Just \"\"@ is EOF.
+--   On Windows this always returns 'Nothing'.
```
Vlix: Does this still always return `Nothing`, now that we're using `recvBufNoWait`?

warp/Network/Wai/Handler/Warp/Run.hs
```diff
@@ -144,6 +144,18 @@ socketConnection set s = do
 -- actively using this 'Socket'.
 makeGracefulRecv :: Socket -> BufferPool -> ServerState -> TVar Int -> Recv
 makeGracefulRecv sock pool ss appsInProgress = do
+    tryFastPath <- not <$> atomically (currentShuttingDownStateSTM ss)
```
Vlix: We don't need to use a transaction here. `currentShuttingDownState` uses `readTVarIO`, which is faster than running in STM.

Vlix (changes requested):
Hard blocker is `currentShuttingDownStateSTM`. We don't need to use an STM transaction there.
And the comment about Windows always returning `Nothing` is also wrong, I think? Now that we use `recvBufNoWait`.

We also need to add `network-3.2.9.0` in the `stack*.yaml` files to make the CI run.

Rest looks good though 👍

recv/Network/Socket/BufferPool/Buffer.hs
kazu-yamamoto: Let's stay as is.

recv/Network/Socket/BufferPool/Recv.hs
kazu-yamamoto: Nice catch.
The last line has been deleted.

warp/Network/Wai/Handler/Warp/Run.hs
kazu-yamamoto: Thanks and done.

kazu-yamamoto:
Now all green.

Vlix (approved):
LGTM. Thank you @kazu-yamamoto and @seanparsons

recv/recv.cabal
Vlix: Woops, almost forgot this dependency

Vlix (changes requested):
Did final checks, and didn't see additions to the ChangeLog of `recv` and the `network` dependency of the library is also still on `network >= 3.1.0`

Also a good idea to add the tests from #1099 ([here](https://github.com/yesodweb/wai/pull/1099/changes/120f829d05da6dcbe63d90cab2dd40a1f8442892..537e2e8e5cbbd0d35d139552e842c120ccab7929#diff-20c288017412aa604554b2049d0f586d9fc8a8dc194839b944ddc167f2e6e832)) to have the `tryWithBufferPool` tested as well.
(make sure to add Sean as `Co-Authored-By` in that commit?)

kazu-yamamoto:
OK. I think we are done.

Vlix (approved):

PR #1109, opened 2026-08-27
Perf/time manager tickle

Vlix:

After submitting your PR:

---

My interpretation of #1097 and the lessons we learned from it.

## Summary

- I wrote most of the reasoning in a big comment inside `System.TimerManager`.
- `tickle` should not call the IO manager in hot loops.
- `pause` does not have to call the IO manager at all.
  - This means the registered action might run before the timeout has passed.
- We now need locks to guarantee we don't get race conditions, because of this.
- `cancel` completely stops the timeout, making a new `register` necessary.
  - (this is also the reason I bumped it to a new major version)


Vlix:
I still want to add tests to make sure it does work as expected, but I think the logic now at least is in place as I imagined it.

Vlix:
Running the benchmarks @seanparsons provided, this change vs `time-manager-0.3.2` looks as follows, taking an average run on my laptop:

## `time-manager-0.3.2`

```
Running 1 benchmarks...
Benchmark response: RUNNING...
benchmarking sendResponse/builder 4 headers content-length
time                 850.4 ns   (835.4 ns .. 868.9 ns)
                     0.996 R²   (0.994 R² .. 0.998 R²)
mean                 850.6 ns   (836.6 ns .. 875.0 ns)
std dev              61.44 ns   (42.89 ns .. 98.22 ns)
variance introduced by outliers: 81% (severely inflated)

benchmarking sendResponse/stream 64 fragments
time                 54.56 μs   (51.14 μs .. 58.57 μs)
                     0.964 R²   (0.948 R² .. 0.977 R²)
mean                 64.66 μs   (59.80 μs .. 73.52 μs)
std dev              22.62 μs   (13.61 μs .. 43.54 μs)
variance introduced by outliers: 98% (severely inflated)

Benchmark response: FINISH
```

## `time-manager-0.4.0`

```
Running 1 benchmarks...
Benchmark response: RUNNING...
benchmarking sendResponse/builder 4 headers content-length
time                 429.9 ns   (426.1 ns .. 436.8 ns)
                     0.998 R²   (0.995 R² .. 1.000 R²)
mean                 428.5 ns   (425.6 ns .. 438.4 ns)
std dev              16.24 ns   (5.729 ns .. 32.57 ns)
variance introduced by outliers: 55% (severely inflated)
                  
benchmarking sendResponse/stream 64 fragments
time                 17.99 μs   (17.96 μs .. 18.02 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 17.96 μs   (17.94 μs .. 17.99 μs)
std dev              71.63 ns   (49.64 ns .. 103.1 ns)
                  
Benchmark response: FINISH
```

Mostly focusing on the `stream 64 fragments` part, since that is what's different in regards to #1097. Keeping the `resume/pause` in place, but not calling to the IO manager on a `pause`.

kazu-yamamoto:
I have one question before I start reviewing.
The term "IO manager" is used in this issue; does it actually refer to `SystemTimerManager`?

time-manager/System/TimeManager/Internal.hs
```diff
@@ -31,13 +33,61 @@ type TimeoutAction = IO ()
...
+    -- ^ The system timer manager the timeout key was registered with.
+    --   Cached so that per-request operations don't re-fetch it.
     , handleState :: ~(IORef HandleState)
+    , handleLastRenewed :: ~(IORef Word64)
+    -- ^ Monotonic time (in nanoseconds) when the timeout was last
+    --   registered or updated.
+    , handleMinRenewGap :: Word64
+    -- ^ 'tickle' is a no-op unless at least this many nanoseconds have
+    --   passed since the last renewal.
+    , handleLock :: ~Lock
```
kazu-yamamoto: I'm not entirely sure, but I have a feeling that if you define it as `handleState :: MVar HandleState`, the `MVar` itself would act as the lock, making this field unnecessary.

time-manager/System/TimeManager.hs
```diff
@@ -121,64 +137,300 @@ withHandleKillThread mgr onTimeout action
 
 ----------------------------------------------------------------
 
+-- ============== NOTE ABOUT THREAD SAFETY ==============
+--
+-- The use of 'IORef's are fine in the current situation where
+-- the 'TimeManager' is supposed to be used in a single thread.
+--
+-- The triggered action, though, is run by the IO manager of
```
kazu-yamamoto: I guess that "the IO manager" should be the system timer manager.

Vlix:
> I have one question before I start reviewing. The term "IO manager" is used in this issue; does it actually refer to `SystemTimerManager`?

I have used the same terminology as used in [`GHC.Event`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/GHC-Event.html#t:TimeoutCallback). The I/O manager is the runtime that runs the registered action, and where `registerTimeout` and `updateTimeout` call out to.

time-manager/System/TimeManager/Internal.hs
Vlix: Semantically, that would not change anything, but it would introduce more mutex grabbing, which I think is slightly slower than grabbing an `IORef`?
I can run the benchmarks with both to see if there's any difference. I'll adjust it if there's no real difference.

kazu-yamamoto: Using `MVar` guarantees that a lock is always acquired upon modification.

kazu-yamamoto:
I’ve read the `GHC.Event` documentation, and I’m surprised at how poor it is.
"Event manager" should be the collective term for the IO manager and the Timer manager.
Shouldn't we use the correct terminology?

time-manager/System/TimeManager/Internal.hs
Vlix: That is true, but reading from an `MVar` is also slower than reading from an `IORef`, right?
So if we want to maximize performance, sticking to an IORef as much as possible could be quicker.

Vlix:
That's fine by me. Should we stick with "Event manager"? Or "Timer manager"?

kazu-yamamoto:
Since this is the `time-manager` package, I think using the term "Timer manager" would minimize user confusion.

time-manager/System/TimeManager/Internal.hs
kazu-yamamoto: If `IORef` is being used intentionally, I will respect that choice.
A reasonable compromise would be to document the intent within the definition.

kazu-yamamoto:
You might be planning to address this in another PR, but I am noting it here for the record.
In the current implementation of `warp`, `pause` and `resume` can be far apart, making their relationship non-intuitive.
If possible, I think it would be better to place them closer together.

Vlix:
Do you mean in #1098? That PR would become obsolete if we accept this PR as it fixes the issue #1098 is trying to work around.

The current code in `warp` has `resume` and `pause` closely together, which will not change.

kazu-yamamoto:
Never mind. It seems I misread it.

time-manager/System/TimeManager.hs
Vlix: I've changed all the mentions of "IO manager" to "Timer Manager"

Vlix:
I've run a few benchmarks using the `IORef` + `Lock` code, and with `MVar HandleState`, and this is the result on my laptop:

<details>
<summary>
Benchmark results: <code>IORef</code> + <code>Lock</code>
</summary>

```
Running 1 benchmarks...
Benchmark response: RUNNING...
benchmarking sendResponse/builder 4 headers content-length
time                 409.7 ns   (408.9 ns .. 410.3 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 408.6 ns   (408.2 ns .. 409.1 ns)
std dev              1.607 ns   (1.337 ns .. 1.964 ns)

benchmarking sendResponse/builder 3 headers chunked
time                 431.2 ns   (430.7 ns .. 431.5 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 430.6 ns   (430.1 ns .. 431.3 ns)
std dev              2.075 ns   (1.296 ns .. 3.901 ns)

benchmarking sendResponse/builder 20 headers content-length
time                 1.167 μs   (1.164 μs .. 1.169 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.163 μs   (1.161 μs .. 1.165 μs)
std dev              5.427 ns   (4.286 ns .. 6.937 ns)

benchmarking sendResponse/no body 204
time                 146.6 ns   (146.5 ns .. 146.7 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 146.5 ns   (146.4 ns .. 146.6 ns)
std dev              363.7 ps   (282.2 ps .. 569.1 ps)


benchmarking sendResponse/stream 64 fragments
time                 17.68 μs   (17.67 μs .. 17.69 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 17.65 μs   (17.63 μs .. 17.67 μs)
std dev              60.61 ns   (48.89 ns .. 83.41 ns)


benchmarking headers/composeHeader 5 headers
time                 117.5 ns   (117.4 ns .. 117.6 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 117.2 ns   (117.1 ns .. 117.3 ns)
std dev              375.9 ps   (327.7 ps .. 435.9 ps)

benchmarking headers/indexRequestHeader
time                 58.03 ns   (57.98 ns .. 58.08 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 58.01 ns   (57.98 ns .. 58.06 ns)
std dev              134.1 ps   (107.5 ps .. 168.6 ps)

benchmarking headers/indexResponseHeader
time                 37.31 ns   (37.30 ns .. 37.33 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 37.29 ns   (37.28 ns .. 37.31 ns)
std dev              56.56 ps   (43.39 ps .. 80.78 ps)

Benchmark response: FINISH
```
</details>

<details>
<summary>
Benchmark results: <code>MVar HandleState</code>
</summary>

```
Running 1 benchmarks...
Benchmark response: RUNNING...
benchmarking sendResponse/builder 4 headers content-length
time                 442.8 ns   (442.4 ns .. 443.2 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 443.6 ns   (443.0 ns .. 445.6 ns)
std dev              3.094 ns   (1.699 ns .. 6.242 ns)

benchmarking sendResponse/builder 3 headers chunked
time                 458.3 ns   (457.9 ns .. 459.0 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 459.5 ns   (458.7 ns .. 461.7 ns)
std dev              3.777 ns   (1.878 ns .. 7.500 ns)

benchmarking sendResponse/builder 20 headers content-length
time                 1.240 μs   (1.239 μs .. 1.242 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.241 μs   (1.241 μs .. 1.243 μs)
std dev              2.996 ns   (2.185 ns .. 4.524 ns)

benchmarking sendResponse/no body 204
time                 159.5 ns   (157.6 ns .. 160.6 ns)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 156.5 ns   (155.6 ns .. 157.6 ns)
std dev              3.030 ns   (2.341 ns .. 3.571 ns)
variance introduced by outliers: 25% (moderately inflated)

benchmarking sendResponse/stream 64 fragments
time                 42.97 μs   (41.06 μs .. 45.03 μs)
                     0.982 R²   (0.974 R² .. 0.991 R²)
mean                 40.89 μs   (39.70 μs .. 42.50 μs)
std dev              4.740 μs   (3.628 μs .. 6.175 μs)
variance introduced by outliers: 87% (severely inflated)

benchmarking headers/composeHeader 5 headers
time                 134.7 ns   (134.4 ns .. 135.0 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 134.6 ns   (134.4 ns .. 135.0 ns)
std dev              1.046 ns   (858.2 ps .. 1.280 ns)

benchmarking headers/indexRequestHeader
time                 56.00 ns   (55.88 ns .. 56.11 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 56.02 ns   (55.91 ns .. 56.13 ns)
std dev              376.9 ps   (305.8 ps .. 558.3 ps)

benchmarking headers/indexResponseHeader
time                 40.53 ns   (40.50 ns .. 40.56 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 40.55 ns   (40.52 ns .. 40.58 ns)
std dev              108.3 ps   (81.29 ps .. 156.7 ps)

Benchmark response: FINISH
```
</details>

Using a lock everywhere does slow the code down slightly, so I'll stick to the original `IORef` + `Lock` implementation.
I've added the explanation to `handleState` and `handleLock`.

I've also renamed `Cancelled` to `Terminated`, since running the action also results in a state that stops completely.

Vlix:
Tests are now added and PR comments are integrated.

If CI is all-green, this should be good to release.

time-manager/System/TimeManager/Internal.hs
```diff
@@ -31,13 +33,69 @@ type TimeoutAction = IO ()
 data Handle = Handle
     { handleTimeout :: Int
     , handleAction :: TimeoutAction
-    , handleKeyRef :: ~(IORef EV.TimeoutKey)
+    , handleTimerManager :: ~TimerManager
+    -- ^ The system timer manager the timeout key was registered with.
+    --   Cached so that per-request operations don't re-fetch it.
     , handleState :: ~(IORef HandleState)
```
kazu-yamamoto, line 39: I would like to suggest to move `handleState` just after `handleLock` so that readers can easily understand the lock protects the state.

Vlix:
I'm trying to make the CI better (better caching, and re-enabling the tests for Windows) and I've found that the tests fail on `time-manager` on Windows (and 2 tests for `warp`).

I'm going to try to get the CI to work on `macos` after caching (it's not caching the `stack` executable for some reason) and then try to find out why `time-manager` tests fail on Windows.

_WIP on branch `ci`_

kazu-yamamoto:
It seems like it's time to rethink timers. @goertzenator argues that timers are unnecessary in HTTP/2. What about HTTP/1.1?

Timers come with a lot of issues, so I’d like to get rid of them if possible.

Vlix:
I don't think we can get around using timeouts. They are pretty essential for web servers, no?

And I don't know why they wouldn't be necessary in HTTP/2?
If a connection is started, but no frames are sent afterwards, the connection stays open indefinitely, unless there's a timeout (i.e. a timer), right?

From what I could find is that Apache [has the following settings](https://httpd.apache.org/docs/2.4/mod/mod_reqtimeout.html) (at least for HTTP 1.1, maybe also 2?) and that this is seen, at least by some, to be one of the best slow loris protection implementations.

kazu-yamamoto:
@goertzenator says:

>  The current effect is that they put timeout constraints on application code, something that Warp http/1.1 and the other language implementations do not do. 

So, I think we just recommend to simplify timers in HTTP/2.

@Vlix I apologize for any confusion caused by my previous remarks, but please proceed with this PR independently of the HTTP/2 timer.

goertzenator:
What I meant was we still need timeouts, but the current timer mechanism is hard to understand.  From reading the code over the past few days it was hard to understand what was being protected and what even should be protected.  I see Kazu touching on this issue too in a message above regarding locality of pause/resume.  In my http2 PR message I suggested simplifications using scoped timeouts and a separate timeout thread.

Now that Kazu has pointed me at this thread (thank you) I see that there are performance concerns with timers.  That turns some gears in my head and makes we wonder if the "timeout thread" idea can solve these problems.  Setting http2 aside for the moment, consider Warp and it's http1 implementation:

1. Instead of various tickle/pause/resume calls, write what we are currently doing into a TVar (ie, reading headers, reading request chunk X, application processing request chunk X, writing response chunk X, application generating response chunk X).
2. Spin up a separate thread that watches what happens to that TVar and implements various `threadDelay`s depending on current state.  If anything times out, exit and kill the connection.

This approach would consolidate all timeout behavior into one place making it easier to understand.

The timeout thread can be very "low load".  It does not need to see all TVar transitions, and can therefore have `threadDelay`s to slow things down.  You could even have it `threadDelay` for a few seconds right off the bat, and many http1.1 sessions would conclude within that time.

There are some performance assumption here:
1. Writing to a TVar is cheap (TVar writes replace tickle/pause/resume).
2. The overhead of starting the timeout thread is low.

kazu-yamamoto:
If we implement this proposal, it looks like it could also be used to monitor the connection status.

kazu-yamamoto:
Instead of using `threadDelay`, if we set up a `TVar` and have the `SystemTimerManager` write to it, we can monitor the timeout via STM.

In my recent code, I represent events using STM and monitor the composed STM.

Vlix:
> If we implement this proposal, it looks like it could also be used to monitor the connection status.

What do you mean with "monitor the connection status"?

---

@goertzenator
I'm all for implementing a better way to do the timeouts. The `TVar` approach sounds plausible and doable, so if anyone can make a small proof of concept we can discuss if it fulfills all the prerequisites, like being prompt in killing the thread when the appropriate timeout is hit.

goertzenator:
I also did not understand the connection status comment.  The avoidance of threadDelay is a solid optimization that saves a thread.

Anyway, I have been playing with this most of the day and will share a repo when it is ready.  I am also working on more thorough benchmarks to capture real world performance.

kazu-yamamoto:
> What do you mean with "monitor the connection status"?

If we export such an STM variable for a worker, other threads can observe the status of the worker.

Vlix:
I've tried the Windows tests on master for `time-manager` and I've tried them on this branch, and they are equally broken on master, as they are after this PR, so I won't let that block a release of `time-manager-0.4.0`.

After the CI is green for everything else, I'll merge and release this new version.
And then continue onward with the minor release of `warp` we have in the pipeline.

PR #1111, opened 2026-08-31
Use atomicWriteIORef in auto-update Event implementation

mkleczek:
A follow-up to #1103 which only addressed Threaded implementation. This change addresses Event.

### DIFF
--- a/auto-update/Control/AutoUpdate/Event.hs
+++ b/auto-update/Control/AutoUpdate/Event.hs
@@ -119,6 +119,6 @@ onceOnTimeHasCome UpdateState{..} action = do
 getUpdateResult :: UpdateState a -> IO a
 getUpdateResult us@UpdateState{..} = do
     onceOnTimeHasCome us $ do
-        writeIORef usLastResult_ =<< usUpdateAction_ =<< readIORef usLastResult_
-        writeIORef usDeleteTimeout_ =<< mkDeleteTimeout usTimeHasCome_ usIntervalMicro_
+        atomicWriteIORef usLastResult_ =<< usUpdateAction_ =<< readIORef usLastResult_
+        atomicWriteIORef usDeleteTimeout_ =<< mkDeleteTimeout usTimeHasCome_ usIntervalMicro_
     readIORef usLastResult_


Vlix:
I guess seeing as most things are threaded/parallel nowadays, every `writeIORef` should just be an `atomicWriteIORef` 🤔

mkleczek:
> I guess seeing as most things are threaded/parallel nowadays, every `writeIORef` should just be an `atomicWriteIORef` 🤔

Only if `IORef` is shared between threads, which is not always the case.

Vlix:
True, but when making a library, you don't know how your users will use it, so unless you can guarantee it won't be shared between threads, or you've added a big warning that it should only be used in single-threaded environments, then it's just better to err on the safe side.

I'll take a closer look at all the `writeIORef`s in this repo, just to see which we also might have missed.

steve-chavez:
@Vlix I've just tried this patch on https://github.com/PostgREST/postgrest/issues/5196#issuecomment-5500511080 and looks there's another bug. The tldr is that we use auto-update to cache clock time every second for jwt validation, currently the update seems to take longer than 1 sec in some cases (which looks like this patch fixes) and in other cases much longer (~100/47 seconds as seen on the issue); the latter bug still persists.

kazu-yamamoto:
We agree to always use `atomicWriteIORef` instead of `writeIORef`.

Similarly, we should always use `atomicModifyIORef'`—rather than `atomicModifyIORef`—even in test code.
Since `atomicModifyIORef` is currently widely used in this repo, we should replace it.

kazu-yamamoto:
Cc: @khibino who implemented the new logic of auto-update.

Vlix:
I don't have a lot of experience with `auto-update`, except for the `Control.Debounce` module. So if anyone else has more knowledge (maybe `khibino`) I'd like to defer to them.

kazu-yamamoto:
@khibino is a colleague of mine, so I'll speak to him about it in person later.
If it turns out he doesn't have the time to work on this issue, let's revert to the original code.

Vlix:
Incorporated this change into #1112

Vlix:
Merged #1112 and published [`auto-update-0.2.8`](https://hackage.haskell.org/package/auto-update-0.2.8)

khibino:
Hi, I'm the original author of the event-based prototype of auto-update.
I'd like to share my perspective on this issue.

First, I highly doubt that not using `atomicWriteIORef` could cause delays on the scale of seconds.
The order of magnitude is simply too different.

I strongly suspect the actual cause is asynchronous exceptions.
If the IO action passed to `onceOnTimeHasCome` is interrupted by an async exception, a critical state is reached: the timeout event is completely lost and never recreated.
Consequently, the value will never be updated again.

Since handling async exceptions perfectly correctly in Haskell is notoriously difficult, my primary recommendation would be to stop using asynchronous exceptions in the code that triggers this.

However, if we absolutely must implement a defensive workaround on the library side (though I am somewhat reluctant), using `bracket` might be a viable approach to ensure the timer registration always happens.

Here is an example of what that could look like:

```hs
onceOnTimeHasCome :: UpdateState a -> IO () -> IO () -> IO ()
onceOnTimeHasCome UpdateState{..} action next =
    bracket init term body
  where
    init = atomically $ do
        timeHasCome <- readTVar usTimeHasCome_
        when timeHasCome $ writeTVar usTimeHasCome_ False
        pure $ if timeHasCome then (action, next) else (pure (), pure ())
    body ( action', _next') = action'
    term (_action',  next') = next'

getUpdateResult :: UpdateState a -> IO a
getUpdateResult us@UpdateState{..} = do
    onceOnTimeHasCome us
        (writeIORef usLastResult_ =<< usUpdateAction_ =<< readIORef usLastResult_)
        (writeIORef usDeleteTimeout_ =<< mkDeleteTimeout usTimeHasCome_ usIntervalMicro_)
    readIORef usLastResult_
```

This ensures that even if the update action is aborted, the term block will reliably register the next timer event.

I hope this helps move the discussion in the right direction.

Vlix:
@mkleczek Could it be what @khibino mentions?
Did the `UTCTime` at some point just not update anymore? And could that have been the reason you were getting timestamps from far in the past?

mkleczek:
> @mkleczek Could it be what @khibino mentions? Did the `UTCTime` at some point just not update anymore? And could that have been the reason you were getting timestamps from far in the past?

Can confirm values far from the past but cannot confirm they stopped refreshing completely.

PostgREST by itself does not use `throwTo`, not sure about `warp` internally. `getCurrentTime` should not throw either. So explanation looks plausible but not immediately obvious.

khibino:
> PostgREST by itself does not use `throwTo`, not sure about warp internally.

I haven't looked closely at the internals of PostgREST, but as a general point, completely avoiding asynchronous exceptions can be a little tricky.

One thing to keep in mind is the `async` library. For example, `concurrently` and `race` will terminate the other thread with an asynchronous exception if one of the threads finishes with an exception.

So, even if the application itself does not directly use `throwTo` or `killThread`, it might be worth considering whether any of the libraries it uses do so internally.

For example, a session-management library that takes an `IO` action might terminate that action on a timeout by sending it an asynchronous exception.

steve-chavez:
> Did the UTCTime at some point just not update anymore?

JFYI, we did had one reported case where this happened and only a restart fixed postgREST JWT validation. This was pretty rare though and we couldn't gather much info. ([mentioned before here](https://github.com/PostgREST/postgrest/issues/5196#issuecomment-5462331503))

Vlix:
This sounds like something that isn't impossible to happen, and that we can fix.
I'll create a new issue to not forget and we can address this in the future.

Thank you @khibino for the insight 🙏

PR #1112, opened 2026-09-02
Use `atomicWriteIORef` when in doubt

Vlix:
I've taken a look at the entire repo and changed any `writeIORef` where:

- the ref was provided as (part of) a function argument
- the ref was returned by the function
- the function returned a function that uses the ref

In short, only if the `IORef` was created and used inside the function itself and there's no way a user could use the ref concurrently.


kazu-yamamoto (approved):
LGTM

Vlix:
Hmmm, `auto-update` tests for `trailingDelayEdge` are a bit flaky. Will try to fix that before merging.

recv/Network/Socket/BufferPool/Buffer.hs
```diff
@@ -37,7 +37,7 @@ newBufferPool l h = BufferPool l h <$> newIORef BS.empty
 withBufferPool :: BufferPool -> (Buffer -> BufSize -> IO Int) -> IO ByteString
 withBufferPool pool@(BufferPool _ _ ref) f = do
     (buf, consumed) <- applyBufferPool pool f
-    writeIORef ref $ BS.drop consumed buf
+    atomicWriteIORef ref $ BS.drop consumed buf
```
mkleczek, line 40: I've analyzed warp and it looks like buffer pool is accessed only from a single thread.

The problem with this change is that it will affect throughput/latency negatively adding memory barrier in uncontended cases.

mkleczek:
> In short, only if the `IORef` was created and used inside the function itself and there's no way a user could use the ref concurrently.

More nuance is needed, I'm afraid - see my comment to Buffer module changes.

`atomicWriteIORef` and friends are costly comparing to their non-atomic counterparts. In case of hot path usage in single threaded cases they add mostly unacceptable overhead. I'd say use of `writeIORef` is OK **if documented properly** and clearly stating that callers have to make sure proper locking and memory barriers are added at the call sites.

Vlix:
I've read up a bit on what memory barriers are and how much they can impact performance. And I think you make a good point to not implement it everywhere.
I'll take a better look at the locations I've changed to see if a `writeIORef` would be fine if the usage is clearly documented/correctly enforced.

wai-extra/Network/Wai/Parse.hs
```diff
@@ -436,10 +439,10 @@ sinkRequestBodyEx
     -> IO ([Param], [File y])
 sinkRequestBodyEx o s r body = do
     ref <- newIORef ([], [])
-    let add x = atomicModifyIORef ref $ \(y, z) ->
+    let add x = modifyIORef' ref $ \(y, z) ->
```
Vlix, line 442: We don't need a memory barrier here, for the same reason we don't need them when using the `writeIORef`s.

Vlix:
@kazu-yamamoto 
I've reviewed the changes, and reverted most of them. I _have_, however, added comments near most of the `writeIORef` usage to explain why we don't need memory barriers (i.e. `atomicWriteIORef`)

I've also changed most `atomicModifyIORef`'s to `atomicModifyIORef'` to keep things strict.

Any other change is a refactor I've found along the way, which I thought would be OK, since this is a refactor PR anyway.

kazu-yamamoto:
@Vlix  Thank you for this work. LGTM. Please merge this PR whenever you are satisfied.

I guess we should make use of AI to find places where memory barrier is necessary.

kazu-yamamoto:
I’ve come up with an idea: rather than using comments to indicate whether an `IORef` is shared or not, it would be better to distinguish them at the type level. For instance, we could use `SVar` for cases where the reference is shared.

It might be a good idea to create a new module named `Network.Wai.Types` within the `wai` package, upon which so many others depend.

```Haskell
newtype SVar a = SVar (IORef a)

readSVar :: SVar a -> IO a
readSVar (SVar ref) = readIORef ref

writeSVar :: SVar a -> a -> IO ()
writeSVar (SVar ref) = atomicWriteIORef ref

modifySVar :: SVar a -> (a -> (a, b)) -> IO b
modifySVar (SVar ref) = atomicModifyIORef' ref
```

Vlix:
I'll try it out and see if it's an improvement to add the type.

Vlix:
@kazu-yamamoto 
On further review, it would probably have to be an entire separate `SVar` package, OR the same module would have to be copied to a bunch of libraries, because we're using `atomicIORef`s across almost half of the package. From `auto-update` and `time-manager, to `warp` and `wai-extra`.

I don't think it will add too much guarantees or readability compared to the current situation, so I've decided to forego the effort at the moment.

I've updated the ChangeLogs of the packages this PR touches, so if this gets all green CI, I'll probably merge and release `auto-update-0.2.8`

kazu-yamamoto:
OK. Please go ahead.

PR #1114, opened 2026-09-13
warp: expose the peer address for connection worker exceptions

### LINKED ISSUE #1113: Warp/WarpTLS: expose accepted peer for exceptions before a Request exists
We found this during application E2E testing of a server built on WarpTLS: pre-TLS failures need to be attributed to the TCP connection that failed, including when clients use different source addresses and connect concurrently.

**Why it matters:** when a TLS connection fails before an HTTP request exists, missing connection context can cause application logs to attribute the failure to the wrong client IP. This can misdirect incident investigations, hide the actual source of suspicious traffic, or - if those logs drive automated blocking - penalize an innocent client.

### Full explanation

Warp's exception observer receives `Nothing` for the request when connection creation fails before a WAI `Request` exists. The accepted `SockAddr` is not supplied through that hook, so an observer cannot reliably identify the failing peer from `setOnException` alone. A shared “last accepted address” is not a safe substitute across connections.

Could Warp expose accepted-peer context for these exceptions, for example through an additive connection-exception callback? This would let applications log pre-request failures without replacing the accept/fork machinery. There is no need to fabricate a WAI Request or parse an untrusted request merely to obtain the TCP peer.

### Tested versions and behavior

All standalone tests below ran on Linux with GHC 9.14.1/base 4.22:

| Warp | WarpTLS | TLS | Real-socket characterization | Desired peer-aware assertion |
| --- | --- | --- | --- | --- |
| 3.4.12 | 3.4.9 | 1.9.0 | failures observed, request absent | Fail: no peer supplied |
| 3.4.15 | 3.4.14 | 2.4.3 | failures observed, request absent | Fail: no peer supplied |

The characterization sends plaintext to a TLS listener from **127.0.0.1 and 127.0.0.2**, sequentially and concurrently. The handler observes `InsecureConnectionDenied`, but `Wai.remoteHost <$> maybeRequest` is `Nothing`. The separate desired-capability assertion fails on the first sequential connection:

```text
expected: Just (Just 127.0.0.1:<client-port>)
 but got: Just Nothing
2 examples, 1 failure
```

The outer `Just` means the callback was observed before the timeout; the inner `Nothing` means no request/peer context was available. This is not a failure to trigger an exception.

A reduced Warp-only test, throwing from the connection maker passed to `runSettingsConnectionMakerSecure`, reproduces the same observer limitation on Warp 3.4.12 and 3.4.15. The current worker path also wraps connection creation with `settingsOnException set Nothing`.

These are exact tested combinations, not a claim that every intermediate release was tested. The newer stack required scoped base/containers/time bound relaxations for its selected Serialise/Cborg dependencies on base 4.22; that compiler-compatibility work is separate from the peer-context finding. Both tested triples satisfy their WarpTLS releases' direct Warp/TLS bounds. Neither reproduction installs our accept/fork workaround or patches Warp/WarpTLS/TLS implementation source.

### Minimal real-socket reproduction

In an unpacked WarpTLS release, save the following as `test/regression/Main.hs`:

```haskell
module Main (main) where
import Control.Concurrent (newChan, readChan, writeChan, newEmptyMVar, putMVar, takeMVar)
import Control.Concurrent.Async (withAsync, mapConcurrently)
import Control.Exception (bracket, fromException)
import Control.Monad (forM_, replicateM)
import qualified Network.Socket as S
import qualified Network.Socket.ByteString as B
import qualified Network.Wai as Wai
import qualified Network.Wai.Handler.Warp as W
import qualified Network.Wai.Handler.WarpTLS as TLS
import System.Timeout (timeout)
import Test.Hspec

main :: IO ()
main = hspec $ describe "real pre-TLS peer context without accept/fork overrides" $ do
  it "observes failures from both peers, with no request context (characterization)" $
    exercise False
  it "retains both peers for pre-handshake error reporting (desired capability)" $
    exercise True

exercise :: Bool -> IO ()
exercise requirePeer = S.withSocketsDo $
  bracket (S.socket S.AF_INET S.Stream S.defaultProtocol) S.close $ \listener -> do
    S.bind listener (S.SockAddrInet 0 (S.tupleToHostAddress (127,0,0,1)))
    S.listen listener 10
    port <- S.socketPort listener
    events <- newChan
    ready <- newEmptyMVar
    let settings = W.setBeforeMainLoop (putMVar ready ()) $
          W.setOnException (\request exception ->
            case fromException exception of
              Just TLS.InsecureConnectionDenied -> writeChan events (Wai.remoteHost <$> request)
              _ -> pure ()) W.defaultSettings
        app _ _ = fail "plaintext must be rejected before application"
        tls = TLS.tlsSettings "test/regression/cert.pem" "test/regression/key.pem"
        peers = map S.tupleToHostAddress [(127,0,0,1),(127,0,0,2)]
        drain client = do
          bytes <- B.recv client 4096
          if bytes == "" then pure () else drain client
        send peer = bracket (S.socket S.AF_INET S.Stream S.defaultProtocol) S.close $ \client -> do
          S.bind client (S.SockAddrInet 0 peer)
          S.connect client (S.SockAddrInet port (S.tupleToHostAddress (127,0,0,1)))
          own <- S.getSocketName client
          B.sendAll client "GET / HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n"
          -- Keep the socket open until WarpTLS sends its denial and closes.
          -- Closing immediately can cause a send error before its TLS exception.
          timeout 2000000 (drain client) `shouldReturn` Just ()
          pure own
        receive = timeout 2000000 (readChan events)
    withAsync (TLS.runTLSSocket tls settings listener app) $ \_ -> do
      timeout 2000000 (takeMVar ready) `shouldReturn` Just ()
      forM_ peers $ \peer -> do
        own <- send peer
        observed <- receive
        observed `shouldBe` (if requirePeer then Just (Just own) else Just Nothing)
      ownPeers <- mapConcurrently send peers
      observations <- replicateM 2 receive
      observations `shouldMatchList` (if requirePeer then map (Just . Just) ownPeers else [Just Nothing, Just Nothing])

```

Generate disposable test certificate fixtures from the package root:

```sh
openssl req -x509 -newkey rsa:2048 -nodes \
  -keyout test/regression/key.pem -out test/regression/cert.pem \
  -days 1 -subj /CN=localhost
```

Add the following test component to `warp-tls.cabal`:

```cabal
test-suite lifecycle-regression
  type: exitcode-stdio-1.0
  main-is: Main.hs
  hs-source-dirs: test/regression
  default-language: Haskell2010
  ghc-options: -Wall -threaded
  build-depends: base, warp-tls, hspec, async, network, bytestring, wai, warp
```

For the newer case, our project configuration was:

```cabal
packages: .
with-compiler: ghc-9.14.1
optimization: 0
constraints: warp ==3.4.15, tls ==2.4.3, http2 ==5.4.0, time-manager ==0.2.4
allow-newer: serialise:base, serialise:containers, serialise:time, cborg:base, cborg:containers
```

Run `cabal test warp-tls:test:lifecycle-regression --test-show-details=direct`. For the older case, use WarpTLS 3.4.9 source and change the Warp/TLS constraints to 3.4.12/1.9.0. The certificate/key are local test fixtures, not production credentials.

The client waits for the server to close after sending its denial. Closing the client immediately after sending can instead cause a send error before the intended `InsecureConnectionDenied` exception, which would not establish this result.

### Application workaround and requested contract

Our application currently pairs `Warp.setAccept` and `Warp.setFork` to transfer the accepted address to the particular worker, with exception-safe ownership and cleanup. Its regression covers both insecure plaintext and prematurely closed TLS connections, sequentially and concurrently. Removing that handoff still breaks attributed observations after upgrading to the newer tested stack.

The standalone test proves **missing pre-request peer context**. It does not claim WarpTLS itself logs another socket's address, and the desired-capability assertion is a proposed API expectation, not a claim that the existing `Maybe Request` contract promises a request in this situation.

Related: [#1033](https://github.com/yesodweb/wai/issues/1033) discusses richer `OnInsecure` handling and the difficulty of creating a Request before TLS completes. This request concerns peer-aware exception observability, including other pre-request failures, rather than richer insecure-request responses.

Would an additive peer-aware exception hook be appropriate? We can contribute the reduced regression and real-socket tests against the agreed API while retaining the existing observer for compatibility.

gorban:
Connection setup can fail before a WAI Request exists, leaving `setOnException` without the peer address. This adds `setOnConnectionException` / `getOnConnectionException`, passing the address already owned by the connection worker. No accept/fork override, shared peer state or synthetic Request is needed.

Fixes https://github.com/yesodweb/wai/issues/1113.

The default delegates to the existing observer with `Nothing`. Installing the new hook replaces that fallback for worker exceptions; request-specific errors and accept failures keep their existing observer. The API documents this distinction and Warp is bumped to 3.4.17.

As requested, the first commit (`20f89aaf`) contains failing regressions and the second (`e63ffc89`) makes them pass. The second changes only the observer adapters for the original regressions, retaining their assertions, and adds compatibility checks.

| Evidence | Before | After |
| --- | --- | --- |
| Sequential/reverse-order concurrent connection-maker peer assertions | 2 failures; legacy control passes | Pass |
| Real TLS failures, portable socket pairs plus Linux 127.0.0.1/.2 | 4 failures: exception observed, peer absent | Pass |
| All enabled Cabal project suites | 343 examples, 6 regression failures | 347 examples, 0 failures |
| Warp/WarpTLS with `-Werror` | Builds; 129 examples, 6 regression failures | 133 examples, 0 failures |
| Additional wai-http2-extra doctests | 9 examples, 0 failures | 9 examples, 0 failures |

Both columns were run locally: “Before” is the test-only commit, including the failing regressions. Its six failures are exactly those regressions; all other examples pass. The fix adds four compatibility examples, accounting for the higher “After” totals. The `-Werror` run compiles successfully on both commits; its “Before” failures are runtime assertions.

Validated on Linux with GHC 9.10.3. Warp Haddock builds. The extra doctests require their library package environment under Cabal, so they were run through `cabal exec` after building the library. Warp's own doctest component remains upstream-disabled.

### Why add a WarpTLS test suite?

Warp already has a suite, which this PR extends with connection-maker regressions and observer compatibility checks. I checked the existing packages, scripts and CI: WarpTLS had example servers (`pong.hs` and `graceful-pong.hs`), but no automated suite elsewhere to extend. The new suite checks the original failure through real sockets and WarpTLS, without application-provided `setAccept` or `setFork` workarounds.

`certificate.pem` and `key.pem` already exist in the repository and are used by those examples; this PR adds no certificate files. Listing them in `extra-source-files` includes the fixtures in source distributions so the tests can run there too. The added dependencies, including `warp >= 3.4.17`, belong only to the test component; WarpTLS's production dependency bounds are unchanged.

No CI workflow change is needed: every CI Stack configuration already includes `./warp-tls`, and the workflow runs `stack test` across the project. It will discover the new `spec` suite and run it on Linux and macOS; Windows builds tests with `--no-run-tests`, as before. Portable cases compare full socket addresses, with additional 127.0.0.1/.2 cases on Linux; the connection-maker regressions check distinct IPs without OS-specific binding requirements. Actual matrix results remain to be verified after submission.

After submitting your PR:


kazu-yamamoto:
@Vlix This PR adds `settingsOnConnectionException` to `Settings`. The test in the initial commit fails on Linux but passes on macOS. Since the default value for `settingsOnConnectionException` is `Nothing`—in which case `settingsOnException` is called—backward compatibility is maintained. I think this PR is acceptable; what do you think?

warp/Network/Wai/Handler/Warp.hs
```diff
@@ -210,6 +212,23 @@ setOnException
...
+-- by its connection source. This includes connection creation (such as a TLS
+-- handshake) and cleanup failures, even when no 'Request' exists. For socket
+-- listeners this is the accepted TCP peer, before any PROXY protocol rewriting.
+--
+-- When installed, this handler replaces 'setOnException' with 'Nothing' for
+-- these exceptions. Request-specific exceptions still use 'setOnException',
+-- as do accept-loop failures for which no peer is available. By default Warp
+-- delegates to the current 'setOnException' handler with 'Nothing'.
+--
+-- This additive hook preserves the existing observer contract without requiring
```
Vlix: This sentence seems irrelevant to users of the API. Please remove.

warp/Network/Wai/Handler/Warp.hs
```diff
@@ -334,6 +353,13 @@ getOnClose = settingsOnClose
 getOnException :: Settings -> Maybe Request -> SomeException -> IO ()
 getOnException = settingsOnException
 
+-- | Get the connection exception handler, including the fallback to
```
Vlix: This is entire description is somewhat confusing. What is this `'Nothing'` referring to? Please rephrase.

warp/test/ConnectionExceptionSpec.hs
Vlix, line 1: Using `1` for everything makes it easy to get false positives. please use different mock numbers for different situations/types. (e.g. use `peer 100` and `ConnectionFailure 1`)

warp/test/ConnectionExceptionSpec.hs
```diff
@@ -0,0 +1,130 @@
...
+        runSettingsConnectionMakerSecure settings (ioError (userError "accept failed")) unusedApplication
+        readIORef calls `shouldReturn` [("legacy", False)]
+
+    it "keeps application exceptions on the legacy observer with their request" $
+        bracket (N.bindRandomPortTCP "127.0.0.1") (close . snd) $ \(port, listener) -> do
+            events <- newChan
+            connectionCalled <- newIORef False
+            ready <- newEmptyMVar
+            let legacy request exception = writeChan events (isJust request, show exception)
+                connection _ _ = modifyIORef' connectionCalled (const True)
```
Vlix: `modifyIORef' ref (const x) === writeIORef ref x`

Vlix (changes requested):
Tests seem good, and the change is fine to me.

I only have a bit of (documentation) change requests.

warp/Network/Wai/Handler/Warp.hs
gorban: Removed that implementation-rationale sentence from the public Haddock in 9adb420a.

warp/Network/Wai/Handler/Warp.hs
gorban: Rephrased in 9adb420a: without an installed connection handler, the returned function ignores the peer address and calls the ordinary exception handler with `Nothing` for its `Maybe Request` argument, forwarding the exception. I clarified the setter’s corresponding paragraph as well.

warp/test/ConnectionExceptionSpec.hs
gorban: Updated in 9adb420a: exception IDs use 1–8 and mock peer IDs use 100–160, with distinct values across scenarios. The sequential and reverse-order concurrent cases still assert the exact exception/peer pairing. All 133 Warp/WarpTLS examples pass with `-Werror`.

warp/test/ConnectionExceptionSpec.hs
gorban: Replaced it with `writeIORef connectionCalled True` in 9adb420a.

gorban:
All tests passed on my commit before the documentation changes:
> [All checks have passed](https://github.com/yesodweb/wai/pull/1114/commits/cd0dcdaf2bd36b69c34826e90c5100c4d47b1d0c)
> 18 successful checks

But failed with doc changes? I am thinking the flakiness has nothing to do with my changes. I put in an empty commit to re-run CI to test for flakiness, but it hasn't been approved? If someone is more familiar with the normal course of action, could they suggest anything?

Vlix (commented):
Looks good. I'll rerun the actions and add this PR to the release schedule.

Vlix:
Also, this change would be included in the `warp-3.4.16` release, so please keep the version in cabal at `3.4.16` and add the entry in the Changelog to the `3.4.16` section.

gorban:
warp.cabal is at 3.4.16, with the ChangeLog entry in the ## 3.4.16 section (alongside #1112).

warp-tls is aligned the same way: cabal version, ChangeLog heading, and the test-suite's warp lower bound all at 3.4.16.

Branch is rebased onto current master; warp:spec and warp-tls:spec both pass with -Werror. So I don't have any history of successful CI builds anymore (new SHAs and all), but it was passing before docs changes started.

Vlix:
`warp-tls` does not need a version bump. Please revert the cabal file change and the ChangeLog entry.

gorban:
Done in a871d0c3 — `warp-tls` no longer claims a version of its own:

- `warp-tls/warp-tls.cabal` is back to `Version: 3.4.14`, matching master.
- `warp-tls/ChangeLog.md` is byte-identical to master; the added section is removed entirely rather than renamed back to `Unreleased`.

What remains in warp-tls is test material only: the new `test/Spec.hs` and its `Test-Suite spec` stanza, plus `certificate.pem`/`key.pem` in `extra-source-files` so the fixtures ship in the sdist. The suite's `warp >= 3.4.16 && < 3.5` bound stays, because it exercises the new `setOnConnectionException`.

The `Tests` run for a871d0c3 is sitting at `action_required` (https://github.com/yesodweb/wai/actions/runs/35416553706)

Vlix (approved):
Looks good.
This will be added to the `warp-3.4.16` release.

PR #1115, opened 2026-09-17
ci: better caching for quicker CI feedback.

Vlix:
* more granular caching to share caches more and to lower the size of caches
* added lock files to repo to cache dependencies better
* tweaking scripts to build as quick as possible
* YAML linter also changed a bit, but not too much

---

I was getting annoyed at the CI still taking too long.
Mostly because we build so much, that we blow our 10 GB GitHub cache size with every CI run we do. So this is taking some advice I found from the stack repo (or related) to do better caching in GitHub actions.


kazu-yamamoto:
Never mind. I did not notice that this PR is already merged.
Good job, anyway!

PR #1116, opened 2026-09-17
Timeout watchdog

goertzenator:

After submitting your PR:

## Summary
This is a proof of concept of the "timeout watchdog" described in PR #1109.  The idea is that instead of having connection code directly manipulate the timer to manage timeouts, the code instead records the current "phase" (ie, receiving headers, running application, sending response) into a `TVar` and a separate watchdog thread monitors that `TVar` to make timeout judgments.

## Design
As described above, the main connection flow will now write the current "phase" into a `TVar`.  Also in the `TVar` is a `Int` counter tag so that tickles without phase changes can be observed by the watchdog.

The watchdog thread runs alongside the connection thread using the `race` function.  The watchdog thread can kill the connection by just exiting. Before the watchdog does anything interesting, it does a grace period `threadDelay` of up to 5 seconds, the purpose of which is to minimize work done on short lived connections.  After the grace period expires the watchdog will begin observing the TVar and running additional timeouts.  The watchdog loop includes a throttling 1 second `threadDelay` which approximates the throttling behavior from PR #1109 and PR #1097.

The watchdog thread implementation makes it easy to add more complex timeout behavior.  This proof of concept implements a total timeout for reading the headers; this is a simpler and more efficient way to implement slowloris protection.  The single timeout value parameter here is used for all timeouts, but this could be easily extended to provide different timeouts for each phase.

## Scope

This proof of concept is for http/1.1 implemented within Warp.  http2 is unchanged, although the hope is that work here carries over to http2.

## Further Development
- Multiple timeout parameters for different phases.
- Integration to http2 (and http3?).
- Store the TVar in a more accessible place.  The TVar could be useful for debugging or other observers.
- Grace period optimization: Move the grace period into a timer that launches the watchdog thread upon completion.  A short running connection would end up not launching the watchdog thread at all.  I had this running while developing this PoC, but it was complicated and I could not be sure it was free of races.

## Benchmarks
Included in this PR is an extensive socket-based benchmark.  It is AI written.  I have given it some review, but it does need more.  It has a nice support script for comparing multiple branches.  Here is the invocation to compare master, PR #1109, and this PR:

`warp/bench/compare.sh origin/master origin/perf/time-manager-tickle .`

The benchmarks are:

**keepalive** — 100 connections issuing requests back to back, 8s after a 2s
warm-up. Connections are reused, so this isolates the per-request hot path.

**stream** — 16 connections, chunked responses of 64 fragments each, 8s after
a 2s warm-up. Isolates per-chunk cost.

**churn** — 150,000 connections at 50 concurrent, each a connect, one request
with `Connection: close`, and a close. Count-bounded to stay inside the
ephemeral port range. Isolates per-connection cost.

**idle** — 2000 connections established unpaced, one request each, then held
idle for two 8s windows. Establishment and the two windows are reported
separately so a one-off per-connection cost isn't averaged into a rate.

edit: My CPU is a AMD Ryzen 9 3900X.

```
--- keepalive ---
metric                      origin-master  origin-perf-time-manager-tickle    working-tree   best
req/s                              14,999                          180,594         205,125   working-tree
server cpu ns/req                 127,388                           14,224          11,130   working-tree
p50 ns                          6,677,000                          543,000         463,000   working-tree
p99 ns                          7,809,000                        1,051,000       1,057,000   origin-perf-time-manager-tickle

--- stream ---
metric                      origin-master  origin-perf-time-manager-tickle    working-tree   best
fragment/s                         80,303                          418,270         419,184   working-tree
cpu ns/fragment                    23,242                            5,804           5,711   working-tree
p50 ns                         12,489,000                        2,296,000       2,274,000   working-tree

--- churn ---
metric                      origin-master  origin-perf-time-manager-tickle    working-tree   best
conn/s                             18,598                           26,540          29,742   working-tree
server cpu ns/conn                112,333                          101,733          83,067   working-tree
p50 ns                          2,619,000                        1,926,000       1,679,000   working-tree
p99 ns                          4,397,000                        2,719,000       2,541,000   working-tree

--- idle ---
metric                      origin-master  origin-perf-time-manager-tickle    working-tree   best
establish conn/s                   10,422                           14,810          17,741   working-tree
establish cpu ns/conn             150,000                          110,000          90,000   working-tree
first-win cpu ns/conn              25,000                           10,000          50,000   origin-perf-time-manager-tickle
total cpu ns/conn                 175,000                          125,000         140,000   origin-perf-time-manager-tickle
rss kb/conn                         30.14                            26.45           38.06   origin-perf-time-manager-tickle
steady cpu ns/conn/s                 0.00                             0.00            0.00   (tie)
```


warp/Network/Wai/Handler/Warp/Watchdog.hs
```diff
@@ -0,0 +1,296 @@
...
+phaseBudget
+    :: Int
+    -- ^ Timeout in microseconds.
+    -> Phase
+    -> Budget
+phaseBudget us ReadingHeaders = Total us
+phaseBudget us ReadingBody = Extends us
+phaseBudget _ RunningApp = NoLimit
+phaseBudget us SendingResponse = Extends us
+phaseBudget _ Delegated = NoLimit
```
Vlix, line 151: I personally find a case statement to be more readable:

```suggestion
phaseBudget
    :: Int
    -- ^ Timeout in microseconds.
    -> Phase
    -> Budget
phaseBudget us phase =
    case phase of
        ReadingHeaders -> Total us
        ReadingBody -> Extends us
        RunningApp -> NoLimit
        SendingResponse -> Extends us
        Delegated -> NoLimit
```

warp/Network/Wai/Handler/Warp/Watchdog.hs
```diff
@@ -0,0 +1,296 @@
...
+            stillMoving <-
+                withTimer tmgr t (atomically $ writeTVar expired True) $
+                    atomically $ do
+                        tagged <- readTVar cs
+                        e <- readTVar expired
+                        check $ e || moved tagged
+                        return $ moved tagged
+            if stillMoving
+                then threadDelay rateLimit >> loop
+                else return phase
```
Vlix, line 263: This technically returns the "old" phase.

Thankfully this is ok since both `moved` arguments that are used are `True` on `/= phase`, but that is not obvious from inside `attempt`.
I think it'll be fine if we return the `phase` from `tagged`, so that the returned phase is the latest phase we waited on when the timeout triggered.

On the other hand, we could also just return the `tagged :: TaggedPhase` to lose less information, and extract the `Phase` in `loop`?

```suggestion
        attempt :: Int -> (TaggedPhase -> Bool) -> IO Phase
        attempt t moved = do
            expired <- newTVarIO False
            (stillMoving, phase) <-
                withTimer tmgr t (atomically $ writeTVar expired True) $
                    atomically $ do
                        tagged <- readTVar cs
                        e <- readTVar expired
                        check $ e || moved tagged
                        return (moved tagged, getPhase tagged)
            if stillMoving
                then threadDelay rateLimit >> loop
                else return phase
          where
            getPhase (TaggedPhase _ p) = p
```

warp/Network/Wai/Handler/Warp/Watchdog.hs
```diff
@@ -0,0 +1,296 @@
...
+
+-- | What a connection is currently doing.
+--
+-- These are exactly the distinctions Warp's timeout policy is able to make,
+-- because 'phaseBudget' is defined on them and nothing else.
+data Phase
+    = -- | Waiting for, or reading, a request line and headers. Covers the gap
+      -- between keep-alive requests. This is the slowloris case, which is why
+      -- 'phaseBudget' gives it a 'Total' budget rather than an idle one.
+      ReadingHeaders
```
Vlix, line 60: Should we also count the making of the connection as a phase?

I think I've read that when making a TLS connection, it could also be a vector of a slow loris attack. 🤔

warp/Network/Wai/Handler/Warp/Watchdog.hs
```diff
@@ -0,0 +1,296 @@
...
+    -- ^ Timeout in microseconds.
+    -> (ConnState -> IO ())
+    -> IO ()
+withConnWatchdog us action = do
+    cs <- newConnState
+    if us <= 0
+        then
+            action cs
+        else
+            race_ (watchdog us cs) (action cs)
```
Vlix, line 176: I believe this will throw an `AsyncCancelled` to the `action`, right?

Is there any way we can keep using the `TimeoutThread` exception, so that the behaviour expecting that particular exception isn't broken? (possible in user code)

If we get rid of `time-manager` as a dependency, we'd have to use a custom exception, probably, which would warrant a major version bump.

Vlix (commented):
Only reviewed the `Watchdog.hs` file. Will look at the implementation at a later time.

Vlix:
Also, rebase on master to get new and improved CI caching.

warp/Network/Wai/Handler/Warp/Watchdog.hs
goertzenator: Good catch.  I think we would have to unroll our own `race_` to get the right exception.  But at that point we may as well go with the grace timer optimization too (watchdog doesn't start at all until after grace period).

warp/Network/Wai/Handler/Warp/Watchdog.hs
goertzenator: That's a good point.  The watchdog would have to hook in at some earlier point in the code.

goertzenator:
Thanks for the review @Vlix.  I am happy to do further work on this, but I would want confirmation that this has a good chance of being accepted first.

Vlix:
I hope @kazu-yamamoto will be able to take a look at this as well, before we decide on it, but my personal opinion is that this is probably a good direction to go in.

I've found that Windows tests for `time-manager` also fail (which is the reason we don't run the Windows tests, which is kind of bad, tbh) and if we can remove an entire "dependency" in the behaviour of the server, that seems like a win to me.

We'll probably want to be careful when testing/releasing this change, though. Since doing it wrong could very negatively impact users. So I want to make sure this is strictly an improvement (or at least keeps the status quo).

kazu-yamamoto:
I would like to review this PR. But it would take time since I have a lot of PRs to review. ^^;

Vlix:
I don't think there's any rush, so take your time :)

goertzenator:
`http2` is still depending on `time-manager`, so that would also need work.  However, timeout problems in `http2` is what brought me here in the first place (https://github.com/kazu-yamamoto/http2/pull/174) so I will take a swing at that too.

PR #1117, opened 2026-09-17
Re enable windows tests

Vlix:
Wanted to get testing running again to be able to catch regressions on Windows (on tests other than the ones we know are failing)


Vlix:
Not quite sure why, but the tests hang on Windows/nightly/`warp`. 

The moment it hits the `HTTP/2 Early Hints` test, it seems to just stop and wait forever.

Vlix:
Opted for "commenting out" the failing Windows tests, so that we at least are testing everything else on Windows.

